PR #226, opened 2024-05-10
feat: optimize dockerfiles

pumbas600:
I was having issues building the original docker files. These also only include what's necessary so the images are MUCH smaller: ~1.6GB to 160 MB.


pumbas600:
Superseded by #467

PR #228, opened 2024-05-10
feat: stats me skeleton

sou323:


packages/frontend/src/app/me/StatsTable.tsx
```diff
@@ -39,39 +46,54 @@ export default function StatsTable({
...
+          </th>
           <td>
             {statsAreLoading ?
-              "Loading…"
+              <Skeleton width={150} />
             : mostRecentTimestamp ?
-              formatTimestampLocalTZ(mostRecentTimestamp)
+              <span title={formatTimestamp(mostRecentTimestamp)}>
+                {formatTimestampLocalTZ(mostRecentTimestamp)}
+              </span>
```
jaskfla, line 96: [`<time>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time) 😉

packages/frontend/src/app/me/StatsTable.tsx
```diff
@@ -39,39 +46,54 @@ export default function StatsTable({
...
 }: StatsTableProps) {
-  if (!stats)
+  if (!stats && !statsAreLoading)
     return (
       <EmptyStateMessage>You don’t have any stats (yet)!</EmptyStateMessage>
     );
 
-  const { totalPixels, mostFrequentColor, mostRecentTimestamp, rank } = stats;
+  const { totalPixels, mostFrequentColor, mostRecentTimestamp, rank } =
+    stats ?? {};
```
jaskfla, line 55: Component returns early if `!stats`, so this `??` is redundant

PR #246, opened 2025-02-25
chore(deps): bump deps

jaskfla:
Bumps minor and patch versions, but not major versions


hwan513 (approved):
Builds on my machine. pnpm update will still perform some weird shenanigans on running further updates on the backend package though.

PR #248, opened 2025-02-25
refactor: helper for not forwarding transient props

jaskfla:
Not yet tested—written blind

### DIFF
--- a/packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
@@ -1,6 +1,6 @@
 import { styled } from "@mui/material";
 
-import { startsWith$ } from "@/util";
+import { doesNotStartWith$ } from "@/util";
 
 export const ActionPanelTabBody = styled("div")`
   display: block flex;
@@ -34,7 +34,7 @@ export const Block = styled("div")`
 `;
 
 export const TabBlock = styled(Block, {
-  shouldForwardProp: startsWith$,
+  shouldForwardProp: doesNotStartWith$,
 })<{
   $active?: boolean;
 }>`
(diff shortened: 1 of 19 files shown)


packages/frontend/src/components/button/DynamicButton.tsx
```diff
@@ -73,7 +74,7 @@ export default function DynamicButton({
   const rgba = color?.rgba;
   const rgb = rgba?.slice(0, 3).join(" ");
 
-  const backgroundColorStr = color ? `rgb(${rgb})` : undefined;
+  const $backgroundColorStr = color ? `rgb(${rgb})` : undefined;
```
pumbas600, line 77: Interesting, do you think `$` should be used for the variable outside the context of the styled component?

pumbas600 (approved):
I haven't tested this, but it looks good. Just one small comment about the convention around using `$` :)

packages/frontend/src/components/button/DynamicButton.tsx
jaskfla: Negative! Just a find and replace gone bad

Good catch

pumbas600 (approved):

packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
```diff
@@ -32,9 +34,11 @@ export const Block = styled("div")`
...
 
 export const TabBlock = styled(Block, {
-  shouldForwardProp: (prop) => prop !== "active",
-})<{ active?: boolean }>`
-  display: ${({ active }) => (active ? "grid" : "none")};
+  shouldForwardProp: doesNotStartWith$,
+})<{
+  $active?: boolean;
+}>`
+  display: ${(props) => (props.$active ? "grid" : "none")};
```
pumbas600, line 41: You've not updated the usages of this to use the new `$active`

pumbas600 (changes requested):
There is something funky going on in `ActionPanel.tsx`. We've got an extra `}` in the `TabBar`, but there's also a few errors in the `Tab` styled component (Also in that file), but I can't see why it's not happy...

Also can you update the `Tab` component to use this for the `active` prop it has too!

jaskfla:
Not convinced this is an improvement

PR #249, opened 2025-02-25
refactor: CSS hygiene

jaskfla:


pumbas600 (commented):
Whoops. I forgot to complete this review 😓

packages/frontend/src/components/ColorCodeChip.tsx
```diff
@@ -14,18 +14,18 @@ const Container = styled("code", {
...
 
-  :focus,
-  :focus-visible,
-  :hover {
-    background-color: oklch(from var(--discord-white) l c h / 20%);
+  @media (hover: hover) and (pointer: fine) {
+    &:hover {
+      background-color: oklch(from var(--discord-white) l c h / 20%);
+    }
   }
```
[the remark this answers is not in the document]
jaskfla: It prevents the hover effect from showing up when you tap (or tap & hold) on a touch screen

packages/frontend/src/components/VisuallyHidden.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  useEffect(function forceShowOnAltDown() {
+    if (process.env.NODE_ENV === "production") return;
+
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === "Alt") setForceShow(true);
+    };
+    const handleKeyUp = (e: KeyboardEvent) => {
+      if (e.key === "Alt") setForceShow(false);
+    };
```
pumbas600, line 33: This is very nifty!

packages/frontend/src/components/action-panel/tabs/BotCommandCard.tsx
```diff
@@ -52,7 +55,8 @@ export default function BotCommandCard() {
     <Wrapper>
       <code>{command}</code>
       <CopyButton onClick={() => navigator.clipboard.writeText(command)}>
-        <StyledCopyIcon />
+        <StyledCopyIcon aria-hidden />
+        <VisuallyHidden>Copy bot command</VisuallyHidden>
```
pumbas600, line 62: What is the benefit of using visually hidden text here rather than an `aria-label` on the button?

packages/frontend/src/components/canvas/CanvasPicker.tsx
```diff
@@ -44,18 +43,21 @@ const Select = styled(NativeSelect)`
     cursor: wait;
   }
 
-  :hover:not(.${nativeSelectClasses.disabled}) {
-    background-color: var(--discord-legacy-greyple);
+  @media (hover: hover) and (pointer: fine) {
+    &:hover {
+      content: unset;
```
pumbas600: This looks weird to me, especially as this isn't in a `::before`/`::after` pseudo element (At least as far as I can tell). Would you mind adding a comment?

pumbas600 (changes requested):
I've not been able to check it out locally, but looking at the code I just had a few minor queries :)

packages/frontend/src/components/action-panel/tabs/BotCommandCard.tsx
jaskfla: Not a lot, the biggest is that to date no translation services (think Google Translate) recognise ARIA labels

The other is the new devX detail where our implementation lets us hold <kbd>Alt</kbd> to show these accessibility labels in dev environments

packages/frontend/src/components/canvas/CanvasPicker.tsx
jaskfla: Oh whoa I think this is a bad merge resolution. Thanks

jaskfla: efd07c5cba58e8b2fb1d9106d7d27717ea7ffbf7

Rocked03 (approved):

pumbas600 (approved):
This looks great, thanks Jasper! :)

PR #250, opened 2025-03-08
feat: mobile support

pumbas600:
I have not tested this with loaded data as we don't have a database, but it's relatively usable now. The mobile menu icon is just using a MUI `Menu` because I couldn't be bothered making something more complicated :sob:

I do want to look into using a MUI drawer for the action panel as it has support for swiping it up/down which would be awesome, but I think that's for another PR.

image

### DIFF
--- a/packages/frontend/src/components/action-panel/ActionPanel.tsx
+++ b/packages/frontend/src/components/action-panel/ActionPanel.tsx
@@ -10,7 +10,8 @@ import { PixelInfoTab, PlacePixelTab } from "./tabs";
 const Wrapper = styled("div")`
   --padding-width: 1rem;
   background-color: var(--discord-legacy-not-quite-black);
-  border-radius: var(--card-border-radius) var(--card-border-radius) 0 0;
+  border-start-start-radius: var(--card-border-radius);
+  border-start-end-radius: var(--card-border-radius);
   border: var(--card-border);
   display: grid;
   gap: 0.5rem;
--- a/packages/frontend/src/components/action-panel/ActionPanel.tsx
+++ b/packages/frontend/src/components/action-panel/ActionPanel.tsx
@@ -10,7 +10,7 @@ import { PixelInfoTab, PlacePixelTab } from "./tabs";
 const Wrapper = styled("div")`
   --padding-width: 1rem;
   background-color: var(--discord-legacy-not-quite-black);
-  border-radius: var(--card-border-radius);
+  border-radius: var(--card-border-radius) var(--card-border-radius) 0 0;
   border: var(--card-border);
   display: grid;
   gap: 0.5rem;
@@ -22,25 +22,18 @@ const Wrapper = styled("div")`
   > * {
     border-radius: calc(var(--card-border-radius) - var(--padding-width));
   }
+
+  ${({ theme }) => theme.breakpoints.up("md")} {
+    border-radius: var(--card-border-radius);
+  }
 `;
 
 const TabBar = styled("ul")`
-  border-radius: .5rem;
+  border-radius: 0.5rem;
   display: grid;
-  gap: .5rem;
+  gap: 0.5rem;
   grid-template-columns: repeat(2, 1fr);
   list-style-type: none;
-
-
-
-    /*
-     * Workaround for accessibility issue with VoiceOver.
-     * See https://gerardkcohen.me/writing/2017/voiceover-list-style-type.html
-     */
-    li::before {
-      content: "\\200B"; /* zero-width space */
-    }
-  }
 `;
 
 const Tab = styled("li")<{ active?: boolean }>`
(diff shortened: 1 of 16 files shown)


packages/frontend/src/components/nav/NavLinks.tsx
```diff
@@ -0,0 +1,133 @@
...
+  transition:
+    background-color var(--transition-duration-fast) ease,
+    opacity var(--transition-duration-fast) ease,
+    outline-width var(--transition-duration-fast) ease;
+
+  :hover {
+    opacity: 55%;
+  }
+
+  :focus,
```
jaskfla, line 43:
```suggestion
```

packages/frontend/src/components/nav/NavLinks.tsx
```diff
@@ -0,0 +1,133 @@
...
+  display: block;
+
+  ${({ theme }) => theme.breakpoints.up("md")} {
+    display: none;
+  }
+
+  & > svg {
+    width: 1.75rem;
+    height: 1.75rem;
+  }
```
jaskfla, line 60: MUI exports a symbol for a classname we can use here, right?

packages/frontend/src/components/nav/NavLinks.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const handleOpen = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorElement(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorElement(null);
+  };
+
+  return (
+    <>
```
jaskfla, line 98: Now that this component exists, it probably makes more sense to make this `Fragment` a `<nav>`, and make the root of `Navbar.tsx` a `<header>`?

jaskfla (commented):
Won’t approve until we’ve tested, but a couple small readability/semantics notes

Thanks!

packages/frontend/src/components/nav/NavLinks.tsx
pumbas600: I'm not sure, because these aren't MUI icons...

packages/frontend/src/components/nav/NavLinks.tsx
pumbas600: Yes you're 100% right!

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -10,7 +10,7 @@ import { PixelInfoTab, PlacePixelTab } from "./tabs";
 const Wrapper = styled("div")`
   --padding-width: 1rem;
   background-color: var(--discord-legacy-not-quite-black);
-  border-radius: var(--card-border-radius);
+  border-radius: var(--card-border-radius) var(--card-border-radius) 0 0;
```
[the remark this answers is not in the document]
pumbas600: I didn't know about this, neat!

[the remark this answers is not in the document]
pumbas600: Hmmm... Styled components doesn't seem to know about this style for some reason

packages/frontend/src/components/nav/NavLinks.tsx
pumbas600: It doesn't seem like lucide exports their class names (Albeit the only classes it has are `lucide` and `lucide lucide-menu`)

packages/frontend/src/components/action-panel/ActionPanel.tsx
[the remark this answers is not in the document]
jaskfla: Oh, my mistake, that was invalid CSS

```suggestion
  border-start-start-radius: var(--card-border-radius);
  border-start-end-radius: var(--card-border-radius);
```

packages/frontend/src/components/nav/NavLinks.tsx
jaskfla: Ah I forgot we don’t use MUI!

packages/frontend/src/components/header/Header.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  ${({ theme }) => theme.breakpoints.up("md")} {
+    display: block;
+  }
+`;
+
+const Logo = styled(Image)`
+  min-width: ${(props) => props.width}px;
+  min-height: ${(props) => props.height}px;
+`;
```
jaskfla, line 46: Purely a preference thing, but another option would be:

```suggestion
const Logo = styled(Image)(props => css`
  min-width: ${props.width}px;
  min-height: ${props.height}px;
`);
```

packages/frontend/src/components/action-panel/ActionPanel.tsx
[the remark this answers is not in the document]
hwan513: Accepted this.

hwan513:
One of the changes seems to break the `overflow-y` behaviour. Looking into this tomorrow.

sou323 (approved):

All looks good to me!

PR #257, opened 2025-03-12
fix(frontend): blurry canvas in safari

hwan513:
`<canvas />` in conjunction with CSS scale and transform was causing the blurry images.

~~Fixed this by removing `<PreviewCanvas />` and applying `hidden` to the canvas canvas. Canvas canvas blob is then used as a source for an `` which isn't as cursed in Safari (still when zoom is like 125% I think).~~

~~Since the blob is recreated each time a pixel is placed I'd assume there is some performance hit especially at higher loads.~~

Remove usage of `<canvas />` to render entire image. Updates are rendered as single pixel images generated from canvas and overlayed over the blurple canvas.

Improvements related to this.
- [x] Optimise image loading (single pixels pngs or whatnot.
- [ ] Make entire canvas container zoomable.
- [ ] Touch and Mouse events.


packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -622,13 +597,19 @@ export default function CanvasView() {
...
+          <DisplayCanvas ref={canvasRef} isLoading={isLoading} hidden />
+          {canvasImageUrl && (
+            <div ref={canvasImageWrapperRef}>
+              <img
+                src={canvasImageUrl}
+                alt="Active Canvas"
+                style={{
+                  imageRendering: "pixelated",
+                  pointerEvents: "none",
+                }}
```
pumbas600, line 609: Can this be moved into a `styled()`

pumbas600 (changes requested):
On Chrome I have to switch to another canvas before I can zoom. I suspect the listeners aren't being correctly registered initially?

pumbas600 (changes requested):
Panning and zooming seems to be working now! :tada:

Unfortunately, the live updating seems to be broken. It appears to not be receiving the pixels updates for some reason? Is it correctly dispatching the event?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -622,24 +595,49 @@ export default function CanvasView() {
...
+    ctx.fillStyle = `rgba(${r}, ${g}, ${b}, ${a / 255})`;
+    ctx.fillRect(payload.x, payload.y, 1, 1);
+    offscreenCanvas.convertToBlob().then((blob) => {
+      const pixelImage = new Image();
+      pixelImage.src = URL.createObjectURL(blob);
+      pixelImage.onload = () => {
+        canvasImageWrapperRef.current?.appendChild(pixelImage);
+      };
+    });
+  }
```
pumbas600, line 642: I'm pretty happy with how this all functions, but just for consistency with the rest of the code base can this be an arrow function above the return...

I understand that `function`s behave slightly differently, so you may need to wrap this in a `useCallback` if it's being used in something with a dependency array

pumbas600 (changes requested):
This is looking good! I'm not sure quite what was happening with my live updating but it seems to be working now (We'll need to keep an eye on this though)

Just one small refactor. I've not tested on a laptop yet though, but I'll do that tomorrow, but otherwise I think this is a big win :)

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Moved it into the useEffect, turns out it wasn't being reused anywhere (and it probably won't be).

pumbas600 (approved):
This is huge! I've just some really small minor comments, but otherwise this is great.

I tested and it works for me on Chrome and Firefox on Windows, and Chrome and Safari (Without blur!) on MacOS!

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -517,26 +510,18 @@ export default function CanvasView() {
    */
   const handleCanvasClick = useCallback(
     (event: MouseEvent): void => {
-      if (!canvasRef.current) return;
+      if (!(event.target instanceof HTMLElement)) return;
```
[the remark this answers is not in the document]
hwan513: The click event is attached to the canvas image wrapper directly so it wouldn't be needed. Ideally the `canvas-image-wrapper` check shouldn't be needed, only doing this currently due to another bug.

PR #260, opened 2025-03-14
fix: early return before all hooks in `BotCommandsCard`

### LINKED ISSUE #259: Fix early exit before hooks in BotCommandCard component
In our BotCommandCard component, we sometimes return early before calling all our hooks. This is not allowed in React and I'm surprised we've not noticed until now 😓 

Image

pumbas600:
Fixes #259 

**What's Changed?**
- Fixed early return before hooks.
- Fixed issue with our biome issue.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -12,7 +12,7 @@
   },
   "javascript": {
     "formatter": {
-      "trailingCommas": "all"
+      "trailingComma": "all"
     }
   },
   "vcs": {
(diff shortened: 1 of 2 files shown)


sou323 (approved):
lgtm

biome.json
jaskfla, line 1: I think you might’ve forgotten to `pnpm i`? The singular has been deprecated

PR #261, opened 2025-03-14
chore: dont clear colour when placing pixel

### LINKED ISSUE #258: Dont unselect colour after placing pixel
Currently, when you place a pixel your selected colour gets unselected, which means you have to select it again before placing the next pixel. As it's very common to want to place multiple pixels in the same colour one after the other, this becomes quite annoying.

pumbas600:
Closes #258


sou323 (approved):
lgtm

PR #262, opened 2025-03-15
feat(frontend): zooming and panning improvements

hwan513:
~~Would end up removing at lot of the glide zoom code in favour of css animations (and potentially the panning velocity code as well).~~

~~Should be able to:~~

~~- [x] Change zoom~~
~~- [x] Change panning~~
~~- [x] Safari...~~

Changes:

- Zoom animation is done using css transition (transition is disabled on Safari due to bad support) for performance reasons (high polling scroll wheels) and also use deltaY extensively in scrolling logic—should feel more responsive on both trackpad and mouse. 
- Panning is not done using css transition (can't be used on Safari) and doesn't seem to have heavy performance issues.
- Updates touchevent and mouseevent handlers to use the same pointer events.
- Allows scrolling on canvas background.
- Fixes scrolling not being able to differentiate from sub pixels.
- Fixes pan decay event firing recursively until floating point precision prevents it to.

Also seems to be a bug where the boundary clamp is applied too late (later event occurrence).

### DIFF
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -168,15 +168,12 @@ export default function CanvasView() {
   const [offset, setOffset] = useState(ORIGIN);
   const [velocity, setVelocity] = useState<Point>({ x: 0, y: 0 });
   const [controlledPan, setControlledPan] = useState(false);
-  const [targetZoom, setTargetZoom] = useState(1);
-  const [mouseOffsetDirection, setMouseOffsetDirection] = useState(ORIGIN);
 
   const handleLoadImage = useCallback((image: HTMLImageElement): void => {
     const initialZoom =
       containerRef.current ? getDefaultZoom(containerRef.current, image) : 1;
 
     setZoom(initialZoom);
-    setTargetZoom(initialZoom);
     setVelocity(ORIGIN);
     setOffset(ORIGIN);
     setIsLoading(false);
@@ -278,6 +275,7 @@ export default function CanvasView() {
         x: event.offsetX,
         y: event.offsetY,
       };
+
       // Ensures that the handler can be added to a parent element but only operates on the canvas image wrapper.
       // Applying the handler to lower elements for some isn't consistently picked up in certain browsers (Firefox and Chrome).
       // Ideally, the scrolling should work outside of canvas-image-wrapper, but I can't seem to get the behaviour correct.
@@ -287,44 +285,15 @@ export default function CanvasView() {
 
       // The mouse position's origin is in the top left of the canvas. The offset's origin is the
       // center of the canvas so we do this to convert between the two.
-      setMouseOffsetDirection(
-        diffPoints(
-          {
-            x: elem.offsetWidth / 2,
-            y: elem.offsetHeight / 2,
-          },
-          mousePositionOnCanvas,
-        ),
+      const mouseOffsetDirection = diffPoints(
+        { x: elem.offsetWidth / 2, y: elem.offsetHeight / 2 },
+        mousePositionOnCanvas,
       );
 
       const scale = Math.exp(Math.sign(-event.deltaY) * SCALE_FACTOR);
-      const newZoom = clamp(targetZoom * scale, MIN_ZOOM, MAX_ZOOM);
+      const newZoom = clamp(zoom * scale, MIN_ZOOM, MAX_ZOOM);
 
-      setTargetZoom(newZoom);
-    };
-
-    containerRef.current?.addEventListener("wheel", handleWheel, {
-      passive: false,
-    });
-
-    return () =>
-      containerRef.current?.removeEventListener("wheel", handleWheel);
-  }, [targetZoom]);
-
-  useEffect(() => {
-    if (zoom === targetZoom) return;
-
-    const glideZoom = () => {
-      const diff = (targetZoom - zoom) / targetZoom;
-      const scale = Math.exp(
-        Math.sign(diff) * SCALE_FACTOR * Math.abs(diff) * 2,
-      );
-      const newZoom = clamp(
-        zoom * scale,
-        diff > 0 ? MIN_ZOOM : targetZoom,
-        diff < 0 ? MAX_ZOOM : targetZoom,
-      );
-
+      // Clamping the zoom means the actual scale may be different.
       const effectiveScale = newZoom / zoom;
 
       setOffset((prevOffset) => {
@@ -348,10 +317,13 @@ export default function CanvasView() {
       setZoom(newZoom);
     };
 
-    const interval = setInterval(glideZoom, 8);
+    containerRef.current?.addEventListener("wheel", handleWheel, {
+      passive: false,
+    });
 
-    return () => clearInterval(interval);
-  }, [zoom, targetZoom, mouseOffsetDirection]);
+    return () =>
+      containerRef.current?.removeEventListener("wheel", handleWheel);
+  }, [zoom]);
 
   /********************************
    * PANNING FUNCTIONALITY.       *
@@ -562,8 +534,8 @@ export default function CanvasView() {
           id="canvas-pan-and-zoom"
           ref={canvasPanAndZoomRef}
           style={{
-            transform: `translate(${offset.x}px, ${offset.y}px)`,
-            scale: zoom,
+            transform: `matrix(${zoom}, 0, 0, ${zoom}, ${offset.x * zoom}, ${offset.y * zoom})`,
+            transition: "transform 0.1s ease-out",
           }}
         >
           <ReticleContainer
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -320,6 +320,7 @@ export default function CanvasView() {
         return clampOffset(addPoints(scaledOffsetDiff, prevOffset));
       });
 
+      // Use css transition for zoom due to macOS trackpads having high polling rates resulting in laggy zooming if implemented differently
       setTransitionDuration(ZOOM_DURATION);
       setZoom(newZoom);
     };
@@ -423,6 +424,7 @@ export default function CanvasView() {
     [handlePan, handlePointerUp, changeGlobalSelectStyle],
   );
 
+  // Could potentially get replaced by a transition animation with ease, however this will work on Safari
   useEffect(() => {
     const decayVelocity = () => {
       if (velocity.x < 0.1 && velocity.y < 0.1) return;
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -300,12 +300,15 @@ export default function CanvasView() {
         mousePositionOnCanvas,
       );
 
+      if (event.deltaY === 0) return;
       // Inclusion of deltaY in calculation to account for different polling rate devices
-      const scale = zoom * (1 + SCALE_FACTOR * -event.deltaY);
-      const newZoom = clamp(scale, MIN_ZOOM * initialZoom, MAX_ZOOM);
+      // Could try logarithmic scale for smoother increments
+      const scale = 1 + SCALE_FACTOR * Math.max(Math.abs(event.deltaY), 1);
+      const newZoom = event.deltaY > 0 ? zoom / scale : zoom * scale;
+      const clampedZoom = clamp(newZoom, MIN_ZOOM * initialZoom, MAX_ZOOM);
 
       // Clamping the zoom means the actual scale may be different.
-      const clampedScale = newZoom / zoom;
+      const clampedScale = clampedZoom / zoom;
 
       setOffset((prevOffset) => {
         // Calculate the of the mouse position relative to the center of the canvas in the container
@@ -320,7 +323,7 @@ export default function CanvasView() {
 
       // Use css transition for zoom due to macOS trackpads having high polling rates resulting in laggy zooming if implemented differently
       setTransitionDuration(ZOOM_DURATION);
-      setZoom(newZoom);
+      setZoom(clampedZoom);
     };
 
     containerRef.current?.addEventListener("wheel", handleWheel, {
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -135,6 +135,13 @@ const MAX_ZOOM = 100;
 const MIN_ZOOM = 0.5;
 
 const ZOOM_DURATION = 0.1;
+// Transition animation on canvas pan and zoom is blurred on Safari and needs to be disabled.
+// If the user spoof their user agent, this is not my problem.
+// Bug in question https://bugs.webkit.org/show_bug.cgi?id=27684
+const IS_SAFARI =
+  navigator.userAgent.includes("Safari/") &&
+  !navigator.userAgent.includes("Chrome/") &&
+  !navigator.userAgent.includes("Chromium/");
 
 // This is to avoid weird business with the reticle not sizing properly
 const RETICLE_ORIGINAL_SCALE = 10;
@@ -543,7 +550,8 @@ export default function CanvasView() {
           ref={canvasPanAndZoomRef}
           style={{
             transform: `matrix(${zoom}, 0, 0, ${zoom}, ${offset.x * zoom}, ${offset.y * zoom})`,
-            transition: `transform ${transitionDuration}s ease-out`,
+            transition:
+              IS_SAFARI ? "" : `transform ${transitionDuration}s ease-out`,
           }}
         >
           <ReticleContainer
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -381,7 +381,7 @@ export default function CanvasView() {
   );
 
   /* A bit heavy handed, but it prevents elements outside of the canvas from being selected during panning */
-  const changeGlobalSelectStyle = useCallback((style: "none" | "auto") => {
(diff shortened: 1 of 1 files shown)


hwan513:
The duplicate commit message was a mistake I made when performing `jj split`

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -36,7 +29,9 @@ const CanvasContainer = styled("div")`
   place-content: center;
   place-items: center;
   /* Fixes blurry canvas in Safari when canvasImage overlaps with overflow, don't ask why */
-  -webkit-transform: translate3d(0, 0, 0);
+  transform: translate3d(0, 0, 0);
+  /* Don't handle panning and zooming with browser */
+  touch-action: none;
```
jaskfla, line 34: I was just thinking about adding this! 😆 

Where did _you_ find out about this? :unamused:

Great stuff

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -130,9 +125,19 @@ function getDefaultZoom(
   return scale;
 }
 
-const SCALE_FACTOR = 0.2;
+const SCALE_FACTOR = 0.002;
```
jaskfla, line 128: Intentional?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -130,9 +125,19 @@ function getDefaultZoom(
...
 const MAX_ZOOM = 100;
-const MIN_ZOOM = 0.5;
+const MIN_ZOOM = 0.9;
+
+const ZOOM_DURATION = 0.1;
```
jaskfla, line 132: Variable name could do with a unit

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -130,9 +125,19 @@ function getDefaultZoom(
...
+const PAN_DECAY = 0.75;
+// Transition animation on canvas pan and zoom is blurred on Safari and needs to be disabled.
+// If the user spoof their user agent, this is not my problem.
+// Bug in question https://bugs.webkit.org/show_bug.cgi?id=27684
+const IS_SAFARI =
+  navigator.userAgent.includes("Safari/") &&
+  !navigator.userAgent.includes("Chrome/") &&
+  !navigator.userAgent.includes("Chromium/");
```
jaskfla, line 140: Should this use a case-insensitive regex pattern, or are we confident about casing?

My preference is to use `includes`, but if we start having to chain with `toLocaleUppercase()` then just `/pattern/i` it

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -265,6 +269,14 @@ export default function CanvasView() {
    * ZOOMING FUNCTIONALITY.       *
    ********************************/
 
+  const getRelativeMousePosition = useCallback(
+    (element: HTMLElement, event: MouseEvent) => {
+      const rect = element.getBoundingClientRect();
+      return { x: event.clientX - rect.left, y: event.clientY - rect.top };
+    },
+    [],
+  );
```
jaskfla, line 278: Looks like this could be a plain function at the top-level scope?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -273,85 +285,53 @@ export default function CanvasView() {
...
-        diff < 0 ? MAX_ZOOM : targetZoom,
+      // The mouse position's origin is in the top left of the container.
+      // Converts this to the offset from the center of the visual container
+      const mouseOffset = diffPoints(
+        { x: elem.offsetWidth / 2, y: elem.offsetHeight / 2 },
+        mousePositionOnCanvas,
       );
 
-      const effectiveScale = newZoom / zoom;
+      if (event.deltaY === 0) return;
```
jaskfla, line 302: Is `=== 0` safe here or does it need to be some sort of epsilon value?

(I have no idea if this is an int value or a number that could technically be `0.00000000000123`)

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -381,116 +361,79 @@ export default function CanvasView() {
...
 
-  const handleMouseMove = useCallback(
-    (event: MouseEvent): void => {
-      const diff = { x: event.movementX, y: event.movementY };
-      setVelocity({ x: diff.x, y: diff.y });
-      updateOffset(diff);
+  const handlePan = useCallback(
+    (event: PointerEvent): void => {
+      // Disable transitions while panning
+      setTransitionDuration(0);
```
jaskfla, line 373: Is this worth it? What transition are we enabling/disabling here?

Is it possible to set `transition-property` more conservatively instead?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -381,116 +361,79 @@ export default function CanvasView() {
...
-    containerRef.current.removeEventListener("mousemove", handleMouseMove);
-    containerRef.current.removeEventListener("mouseup", handleMouseUp);
-    containerRef.current.removeEventListener("mouseleave", handleMouseUp);
-  }, [handleMouseMove]);
+  /* A bit heavy handed, but it prevents elements outside of the canvas from being selected during panning */
+  const changeGlobalSelectStyle = useCallback((style: "none" | "auto") => {
+    document.body.style.userSelect = style;
+    // only -webkit-user-select style exists on Safari: https://caniuse.com/mdn-css_properties_user-select
+    document.body.style.webkitUserSelect = style;
+  }, []);
```
jaskfla, line 388: Appreciate the comments here

Could it be done only with CSS?

```css
body:has(&:active) {
  --webkit-user-select: none;
  user-select: none;
}
```

If not, just switch `auto` to `initial` probably

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -381,116 +361,79 @@ export default function CanvasView() {
...
+   * cause the canvas to pan.
    */
-  const handleStartTouchPan = useCallback(
-    (event: React.TouchEvent<HTMLDivElement>): void => {
-      if (!containerRef.current) return;
-
-      const touchCount = event.touches.length;
+  const handlePointerDown = useCallback(
+    (event: React.PointerEvent<HTMLDivElement>) => {
+      const elem = event.currentTarget;
```
jaskfla, line 415: Great refactor. Much prefer this readability

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -562,8 +497,9 @@ export default function CanvasView() {
           id="canvas-pan-and-zoom"
           ref={canvasPanAndZoomRef}
           style={{
-            transform: `translate(${offset.x}px, ${offset.y}px)`,
-            scale: zoom,
+            transform: `matrix(${zoom}, 0, 0, ${zoom}, ${offset.x}, ${offset.y})`,
+            transition:
+              IS_SAFARI ? "" : `transform ${transitionDuration}s ease-out`,
```
jaskfla, line 502: 1. Try this with the quad or quart variations of ease out from https://easings.net. (If you do use it, set it as `--ease-out-quart` or whatever at `:root` in `core.css`)
2. I think you can use the ``` css`` ``` tagged template literal here (`import { css } from '@mui/style'` or something like that)
3. `null` is probably more semantically appropriate than empty string?

jaskfla (commented):
Neat stuff! Looks great—just haven’t had the chance to run the code yet. Hence not approved

Maybe Wednesday I’ll try get around to it

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Through Google searching. The one and only skill

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Rules were taken from https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent. From my testing it seems to be consistent.

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Duration would be in seconds—I'll add that in. Can't think of would Zoom would be, in this case its tied to 90% of the initial canvas width, so it does end up being different from what `MAX_ZOOM`'s unit is.

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Disables the ease transition on `canvas-pan-and-zoom` during panning as separate react code is being used to give the panning "inertia". 

Also I gave up on trying to implement it with a transition)
Also Safari...

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: 1. Played around with `ease-out-quart`. Seems to be nice, but I can't really tell the difference. Maybe a future thing.
2. Not too sure how to approach this
3. `null` not working was the reason I chose `""`. Turns out `undefined` is the semantically correct manner to do this (I somehow missed this in the function signature).

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: I could have the transition duration set in the CSS and directly boolean check it away (like with `IS_SAFARI`. I initially had the transition duration state in case I wanted a different duration for the panning animation. This would be set at the `pointerup` event by calculating a final destination and then easing to it.

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Ah, this would have be be on this React component, where `&` is `styled`’s alias for the generated className of the current component

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Don’t worry too much about point 2 since it’s an inline style. Can discuss next time it crops up with a `styled` component

`undefined` probably makes sense, then

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: That works, updated.

packages/frontend/src/components/canvas/CanvasView.tsx
pumbas600: Can we move this up so that there's only one early return:

```ts
if (event.deltaY === 0 || !(elem instanceof HTMLElement)) return;
```

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -165,18 +183,18 @@ export default function CanvasView() {
   );
   const [isLoading, setIsLoading] = useState(false);
   const [zoom, setZoom] = useState(1);
+  const [initialZoom, setInitialZoom] = useState(1);
   const [offset, setOffset] = useState(ORIGIN);
   const [velocity, setVelocity] = useState<Point>({ x: 0, y: 0 });
   const [controlledPan, setControlledPan] = useState(false);
-  const [targetZoom, setTargetZoom] = useState(1);
-  const [mouseOffsetDirection, setMouseOffsetDirection] = useState(ORIGIN);
+  const [transitionDuration, setTransitionDuration] = useState(0);
```
pumbas600, line 190: Can we rename this to be `isCssTransitionEnabled` because we toggle between 0 and `ZOOM_DURATION` as if it's a boolean already and I think this will make it more clear.

packages/frontend/src/components/canvas/CanvasView.tsx
pumbas600: Actually, we only want the transition on zoom right? 

I noticed you changed `transform` to use a `matrix` to include the zooming, but is there any issue to going back to what we had and using a `scale`?

If there aren't any issues with that, then we can only add a `transition` for `scale`, avoiding the need for this variable entirely.

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -130,9 +131,27 @@ function getDefaultZoom(
...
+
+const ZOOM_DURATION = 0.1;
+const PAN_DECAY = 0.75;
+// Transition animation on canvas pan and zoom is blurred on Safari and needs to be disabled.
+// If the user spoof their user agent, this is not my problem.
+// Bug in question https://bugs.webkit.org/show_bug.cgi?id=27684
+const IS_SAFARI =
+  navigator.userAgent.includes("Safari/") &&
+  !navigator.userAgent.includes("Chrome/") &&
+  !navigator.userAgent.includes("Chromium/");
```
pumbas600, line 154: This doesn't work on the server, so we need to add a check.

```suggestion
const IS_SERVER = typeof window === "undefined";

const ZOOM_DURATION = 0.1;
const PAN_DECAY = 0.75;
// Transition animation on canvas pan and zoom is blurred on Safari and needs to be disabled.
// If the user spoof their user agent, this is not my problem.
// Bug in question https://bugs.webkit.org/show_bug.cgi?id=27684
const IS_SAFARI =
  !IS_SERVER &&
```

pumbas600 (changes requested):
The zooming feels very smooth, way better than before!

One thing I've noticed is that on Chrome for me, it goes blurry briefly while zooming (Presumably because of the transition?). It's not too bad, but a little annoying

Other than that, I've just left a few minor comments :)

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Seems like that could be an `isServer()` util function?

In any case, I would _also_ extract `isSafari` into its own helper

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: @hwan513 Did this turn out to be necessary, or did you end up doing this with `:active` on a styled element?

hwan513: I'm like 3 stacked PR's deep and I swear I changed the behaviour at some point in time.

hwan513: Its definitely in this current PR (5afc033), I just forget to remove the the state variables.

hwan513: Just deleted my previous comments (got a bit lost being 3.5 stack PRs deep).

I did try setting it for when it was `:not(:active)` which wheel event would fall under but not mouse event. The only problem with that was that the current panning inertia implementation would be affected by this transition. And when I first tried changing the panning inertia implementation to use transition, it felt weird to get right for some reason (also wouldn't work on Safari).

I'll do some renaming on the variables.

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Does this still run on the server when `"use client"` has been defined? Not sure if `IS_SAFARI` is being used outside of this file yet.

hwan513:
> Actually, we only want the transition on zoom right?
>
>  I noticed you changed transform to use a matrix to include the zooming, but is there any issue to going back to what we had and using a scale?
>
> If there aren't any issues with that, then we can only add a transition for scale, avoiding the need for this variable entirely.

@pumbas600. Can't reply to the comment. There were 2 issues I think matrix was trying to fix. First I'm about 50% sure that using `matrix` was a fix to the blurring issue (not too sure anymore though), secondly, the exact "canvas subpixel" zooming behaviour requires all the transforms to be done simultaneously (hence why the equations are now different), and the offset is now in absolute pixel values instead of (`offsetX / zoom`, i.e. the virtual pixel of the canvas (0-99 in the demo canvas) that it was before.

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: @pumbas600 also refer to https://github.com/UOA-CS732-SE750-Students-2024/project-group-golden-giraffes/pull/262#issuecomment-2746171717.

hwan513:
> The zooming feels very smooth, way better than before!
> 
> One thing I've noticed is that on Chrome for me, it goes blurry briefly while zooming (Presumably because of the transition?). It's not too bad, but a little annoying. It's fine on Firefox though.
> 
> Other than that, I've just left a few minor comments :)

~~If it is going blurry on Chrome then it must mean that the transition bug seems to be applying to Webkit in general and not just Safari. What Chrome version (platform etc.) are you using though, I was under the impression that Blink didn't have this bug, and I can't seem to notice it in my own testing.~~

On second slower testing, I can see it occurring now. Not as bad as how it is on Safari though, so I'm assuming that Webkit, Blink and Gecko all have slightly differing implmentations that is causing the bug. It isn't as noticeable on Chrome though, so I'm happy to keep it enable there.

hwan513:
Misclicked oops.

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Using isZooming as the state variable now

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Updated in [72d989c](https://github.com/UOA-CS732-SE750-Students-2024/project-group-golden-giraffes/pull/262/commits/72d989cb8f7245ed89db54081d0ddbe4576de6b0)

packages/frontend/src/components/canvas/CanvasView.tsx
hwan513: Updated in ba59e15

sou323 (approved):

All looks good to me!

PR #263, opened 2025-03-17
fix(frontend): add isLaunching to address initial canvas zoom incorrectness

hwan513:
Technically, the `handleImageLoad` can be refactored into a single useEffect, however this may cause desync between the rendered canvas (loaded image), and the set dimensions initial zoom, as the useEffect will complete before the image is loaded from the backend.

### DIFF
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -93,7 +93,7 @@ const InviteButton = styled(Button)`
 `;
 
 const CanvasImageWrapper = styled("div", {
-  shouldForwardProps: (prop) => !['isLaunching', 'isLoading'].includes(prop)
+  shouldForwardProps: (prop) => !["isLaunching", "isLoading"].includes(prop),
 })<{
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -81,7 +81,10 @@ const InviteButton = styled(Button)`
   }
 `;
 
-const CanvasImageWrapper = styled("div")<{ isLoading: boolean }>`
+const CanvasImageWrapper = styled("div")<{
+  isLoading: boolean;
+  isLaunching: boolean;
+}>`
```
jaskfla, line 87:
```suggestion
const CanvasImageWrapper = styled("div", {
  shouldForwardProps: (prop) => !['isLaunching', 'isLoading'].includes(prop)
})<{
  isLoading: boolean;
  isLaunching: boolean;
}>`
```

pumbas600:
Checking is this ready for review?

hwan513:
> Checking is this ready for review?

Yes, its set as draft as it is a stacked PR

PR #267, opened 2025-03-18
feat(frontend): mobile panning pinch zoom reloaded

hwan513:
The long anticipated sequel to the hit PR `feat: mobile panning` comes `feat(frontend): mobile panning pinch zoom reloaded`

In this new instalment, the newly refactored PointerEvent handlers used to panning the Canvas has discovered a new passion, handling multi-touch pinch interactions in mobile. Follow their journey of self discovery, through multiple helper function refactors and mathematical escapades, to ultimately achieve the goal of mobile support.

<details> 
  <summary>Spoiler Alert!!!</summary>
  The sneaky clampOffset and clampZoom bugs escaped unscathed and has staged a revolt. Find out how the story ends in the final part of the trilogy `fix(frontend): mobile panning pinch zoom clamp adjustments revolutions`
</details>

PR #268, opened 2025-03-23
feat(frontend): mobile slideable drawer

hwan513:
Tested on an iOS simulator on iPhone16

Adds a drawer that slides between three stages of openedness.
Drawer can be slid open using the handle and also the tab bar elements (would be a bit more hacky [even more than currently] to support all non-scrollable elements such as the element for who placed the most recent pixel).

Moves the project blurple invite link onto the top right when in mobile.

Of the 3 stages of openedness:

1. Closed: only shows handle, and the Place and Look tabs
2. Half: 50%: shows everything except the Color and Discord placement commands. (Uses a verbose mode of the place pixel button that includes the color in the string. Looks a bit weird due to the proximity of the place pixel button and the colors.
3. Open: Shows everything. The transition from half to full is a bit jarring as the color indicator and the discord placement command appears out of nowhere. Specifically positioned so that the top of the drawer sits right under the project blurple invite.

Couldn't find a method to hide the toolbar on iOS Safari that worked.

Dependant on 2 PRs, which I can't really stack, which is why the commit count is a bit off.

### DIFF
--- a/packages/frontend/src/components/action-panel/ActionPanel.tsx
+++ b/packages/frontend/src/components/action-panel/ActionPanel.tsx
@@ -10,8 +10,7 @@ import { PixelInfoTab, PlacePixelTab } from "./tabs";
 const Wrapper = styled("div")`
   --padding-width: 1rem;
   background-color: var(--discord-legacy-not-quite-black);
-  border-start-start-radius: var(--card-border-radius);
-  border-start-end-radius: var(--card-border-radius);
+  border-radius: var(--card-border-radius);
   border: var(--card-border);
   display: grid;
   gap: 0.5rem;
@@ -24,8 +23,9 @@ const Wrapper = styled("div")`
     border-radius: calc(var(--card-border-radius) - var(--padding-width));
   }
 
-  ${({ theme }) => theme.breakpoints.up("md")} {
-    border-radius: var(--card-border-radius);
+  ${({ theme }) => theme.breakpoints.down("md")} {
+    border-radius: 0;
+    border: initial;
   }
 `;
 
--- a/packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
@@ -9,6 +9,7 @@ import {
 } from "@/contexts";
 import { usePalette } from "@/hooks";
 import { decodeUserGuildsBase64 } from "@/util";
+import { useCallback, useEffect, useState } from "react";
 import { DynamicAnchorButton, PlacePixelButton } from "../../button";
 import { InteractiveSwatch } from "../../swatch";
 import { Heading } from "../ActionPanel";
@@ -16,7 +17,6 @@ import { ScrollBlock, TabBlock } from "./ActionPanelTabBody";
 import { ActionPanelTabBody } from "./ActionPanelTabBody";
 import BotCommandCard from "./BotCommandCard";
 import ColorInfoCard from "./SelectedColorInfoCard";
-import { useCallback, useEffect, useState } from "react";
 
 const ColorPicker = styled("div")`
   display: grid;
--- a/packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
@@ -16,6 +16,7 @@ import { ScrollBlock, TabBlock } from "./ActionPanelTabBody";
 import { ActionPanelTabBody } from "./ActionPanelTabBody";
 import BotCommandCard from "./BotCommandCard";
 import ColorInfoCard from "./SelectedColorInfoCard";
+import { useCallback, useState } from "react";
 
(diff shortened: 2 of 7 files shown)


packages/frontend/src/app/signin/page.tsx
```diff
@@ -8,7 +8,7 @@ import Link from "next/link";
 
 const Background = styled("div")`
   align-items: center;
-  block-size: 100vh;
+  block-size: 100dvh;
```
jaskfla, line 11:
```suggestion
  block-size: 100dvb;
```

`b` for block`. `h` for `height`. Generally doesn’t make sense to mix these

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -66,6 +60,14 @@ const Tab = styled("li")<{ active?: boolean }>`
...
     : ""}
 
+  /*
+  * Workaround for accessibility issue with VoiceOver.
+  * See https://gerardkcohen.me/writing/2017/voiceover-list-style-type.html
+  */
+  &::before {
+    content: "\\200B"; /* zero-width space */
+  }
+
```
jaskfla, line 70: Feel free to remove this entirely. I have a PR coming in the next week that removes the need for this workaround

packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
```diff
@@ -64,6 +65,20 @@ export default function PlacePixelTab({
...
+  const PlacePixelTabBlockRef = useCallback((elem: HTMLDivElement | null) => {
+    const remPixels = Number.parseFloat(
+      getComputedStyle(document.documentElement).fontSize,
+    );
+    if (!elem) return;
+    const resizeObserver = new ResizeObserver((entries) => {
+      const height = entries[0].target.clientHeight;
+      setIsLarge(height > remPixels * 20);
+    });
+    resizeObserver.observe(elem);
```
jaskfla, line 80: Don’t love use of observer here. Can `isLarge` just store the result of a `useMediaQuery`?

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
```diff
@@ -0,0 +1,205 @@
...
+`;
+
+const HandleWrapper = styled("div")`
+  cursor: grab;
+  padding-block: 0.8rem 0;
+  padding-inline: auto;
+  display: flex;
+  place-content: center;
+  /* Hacky, overwrites the flex-grow from DrawerWrapper */
+  flex-grow: 0 !important;
```
jaskfla, line 49: Huh, generally not recommended to use `flex-grow` on its own. (Not sure why, but MDN says to use `flex` shorthand

In any case, remember that `unset` and `initial` exist. One of those is probably more semantically appropriate here

packages/frontend/src/components/action-panel/ActionPanel.tsx
hwan513: Bear in mind that this PR branches off two feature branches #250 and #267. Github isn't handling the diffs to the best of its abilities, This is actually a change in #250 that is inherited into here. Should hopefully not show up once #250 is merged.

packages/frontend/src/app/signin/page.tsx
hwan513: This should also belong in #250 I think. If we can get it merged earlier, it should hopefully fix some of these issues. There is a weird dependency graph going on from this.

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
hwan513: That kinda makes sense, given that I usually don't have to set it on its own.

For some reason, both flex grow directives have become unneeded in my testing. Definitely needed it when I begun implementations. Might warrant further testing on your end.

packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
hwan513: `isLarge` is tied to the height of the element and while media queries only respond to viewport changes right?
I originally tried a purely CSS implementation with container queries which I couldn't figure out how to get it to work. 
Next I tried to `useRef` to little success.

My biggest gripe with the current `isLarge` implementation is just how the hidden elements would just pop up out of nowhere with no animation. `display: none` doesn't have animations tied to it, meanwhile setting it to `visibility: hidden; height: 0, padding: 0; margin: 0` isn't the most accurate since flex box gap still exists (there probably is some CSS thing I missing though), but tying it to an animation (setting the transition to encompass maybe more behaviours) would be pretty nice.`

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
hwan513: Flex grow on its own became unneeded.

sou323 (approved):

All looks good to me!

PR #271, opened 2025-05-08
fix: use `Math.ceil` not `Math.round` when displaying cooldown

pumbas600:
This should reduce the number of times users accidentally try and place a pixel before they're allowed to.

I don't have time, unfortunately, but what we should be doing is throwing a 400 on the server with the number of milliseconds left. In our error block, we should wait for that number of milliseconds and then automatically try again. Providing the ms left rather than the timestamp again is important because sometimes you can have clients whose clock is out of sync with our server, which is why this happens in the first place.


hwan513 (approved):
Erm actually ☝️🤓, it should probably be a `fix:`

PR #272, opened 2025-05-08
fix(frontend): fix decay velocity inifinite recursion bug

hwan513:
Old method of `decayVelocity` had a bug that called the `decayVelocity` function every pre-render. Used ChatSkibidi to fix it.

To test the fixes, add `console.log()` into the interval, the useEffect, and before the `clearInterval`.


jaskfla:
Something smells off with the Effect, but it’s an improvement so approving as is

THank 🙏

hwan513:
> Something smells off with the Effect, but it’s an improvement so approving as is
> 
> THank 🙏

I mean I feel like you can just easily create the interval setup code directly within the `handlePointerUp` function bypassing the need for the use effect.

PR #275, opened 2025-05-08
fix(frontend): eyeball styling fix for mobile leaderboard

hwan513:
Very scuffed fix for displaying leaderboard in mobile. I think it breaks when it sees something smaller than an iPhone size though.

image


sou323 (approved):
looks good ty

PR #276, opened 2025-05-08
fix: use cooldown time directly instead of returning a timestamp

### LINKED ISSUE #273: Use remaining cooldown in milliseconds instead of timestamp
Mentioned here: #271. Desync in a clients local clock will result in inaccurate cooldown times displayed when placing pixels.

hwan513:
Should fix #273, can't really test it since I can't desync my backend server clock with the frontend, I think. The placement thing still works though.
Doesn't really address the error 400 mentioned in the originating PR however though. also can't really rely on test cases because of, reasons...


sou323 (approved):
lgtm

pumbas600 (approved):
Nice, this should definitely be more reliable

PR #277, opened 2025-05-09
fix(frontend): use pixel overlay together with offscreencanvas

hwan513:
Fixes the performance issues experienced by users when live updating places too many pixels.

Generally start experiencing lag around 100+ overlaid pixels. Solution is the prevent the accumulation of over 100 pixels. There is a variable to adjust the threshold in case lower-end devices can't handle it. See: `pixelOverlayThreshold`.

commit f4665c94c7ae237b518dda7032e8a6a2aafaa786 adds a loop to spoof 1000 places per real placed pixel. 

commit 7fcb56e302ad24b2f633125751d7ccc9a244944b implements a first fix that updates an offscreen canvas which is then converted into a png. Performance testing of 1000 places per pixels is also implemented here.

Use commit 94697ff0cc957d09c3566c41c25966490d9c43b4 to test the performance of the last fix, which combines both of the previous methods above. Only spoofs 100 pixels due to the limitations of React (explained in the code comments and documentation).


sou323 (approved):
looks great! tysm

PR #280, opened 2025-05-09
fix(frontend): inconsistent panning based on browser

hwan513:
Especially big reason on android firefox due for some reason.

Issue is caused by the usage of PointerEvent.movementX and PointerEvent.movementY, which is inconsistent across the different browsers (Safari, Android Firefox). Fix includes a custom movementX and movementY calculation.

https://github.com/w3c/pointerlock/issues/42
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX


sou323 (approved):
lgtm

PR #281, opened 2025-05-12
fix(frontend): add `display: content` to SlideableDrawer

hwan513:
Debugged with ChatGPT, comment is my own though. Fixes scroll overflow not occurring when viewing pixel history on desktop.


Rocked03 (approved):
Haha lmao lgtm

sou323 (approved):
Pretend that above review was from the correct account

PR #282, opened 2025-05-12
feat(backend): update leaderboard query to exclude blocklisted users

stijnvdkolk:
Was discussed with @Rocked03


sou323:
Currently on pause, will revisit at event end

Rocked03 (approved):

PR #296, opened 2026-01-21
fix: git ignore prisma migrations

volcanofr:
Added a `.gitingore` line about backend's Prisma migrations made via [`prisma migrate`](https://www.prisma.io/docs/orm/prisma-migrate/getting-started)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 # Generated code
 **/node_modules/
 **/.pnp
+**/migrations/
 .pnp.js
 .yarn/install-state.gz
 


stijnvdkolk:
Why would we want this?

volcanofr:
> Why would we want this?

It is generated via `prisma generate`, and is not included in the repo.

stijnvdkolk:
> > Why would we want this?
> 
> It is generated via `prisma generate`, and is not included in the repo.

`prisma generate` doesn't make the migrations folder 😅 I'm not aware of the reasoning to not commit the migrations (I think they wrote them manually and just didnt put them in as a migration (yet))

volcanofr:
Oh, how did I made that folder then? x)
I don't remember 💀

PR #229, opened 2024-05-11
feat: coordinates in url (OLD)

sou323:
- can now link direct to coords through `?c=canvasid&x=x&y=y&z=zoom`
- comes with copy link button in look tab

PR #297, opened 2026-01-22
feat: added leaderboard pagination

### LINKED ISSUE #404: Leaderboard pagination

volcanofr:
* updated backend to support pagination in leaderboard API
  + total number of users
  + skip n-1 first users

* modified frontend to fetch and display paginated leaderboard data

| image | image | image |
| --- | --- | --- |

It does not includes query parameter yet, I'm waiting for #229 to be merged first.


pumbas600 (changes requested):
This is looking awesome! Just some minor changes to make it semantically make more sense.

I've not had a look around locally, so I've not tested any of the UI changes yet.

packages/backend/src/services/statisticsService.ts
```diff
@@ -76,13 +78,22 @@ export async function getLeaderboard(
     },
   });
```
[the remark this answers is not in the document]
pumbas600: Just a note, if you see that performance starts to be an issue when you're on a very high page (Not that I think we'll have enough users for that to be an issue), you could use [cursor-based pagination](https://medium.com/better-programming/understanding-the-offset-and-cursor-pagination-8ddc54d10d98). The downside is you wouldn't be able to have the numbered list of pages, but you could have next/prev/first buttons.

I'm happy to explain this in more detail if you want to go down that route though!

volcanofr:
Before:

image

After:

image

packages/types/src/pagination.d.ts
```diff
@@ -0,0 +1,6 @@
+export interface Pagination<T> {
```
jaskfla: Lovely utility type 👌

Opinionated, but perhaps `Paginated<T>` would be a more natural name?

cc @pumbas600, who I realise suggested this initially

packages/frontend/src/hooks/queries/useLeaderboard.tsx
```diff
@@ -6,18 +6,20 @@ import axios from "axios";
 import config from "@/config";
 import { CanvasInfo, LeaderboardRequest } from "@blurple-canvas-web/types";
 
-export function useLeaderboard(canvasId: CanvasInfo["id"]) {
+export function useLeaderboard(canvasId: CanvasInfo["id"], page = 1, size = 10) {
```
jaskfla: Suuper nitpicky, but I think `{ page, size }` really belong together as one unit. That said, I don’t feel _too_ strongly about this

packages/frontend/src/hooks/queries/useLeaderboard.tsx
```diff
@@ -6,18 +6,20 @@ import axios from "axios";
...
     const response = await axios.get<LeaderboardRequest.ResBody>(
       `${config.apiUrl}/api/v1/statistics/leaderboard/${canvasId}`,
+      { params: { page, size } },
     );
     return response.data;
   };
 
   return useQuery({
-    queryKey: ["leaderboard", canvasId],
+    queryKey: ["leaderboard", canvasId, page, size],
```
jaskfla: Similar comment as above, `size` isn’t really a sub-scope of `page`. Wrap ’em in an object and TanStack Query will take care of it the way we expect

```suggestion
    queryKey: ["leaderboard", canvasId, { page, size }],
```

packages/frontend/src/hooks/queries/useLeaderboard.tsx
jaskfla, line 1: I’d be surprised if pages get booted out of the query client cache from just jumping between pages, but nevertheless probably makes sense to use [`keepPreviousData`](https://tanstack.com/query/latest/docs/framework/react/guides/paginated-queries#better-paginated-queries-with-placeholderdata) here 🧐

packages/frontend/src/app/leaderboard/Leaderboard.tsx
```diff
@@ -45,7 +52,10 @@ const UserCell = styled("td")`
 `;
 
 const Username = styled("p")`
-  max-inline-size: 22rem;
+  max-inline-size: min(22rem, 50vw);
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
 `;
```
jaskfla: Nice catch here! Though I might ask that this fix be descoped from this PR

Truncating is arguably more of an issue than the current (admittedly _uglier_ failure), as accessibility issues for sighted users arise when use fixed dimensions (`height: 60px`) and truncate text (`text-overflow: ellipsis`). Considering people may have browser settings/extensions or user style sheets that alter anything viewport or font size to what font is even rendered, it’s important that we allow components to reflow content as much as reasonably possible

The offset border on the avatar is certainly an eyesore (pretty sure it’s my fault too 😓); but I think it’s less bad than lopping off text. I think we can come with a more graceful solution with a similar number of CSS line changes 🙂

jaskfla (changes requested):
Thanks for much for adding this feature, and for already addressing @pumbas600’s first round of comments 🙏

Couple nitpicks from me, but primarily actually just asking for the CSS fixes to be descoped for another PR

Looking really solid overall. Appreciate how closely the code style in this PR matches the existing stuff

packages/frontend/src/hooks/queries/useLeaderboard.tsx
volcanofr: I see the why it can be better with an object param rather than 2 independent parameters.
But there is not a lot of parameters and I don't see any case of adding another, separate params are (a little, not huge difference) easier for defaults.

I would personally not change it - but it's up to you, your the maintainer 👍
(also, if we do make it as an object, what's the name of the param? "query"? but that will split meanings on [L19](https://github.com/project-blurple/Canvas-Web/pull/297/changes#r3031430960))
(also in the backend too then?)

packages/types/src/pagination.d.ts
volcanofr: Like [`Pagination<Entry>`](https://github.com/project-blurple/Canvas-Web/pull/297/changes#r3045968356)? Or did I not understood 😂

packages/frontend/src/app/leaderboard/Leaderboard.tsx
```diff
@@ -45,7 +52,10 @@ const UserCell = styled("td")`
...
 
 const Username = styled("p")`
-  max-inline-size: 22rem;
+  max-inline-size: min(22rem, 50vw);
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
```
volcanofr:
```suggestion
  max-inline-size: 22rem;
```

packages/types/src/pagination.d.ts
```diff
@@ -0,0 +1,6 @@
...
+  total: number;
+  page: number;
+  size: number;
+  entries: T[];
+}
```
volcanofr:
```suggestion
export interface Pagination<Entry> {
  total: number;
  page: number;
  size: number;
  entries: Entry[];
}
```

packages/types/src/pagination.d.ts
pumbas600: @jaskfla is just suggesting we rename the type :). I think the name he suggested is better

packages/backend/src/services/statisticsService.ts
```diff
@@ -48,15 +48,18 @@ export async function getUserStats(
...
- * of 40.
+ * Retrieves the top `size` (max 40), from the rank `(page - 1) * size`
+ * users on the leaderboard for a canvas.
  */
 export async function getLeaderboard(
   canvasId: CanvasInfo["id"],
+  page = 1,
   size = 10,
-): Promise<LeaderboardEntry[]> {
+): Promise<{ total: number; page: number; size: number; entries: LeaderboardEntry[] }> {
```
pumbas600: Can we use the Paginated type in the return type here

pumbas600 (changes requested):

packages/frontend/src/hooks/queries/useLeaderboard.tsx
jaskfla: Let’s keep as-is, then[^f] 🙏

[^f]: if we do make it as an object, what's the name of the param? — I’d’ve named it `options`

packages/types/src/pagination.d.ts
jaskfla: Addressing this one with https://github.com/project-blurple/Canvas-Web/pull/297#discussion_r3047508947

packages/frontend/src/hooks/queries/useLeaderboard.tsx
jaskfla: Will keep this request even though we’re not changing the hook’s signature—it still makes sense scoping-wise to treat `page` and `size` as a pair

packages/types/src/pagination.d.ts
volcanofr: Oh! Right

volcanofr:
Here we go

jaskfla (approved):
Lovely! I’ll get those merge conflicts resolved and merge in the next couple days

Thanks so much for this 🙏

(Also re-requested review from @pumbas600, but I suspect that’ll just be a rubber stamp situation.)

jaskfla:
Ah, darn, looks like one more conflict

@volcanofr When you get a spare moment, can you please sync your fork? 🙏

packages/backend/src/services/statisticsService.ts
```diff
@@ -48,15 +49,18 @@ export async function getUserStats(
...
+ * Retrieves the top `size` (max 40), from the rank `(page - 1) * size`
+ * users on the leaderboard for a canvas.
  */
 export async function getLeaderboard(
   canvasId: CanvasInfo["id"],
+  page = 1,
   size = 10,
-): Promise<LeaderboardEntry[]> {
+): Promise<Paginated<LeaderboardEntry>> {
+  const take = Math.min(Math.max(size, 1), 40); // Arbitrary maximum
```
pumbas600, line 60: Small thing, but perhaps what we should do is assume `size` is valid, and instead add a `max()` to the validation in `LeaderboardQueryModel`?

packages/frontend/src/app/leaderboard/Leaderboard.tsx
```diff
@@ -112,14 +118,26 @@ function leaderboardRecordToTableRow(user?: LeaderboardEntry): JSX.Element {
...
 
 export default function Leaderboard() {
   const { canvas } = useCanvasContext();
-  const { data: leaderboard = [], isLoading: leaderboardIsLoading } =
-    useLeaderboard(canvas.id);
+  const [page, setPage] = useState(0);
+  const {
+    data: { total, page: currentPage, size, entries: leaderboard } = {
+      total: undefined,
+      page: page + 1,
```
pumbas600: We seem to always use `page + 1`? Why not make `page` one-based so that it's consistent with:
1. What we're sending to the API.
2. What we're showing to the user.

pumbas600 (changes requested):
This is looking awesome, just two really small things (Sorry!)

packages/backend/src/services/statisticsService.ts
jaskfla: Current behaviour is more graceful imo. Doesn’t penalise the consumer

Be conservative in what you do; be liberal in what you accept

volcanofr: If there are some crazy people who send manual requests to the backend, it's always better to not have errors (because of negatives).

pumbas600: I don't know, I kind of like when APIs provide helpful errors if I'm using it incorrectly, but I'm not too concerned

pumbas600 (approved):
Looks good! 🚀

Rocked03:
Hi! Can we please merge this?

volcanofr:
Been approved by [pumbas600](https://github.com/pumbas600) and [jaskfla](https://github.com/jaskfla) before a minor change, I believe you can

Rocked03:
~~Our practice is that PRs get merged by their authors, so you get the honour of doing it :)~~

sorry, @jaskfla clarified to me what you two are sorting out, I'll leave you two to figure it out

volcanofr:
I'm not a collaborator, only a community contributor - I don't have access to merge

jaskfla:
> I'm not a collaborator, only a community contributor - I don't have access to merge

Sorry, just a small miscommunication between myself an and Rocked03 there

All sorted! Thank you again for putting this together, and your patience during the review process. Happy with how it turned out 🙌

PR #361, opened 2026-04-07
fix: socket synchronization

volcanofr:
## Describe your changes

Match the Socket `place pixel ${canvasId}` event object on the individual synchronization with the global broadcast.
Now matching the `PlacePixelSocket.Payload` type.

## Screenshots

| Before | After |
| :----: | :---: |
| N/A | N/A |

## Related issues

- None

## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/project-blurple/Canvas-Web?tab=contributing-ov-file)
  > There are none
- [x] I have performed a self-review of my code


packages/backend/src/app/SockerHandler.ts
pumbas600, line 38: Maybe to prevent issues like this in the future, do something like:

```ts
const payload: PlacePixelSocket.Payload = {
  x: pixel.x,
  y: pixel.y,
  rgba: pixel.color.rgba
}

socket.emit(...)
```

pumbas600 (dismissed):
Very good catch!

packages/backend/src/app/SocketHandler.ts
jaskfla, line 1: Thanks for this one! Mind if I trouble you to rename the file too?

Currently Socke-r-Handler 😅

pumbas600: LOL, how did we miss that 😅

packages/backend/src/app/SockerHandler.ts
jaskfla: `emit` method can take a type argument, which might be the ultimate way to go

Can leave that for a separate PR, though 🙏

packages/backend/src/app/SocketHandler.ts
jaskfla: Looks like [I was reviewer](https://github.com/project-blurple/Canvas-Web/pull/202) too 😭

Rocked03 (approved):

PR #369, opened 2026-04-10
feat: Upgraded URL to include canvas, coordinates, and zoom

### LINKED ISSUE #374: URL Search Params

Rocked03:
- When loading URL with search params, canvas loads at the specific coordinates and zoom
- Button to create URL for the currently selected pixel + current zoom
- Canvas persists across reloads via being set as search param

Redoing https://github.com/project-blurple/Canvas-Web/pull/229

### DIFF
--- a/packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
@@ -1,0 +1,40 @@
+import { Tooltip, TooltipProps } from "@mui/material";
+
+interface ActionPanelTooltipProps extends Omit<
+  TooltipProps,
+  "placement" | "slotProps"
+> {
+  children: React.ReactElement;
+}
+
+export default function ActionPanelTooltip({
+  children,
+  ...props
+}: ActionPanelTooltipProps) {
+  return (
+    <Tooltip
+      placement="top"
+      slotProps={{
+        popper: {
+          modifiers: [
+            {
+              name: "offset",
+              options: {
+                offset: [0, -10],
+              },
+            },
+          ],
+        },
+        tooltip: {
+          sx: {
+            bgcolor: "var(--discord-legacy-dark-but-not-black);",
+            boxShadow: "0px 0px 5px rgba(0, 0, 0, 0.25)",
+          },
+        },
+      }}
+      {...props}
+    >
+      {children}
+    </Tooltip>
+  );
+}
--- a/packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
@@ -1,11 +1,6 @@
 import { Tooltip, TooltipProps } from "@mui/material";
 
-interface ActionPanelTooltipProps extends Omit<
-  TooltipProps,
-  "placement" | "slotProps"
-> {
-  children: React.ReactElement;
-}
+type ActionPanelTooltipProps = Omit<TooltipProps, "placement" | "slotProps">;
 
 export default function ActionPanelTooltip({
   children,
@@ -28,7 +23,7 @@ export default function ActionPanelTooltip({
         tooltip: {
           sx: {
             bgcolor: "var(--discord-legacy-dark-but-not-black);",
-            boxShadow: "0px 0px 5px rgba(0, 0, 0, 0.25)",
+            boxShadow: "0px 0px 5px oklch(0 0 0 / 25%)",
           },
         },
       }}
--- a/packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
@@ -1,16 +1,20 @@
 import { PixelHistoryRecord } from "@blurple-canvas-web/types";
 import { styled } from "@mui/material";
-
+import { useState } from "react";
+import { DynamicButton } from "@/components/button";
 import { useCanvasContext } from "@/contexts";
 import { usePixelHistory } from "@/hooks";
+import createPixelUrl from "@/util";
 import { Heading } from "../ActionPanel";
 import {
   ActionPanelTabBody,
   ScrollBlock,
   TabBlock,
 } from "./ActionPanelTabBody";
+import ActionPanelTooltip from "./ActionPanelTooltip";
 import CoordinatesCard from "./CoordinatesCard";
 import PixelHistoryListItem from "./PixelHistoryListItem";
+import { CoordinateLabel } from "./PlacePixelTab";
 
 const PixelInfoTabBlock = styled(TabBlock)`
   grid-template-rows: auto 1fr;
@@ -72,10 +76,32 @@ export default function PixelInfoTab({
   active = false,
   canvasId,
 }: PixelInfoTabProps) {
-  const { coords, adjustedCoords } = useCanvasContext();
+  const { adjustedCoords, canvas, containerRef, coords, zoom } =
+    useCanvasContext();
   const { data, isLoading } = usePixelHistory(canvasId, coords);
 
   const pixelHistory = data?.pixelHistory ?? [];
+
+  const [tooltipIsOpen, setTooltipIsOpen] = useState(false);
+  const closeTooltip = () => setTooltipIsOpen(false);
+  const openTooltip = () => setTooltipIsOpen(true);
+
+  const pixelURL =
+    (adjustedCoords &&
+      containerRef.current &&
+      createPixelUrl({
+        canvasId: canvasId,
+        coords: adjustedCoords,
+        pixelWidth: Math.min(
+          containerRef.current?.clientWidth / zoom,
+          canvas.width,
+        ),
+        pixelHeight: Math.min(
+          containerRef.current?.clientHeight / zoom,
+          canvas.height,
+        ),
+      })) ??
+    "";
 
   return (
     <PixelInfoTabBlock active={active}>
@@ -96,6 +122,32 @@ export default function PixelInfoTab({
           </ActionPanelTabBody>
         </ScrollBlock>
       )}
+      <ActionPanelTabBody>
+        {adjustedCoords && (
+          <ActionPanelTooltip
+            title="Copied"
+            onClose={closeTooltip}
+            open={tooltipIsOpen}
+          >
+            <DynamicButton
+              color={
+                !isLoading && pixelHistory.length > 0 ?
+                  pixelHistory[0].color
+                : null
+              }
+              onAction={() => {
+                openTooltip();
+                navigator.clipboard.writeText(pixelURL);
+              }}
+            >
+              Copy link to share pixel
+              <CoordinateLabel>
+                ({adjustedCoords.x},&nbsp;{adjustedCoords.y})
+              </CoordinateLabel>
+            </DynamicButton>
+          </ActionPanelTooltip>
+        )}
+      </ActionPanelTabBody>
     </PixelInfoTabBlock>
   );
 }
--- a/packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
+++ b/packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
@@ -4,7 +4,7 @@ import { useState } from "react";
 import { DynamicButton } from "@/components/button";
 import { useCanvasContext } from "@/contexts";
 import { usePixelHistory } from "@/hooks";
-import createPixelUrl from "@/util";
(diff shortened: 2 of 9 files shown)


packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
```diff
@@ -0,0 +1,40 @@
+import { Tooltip, TooltipProps } from "@mui/material";
+
+interface ActionPanelTooltipProps extends Omit<
+  TooltipProps,
+  "placement" | "slotProps"
+> {
+  children: React.ReactElement;
```
jaskfla: Why redefine this? Can just delegate it with the rest of `{...props}`

packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+              options: {
+                offset: [0, -10],
+              },
+            },
+          ],
+        },
+        tooltip: {
+          sx: {
+            bgcolor: "var(--discord-legacy-dark-but-not-black);",
+            boxShadow: "0px 0px 5px rgba(0, 0, 0, 0.25)",
```
jaskfla: Not much reason not to use more modern colour functions: `oklch(0 0 0 / 25%)` or `rgb(0 0 0 / 25%)`

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -205,10 +207,59 @@ function calculateTouchOffsetDelta(
...
+            container.clientHeight / params.pixelHeight
+          : Number.POSITIVE_INFINITY,
+        ),
+        initialZoom,
+      )
+    : initialZoom;
+
+  const targetPoint = {
+    x: clamp(canvasX - canvas.startCoordinates[0], 0, canvas.width - 1),
+    y: clamp(canvasY - canvas.startCoordinates[1], 0, canvas.height - 1),
```
jaskfla: Would be nice to restructure these out:

```js
const [startX, startY] = canvas.startCoordinates
```

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -205,10 +207,59 @@ function calculateTouchOffsetDelta(
...
+          params.pixelWidth !== null ?
+            container.clientWidth / params.pixelWidth
+          : Number.POSITIVE_INFINITY,
+          params.pixelHeight !== null ?
+            container.clientHeight / params.pixelHeight
+          : Number.POSITIVE_INFINITY,
+        ),
+        initialZoom,
+      )
+    : initialZoom;
```
jaskfla: Ultimately fine with this as-is, but maybe this could be broken up a bit? Looks like we’re 5 or 6 tabs deep here

Will defer to your call here

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -205,10 +207,59 @@ function calculateTouchOffsetDelta(
     x: movementDelta1.x + movementDelta2.x / 2,
     y: movementDelta1.y + movementDelta2.y / 2,
   };
-  const scale = newMagitude / oldMagitude;
```
jaskfla, line 208: Nice

packages/frontend/src/contexts/CanvasContext.tsx
```diff
@@ -81,6 +94,16 @@ export const CanvasProvider = ({
...
 
+      const url = new URL(window.location.href);
+      url.search = "";
+      if (canvasId !== mainCanvasInfo.id) {
+        url.searchParams.set(
+          SEARCH_PARAM_KEYS.canvasId.canonical,
+          canvasId.toString(),
+        );
+      }
+      router.replace(`${url.pathname}${url.search}${url.hash}`);
```
jaskfla, line 100: Do we need the `url.pathname` here? My hunch is that we can set just the search params and preserve any existing hash

packages/frontend/src/hooks/useCanvasSearchParams.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!value) return null;
+  const parsed = Number.parseFloat(value);
+  return Number.isNaN(parsed) ? null : parsed;
+}
+
+export function useCanvasSearchParams(): CanvasSearchParams {
+  const searchParams = useSearchParams();
+
+  return {
+    canvasId: parseIntParam(extractSearchParam(searchParams, "canvasId")),
```
jaskfla: This should be a path param, not a search param

packages/frontend/src/hooks/useCanvasSearchParams.ts
```diff
@@ -0,0 +1,40 @@
...
+    x: parseIntParam(extractSearchParam(searchParams, "x")),
+    y: parseIntParam(extractSearchParam(searchParams, "y")),
+    zoom: parseFloatParam(extractSearchParam(searchParams, "z")),
+    pixelWidth: parseIntParam(extractSearchParam(searchParams, "pixelWidth")),
+    pixelHeight: parseIntParam(extractSearchParam(searchParams, "pixelHeight")),
+    frameId: extractSearchParam(searchParams, "frameId"),
```
jaskfla, line 47: What’s `frameId`? 👀

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,92 @@
+import { Point } from "@blurple-canvas-web/types";
+import config from "@/config";
+
+type SearchParamConfig = {
+  readonly canonical: string;
+  readonly aliases: readonly string[];
+};
```
jaskfla: I forget—do we preferentially use interfaces in this codebase?

I much prefer them whenever a type can be statically determined

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,92 @@
...
+type ParamVariant<K extends ParamKey> =
+  | (typeof SEARCH_PARAM_KEYS)[K]["canonical"]
+  | (typeof SEARCH_PARAM_KEYS)[K]["aliases"][number];
+
+function getSearchParamVariants<K extends ParamKey>(
+  key: K,
+): readonly ParamVariant<K>[] {
+  const config = SEARCH_PARAM_KEYS[key];
+  return [config.canonical, ...config.aliases];
+}
```
jaskfla, line 39: You can throw this to an LLM agent, but I _think_ we can tighten this return type to infer the string literal types from the `SEARCH_PARAM_KEYS` `const`

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,92 @@
...
+  frameId,
+}: {
+  canvasId?: number;
+  coords?: Point;
+  zoom?: number;
+  pixelWidth?: number;
+  pixelHeight?: number;
+  frameId?: string;
+}) {
+  const parameters = new Map<string, string>();
```
jaskfla: Keyed by `string`, or some derivation of `typeof SEARCH_PARAM_KEYS`?

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,92 @@
...
+      value: pixelHeight?.toFixed(0),
+    },
+    { key: SEARCH_PARAM_KEYS.frameId.canonical, value: frameId?.toUpperCase() },
+  ];
+
+  for (const param of params) {
+    if (param.value) {
+      parameters.set(param.key, param.value);
+    }
+  }
```
jaskfla: This block could be `const parameters = initParameters()`, which just makes it a touch easier to read/parse

Use an IIFE if y’feel like, but I think naming the function helps readability even if this is literally the only usage of the function

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const url = new URL(config.baseUrl);
+
+  url.search = Array.from(parameters.entries())
+    .map(([key, value]) => `${key}=${value}`)
+    .join("&");
```
jaskfla: Would prefer `url.searchParams.set()`

Otherwise, make sure to pass the keys and values through `encodeURIComponent`

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,92 @@
...
+export function extractSearchParam(
+  searchParams: URLSearchParams | null,
+  key: ParamKey,
+): string | null {
+  if (!searchParams) return null;
+
+  return (
+    getSearchParamVariants(key)
+      .map((variant) => searchParams.get(variant))
+      .find((value): value is string => value !== null && value.length > 0) ??
```
jaskfla: Mapping through the whole array only to find one here, whereas it’ll short-circuit if we ditch the `map` and update the `find` callback to call `searchParams.get` and do the value comparison in one iteration

jaskfla (changes requested):
Nitpicks, but clean overall!

packages/frontend/src/hooks/useCanvasSearchParams.ts
S-Ou: lol

packages/frontend/src/contexts/CanvasContext.tsx
S-Ou: now that i'm modifying the pathname, yes

packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
```diff
@@ -96,6 +122,32 @@ export default function PixelInfoTab({
...
+      <ActionPanelTabBody>
+        {adjustedCoords && (
+          <ActionPanelTooltip
+            title="Copied"
+            onClose={closeTooltip}
+            open={tooltipIsOpen}
+          >
+            <DynamicButton
+              color={
+                !isLoading && pixelHistory.length > 0 ?
```
jaskfla: Strictly speaking this check is redundant:

- `data` from `useQuery` is referentially stable
- `?? []` will instantiate a new array on ever render that takes this code path, which (in a very small way) undoes some of `useQuery`’s efforts
- when `isLoading`, `data === undefined`

Non-blocking for this PR, since it’s a small nitpick; but I’d just do `pixelHistory?.[0]?.color ?? null`

packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
```diff
@@ -96,6 +122,32 @@ export default function PixelInfoTab({
...
+                  pixelHistory[0].color
+                : null
+              }
+              onAction={() => {
+                openTooltip();
+                navigator.clipboard.writeText(pixelURL);
+              }}
+            >
+              Copy link to share pixel
```
jaskfla: “Copy pixel link”?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -284,16 +343,75 @@ export default function CanvasView() {
...
     },
     [canvas.id],
   );
 
+  const canvasSearchParams = useCanvasSearchParams();
+  const initialCanvasSearchParamsRef = useRef(canvasSearchParams);
+  const hasAppliedInitialCanvasRef = useRef(false);
+  const hasAppliedInitialViewRef = useRef(false);
+
+  useEffect(() => {
```
jaskfla: Would be nice to use a named function here

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,120 @@
...
+  key: ParamKey,
+): string | null {
+  if (!searchParams) return null;
+
+  for (const variant of getSearchParamVariants(key)) {
+    const value = searchParams.get(variant);
+    if (value !== null && value.length > 0) {
+      return value;
+    }
+  }
```
jaskfla: Did I comment on this loop at some point? 🤨

Could be a `find()`

jaskfla (approved):
Couple optional details, but happy to approve as-is 🙏

packages/frontend/src/util/searchParams.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  const url = new URL(config.baseUrl);
+
+  if (canvasId !== undefined) {
+    url.pathname = `/canvas/${encodeURIComponent(canvasId)}`;
+  }
+
+  parameters.forEach((value, key) => {
+    url.searchParams.set(key, value);
+  });
```
jaskfla: Oh please use a `for`…`of` here

We should ban `forEach` from this repo

packages/frontend/src/util/searchParams.ts
S-Ou: best I'll do, searchParams itself doesn't have .find

PR #373, opened 2026-04-10
feat: Added support for viewing Frames

### LINKED ISSUE #376: View frames

Rocked03:
- Added a new Frames tab that displays all the frames a user owns + owned by a server they belong to
- Will jump to location when clicked
- Currently "read-only"
  - tentative plans for creating/editing frames later

https://github.com/user-attachments/assets/8f5eb249-04af-429d-ae8c-4d6391731ce1

image


packages/backend/src/services/frameService.ts
```diff
@@ -0,0 +1,118 @@
+import { Frame } from "@blurple-canvas-web/types";
+import { prisma } from "@/client";
+import { NotFoundError } from "@/errors";
+
+// biome-ignore lint/suspicious/noExplicitAny: extracting from below
+function frameFromDb(frame: any): Frame {
```
jaskfla: Verbose, but I think still nicer to define something that resolves to
```ts
{
    id: string;
    canvas_id: number;
    owner_id: bigint;
    is_guild_owned: boolean;
    name: string;
    x_0: number;
    y_0: number;
    x_1: number;
    y_1: number;
    style_id: number | null;
}
```

`Pick<T>` should do most of the work, I imagine

packages/backend/src/services/frameService.ts
```diff
@@ -0,0 +1,118 @@
...
+      y_1: true,
+    },
+  });
+
+  return frames.map(frameFromDb);
+}
+
+export async function getFramesByGuildIds(
+  guildIds: string[],
+  canvas_id: number,
```
jaskfla: Intentional snake case? 🧐

packages/backend/src/services/frameService.ts
```diff
@@ -0,0 +1,118 @@
...
+): Promise<Frame[]> {
+  const frames = await prisma.frame.findMany({
+    where: {
+      owner_id: {
+        in: guildIds.map((id) => BigInt(id)),
```
jaskfla:
```suggestion
        in: guildIds.map(BigInt),
```

packages/frontend/src/app/Main.tsx
```diff
@@ -11,10 +11,11 @@ const Wrapper = styled("main")`
     --action-panel-width: 19rem;
     --navbar-height: 4rem;
     column-gap: 1rem;
-    row-gap: 0.5rem;
+    row-gap: 0;
 
     display: grid;
-    grid-template-rows: auto 1fr;
+    // Restricts the height of the page to the viewport
+    grid-template-rows: var(--navbar-height) calc(100dvh - var(--navbar-height));
```
jaskfla, line 18: I don’t 100% hate this because it’s self-contained, but it is a bit hacky and red-flaggy to me

What’s the undesirable behaviour from `auto`, and why not just set `height` on the element itself and let grid take care of the vertical space distribution?

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -93,60 +94,82 @@ export const Heading = styled("h2")`
...
 
-const TABS = {
-  LOOK: "Look",
-  PLACE: "Place",
-};
+enum TABS {
+  LOOK = "Look",
+  PLACE = "Place",
+  FRAMES = "Frames",
+}
```
jaskfla, line 128: Verb, verb, noun

Could very well be fine, but let’s take this opportunity to review all three labels and make sure we’re happy with it as a full set

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -93,60 +94,82 @@ export const Heading = styled("h2")`
...
-      setCurrentTab(TABS.PLACE);
       setColor(tempColor);
     }
+
+    // hiding reticle if we are on frames tab
+    if (newTab === TABS.FRAMES) {
+      setIsReticleVisible(false);
+    } else {
+      setIsReticleVisible(true);
+    }
```
jaskfla:
```suggestion
    setIsReticleVisible(newTab !== TABS.FRAMES);
```

packages/frontend/src/components/action-panel/tabs/BotCommandCard.tsx
```diff
@@ -57,3 +49,14 @@ export default function BotCommandCard() {
     </Wrapper>
   );
 }
+
+export function PlaceBotCommandCard() {
```
jaskfla: Exported, but where is this used?

Don’t mind the component being split, but `BotCommandCard` and `PlaceBotCommandCard` are unclear as a pair of names

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const Thumbnail = styled("canvas")`
+  aspect-ratio: 16 / 9;
+  border-radius: 0.375rem;
+  display: block;
+  image-rendering: pixelated;
+  width: 100%;
+`;
+
+const FrameTitle = styled("p")`
```
jaskfla: Would a heading component be more appropriate here? Or is it perhaps a `<figure>` and `<figcaption>` situation?

A paragraph feels a bit too generic here

packages/frontend/src/components/action-panel/tabs/FrameThumbCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+function getFrameCropRect(
+  frame: Frame,
+  canvasWidth: number,
+  canvasHeight: number,
+) {
+  const bounds = normalizeFrameBounds(frame);
+  const centerX = (bounds.left + bounds.right) / 2;
+  const centerY = (bounds.top + bounds.bottom) / 2;
+
+  // Expand the frame bounds so the frame occupies ~90% of the resulting thumbnail.
```
jaskfla, line 87: Lovely logic, lovely comment

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      crop.x,
+      crop.y,
+      crop.width,
+      crop.height,
+      0,
+      0,
+      THUMB_WIDTH,
+      THUMB_HEIGHT,
+    );
+  }, [frame, sourceImage, canvas.width, canvas.height]);
```
jaskfla: Let’s [name this Effect function](https://neciudan.dev/name-your-effects)

packages/frontend/src/components/action-panel/tabs/FrameThumbCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      crop.height,
+      0,
+      0,
+      THUMB_WIDTH,
+      THUMB_HEIGHT,
+    );
+  }, [frame, sourceImage, canvas.width, canvas.height]);
+
+  return (
+    <CardBody onClick={onClick}>
```
jaskfla: ⚠️ Click handler on a `<div>` is an accessibility issue

We’ll need to make this a `button`. We can style it with `appearance: none` and what not so it doesn’t _look_ like a button. I have some existing code and can point you to it

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </FramesTabBlock>
+    );
+  }
+
+  const guildFrameMap: { [key: string]: Frame[] } = guildFrames.reduce(
+    (map, frame) => {
+      const ownerId = frame.ownerId;
+      if (!map[ownerId]) {
+        map[ownerId] = [];
+      }
```
jaskfla:
```suggestion
      map[ownerId] ||= [];
```

Or, preferably (if safe):

```suggestion
      map[ownerId] ??= [];
```

Also, if it’s called `map`, should it be a `Map`? I haven’t looked close enough to have an opinion

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+      return map;
+    },
+    {} as { [key: string]: Frame[] },
+  );
+
+  const sortedGuildFrameMap = Object.entries(guildFrameMap).sort(
+    ([, framesA], [, framesB]) => {
+      const ownerGuildA = framesA[0]?.ownerGuild?.name || "";
+      const ownerGuildB = framesB[0]?.ownerGuild?.name || "";
+      return ownerGuildA.localeCompare(ownerGuildB);
```
jaskfla, line 96: Fuck yeah `localeCompare`

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+    },
+  );
+
+  function setFrame(frame: Frame) {
+    setSelectedFrame(frame);
+  }
```
jaskfla: What does this… do?

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+                  />
+                ))}
+            </FramesContainer>
+          ))}
+        </ActionPanelTabBody>
+      </ScrollBlock>
+      {selectedFrame && (
+        <ActionPanelTabBody>
+          <FrameInfoCard frame={selectedFrame} />
+          {/* <BotCommandCard command="/frame create" /> */}
```
jaskfla: 🤨

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+    canvasId: canvasId,
+    guildIds: guildIds,
+  });
+
+  if (!user) {
+    return (
+      <FramesTabBlock active={active}>
+        <ActionPanelTabBody>
+          <FramesContainer>
+            <Heading>Your Frames</Heading>
```
jaskfla: We use sentence case, right?

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+            <p>
+              <a href="/signin">Sign in</a> to view frames
```
jaskfla: Plain anchor will reload the whole app, right?

Use `<Link>`

packages/frontend/src/components/action-panel/tabs/SelectedFrameInfoCard.tsx
```diff
@@ -0,0 +1,42 @@
+import { Frame } from "@blurple-canvas-web/types";
+import { styled } from "@mui/material";
+import { User, Users } from "lucide-react";
+
+const Wrapper = styled("div")`
+  // kinda stolen from SelectedColorInfoCard.tsx
```
jaskfla: This is fine

```suggestion
```

packages/frontend/src/components/action-panel/tabs/SelectedFrameInfoCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Heading>{frame.name}</Heading>
+      {frame.ownerGuild ?
+        <OwnerInfo>
+          <Users /> {frame.ownerGuild.name}
+        </OwnerInfo>
+      : frame.ownerUser ?
+        <OwnerInfo>
+          <User /> {frame.ownerUser.username}
+        </OwnerInfo>
+      : null}
```
jaskfla: Bitta duplication here. Probably nicer to short circuit if neither are truthy, and then just do

```ts
<OwnerInfo>
  <Users /> {pred ? a : b}
</OwnerInfo>
```

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -732,6 +795,7 @@ export default function CanvasView() {
         <ReticleContainer
           style={{
             scale: RETICLE_SCALE,
+            display: isReticleVisible ? "block" : "none",
```
jaskfla: `block` display is the default; no need to add it to the cascade

```suggestion
            display: isReticleVisible ? undefined : "none",
```

packages/frontend/src/contexts/CanvasContext.tsx
```diff
@@ -41,9 +46,13 @@ export const CanvasContext = createContext<CanvasContextType>({
     allColorsGlobal: false,
   },
   coords: null,
+  selectedFrame: null,
   adjustedCoords: null,
+  isReticleVisible: false,
   setCoords: () => {},
   setCanvas: () => {},
+  setSelectedFrame: () => {},
```
jaskfla: Can this state be managed elsewhere? This context is getting pretty overloaded

packages/frontend/src/contexts/CanvasContext.tsx
```diff
@@ -58,6 +67,10 @@ export const CanvasProvider = ({
   const [activeCanvas, setActiveCanvas] = useState(mainCanvasInfo);
   const [selectedCoords, setSelectedCoords] =
     useState<CanvasContextType["coords"]>(null);
+  const [selectedFrame, setSelectedFrame] =
+    useState<CanvasContextType["selectedFrame"]>(null);
```
jaskfla: `frame` search param?

packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -0,0 +1,81 @@
...
+        },
+      );
+      console.log("Guild frames response:", response.data);
+      return response.data;
+    }
+
+    return [];
+  };
+
+  return useQuery({
```
jaskfla:
```suggestion
  return useQuery<FrameRequest.ResBody>({
```

packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -0,0 +1,81 @@
...
+        {
+          params: {
+            guildIds: guildIds,
+          },
+          paramsSerializer: {
+            indexes: null,
+          },
+        },
+      );
+      console.log("Guild frames response:", response.data);
```
jaskfla: 🧐

packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const getFrame = async (): Promise<Frame[]> => {
+    if (frameId) {
+      if (canvasId || userId || guildIds) {
+        throw new Error(
+          "Cannot specify multiple query parameters with frameId",
+        );
+      }
+
+      const response = await axios.get<FrameRequest.ResBody>(
+        `${config.apiUrl}/api/v1/frame/${frameId}`,
```
jaskfla: Remember to URI-encode the path params

jaskfla (changes requested):
(Big 😅) handful of concerns, but this looks awesome 🎉

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const guildFrameMap: { [key: string]: Frame[] } = guildFrames.reduce(
+    (map, frame) => {
+      const ownerId = frame.ownerId;
+      if (!map[ownerId]) {
+        map[ownerId] = [];
+      }
+      map[ownerId].push(frame);
+      return map;
+    },
+    {} as { [key: string]: Frame[] },
```
jaskfla: We can get rid of this cast and pass it as a type argument to `.reduce<{ [key: string]: Frame[] }>()`

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+    res.status(200).json(frame);
+  } catch (error) {
+    ApiError.sendError(res, error);
+  }
+});
+
+frameRouter.get("/guilds/:canvasId", async (req, res) => {
+  try {
+    const guildIds: string[] = (req.query.guildIds as string[]) ?? [];
```
pumbas600: This cast feels a little suspicious to me... I would do something like we've done in `pixel.ts` in the `/history` endpoint:
```ts
    // grabbing the x and y from the query
    const queryResult = await PixelHistoryParamModel.safeParseAsync(req.query);
```

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -0,0 +1,43 @@
...
+    const frame = await getFramesByGuildIds(
+      guildIds,
+      Number.parseInt(req.params.canvasId),
```
pumbas600: We also have a safe `parseCanvasId` utility which you should probably be using here. That'll ensure an appropriate error is thrown, rather than the 500 I suspect this will generate

pumbas600 (changes requested):

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
S-Ou: actually wouldn't mind some thoughts on the UI here because it's rather unpolished

Rocked03: image

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
S-Ou: whoops lol, i originally intended to figure out the offset and zoom stuff in this method but forgot to remove this when i ended up doing it elsewhere

packages/frontend/src/components/action-panel/tabs/BotCommandCard.tsx
S-Ou: used in the place tab

S-Ou: it's specifically for /place commands, open to rename suggestions

packages/frontend/src/components/action-panel/tabs/SelectedFrameInfoCard.tsx
S-Ou: The icons are different so did a slightly different thing

packages/frontend/src/contexts/CanvasContext.tsx
S-Ou: will do once search param PR is merged in

packages/frontend/src/hooks/queries/useFrame.tsx
S-Ou: shhhh

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
S-Ou: have commented it out for now since the bot isn't running, but i'll readd it for now and i've created a ticket for dynamically hiding these https://github.com/project-blurple/Canvas-Web/issues/383

packages/frontend/src/contexts/CanvasContext.tsx
S-Ou: Made an issue https://github.com/project-blurple/Canvas-Web/issues/384

packages/frontend/src/app/Main.tsx
Rocked03: having `auto 1fr` meant that the body contents just expanded the page height instead of restricting to viewport height, which basically meant that if the action panel overflowed (a lot of history, or a lot of frames) then it just expanded the entire body instead of turning into a scroll

but on the actual css side, @pumbas600 do you have any thoughts? this is effectively just reimplementing a change you made in your css PR

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
S-Ou: gone for an h3

packages/backend/src/services/frameService.ts
S-Ou: copilot came up with a neat alternative using prisma stuff

packages/frontend/src/components/action-panel/ActionPanel.tsx
S-Ou: "Frame" singular? that's a verb and I suppose describes the action of framing

S-Ou: sticking to verbs feels more appropriate imo

packages/frontend/src/app/Main.tsx
pumbas600: I am also in favour of `auto 1fr` if it's not too hard. Ideally you would have a max-height on the page so that it restricts the height of the grid to ensure it never overflows

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  active?: boolean;
+  canvasId: number;
+}
+
+export default function FramesTab({ active, canvasId }: FramesTabProps) {
+  const { user } = useAuthContext();
+  const { selectedFrame, setSelectedFrame } = useCanvasContext();
+  const sourceImage = useCanvasImage(canvasId);
+  const { width } = useScreenDimensions();
+  const isMobile = width < 768;
```
pumbas600: I'm generally not a huge fan of detecting if we're in a mobile context in JS... Can we using CSS media queries for the related behaviour instead?

What I think might be cool is generating two version of the frame preview image: A 1:1 and a 16:9 version. You can then convert the canvas for each into an image `src` (I think you can do `canvas.toDataURL()`)

Then, you can do something like:
```html
<picture>
  <source srcset={previews.desktop} media="(width >= 50rem)" />
  <img src={previews.mobile}  alt="..." />
</picture>
```

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -93,60 +94,78 @@ export const Heading = styled("h2")`
...
         >
           Look
         </Tab>
+        <Tab
+          active={currentTab === TABS.FRAMES}
+          onClick={() => onSwitchTab(TABS.FRAMES)}
+          onKeyUp={(event) => {
+            if (event.key === "Enter" || event.key === " ")
+              onSwitchTab(TABS.FRAMES);
+          }}
```
pumbas600: I almost feel like we should create our own `Tab` component which triggers `onClick` when these keys are pressed because we've duplicated this for every tab

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const guildIds = user ? decodeUserGuildsBase64(user) : undefined;
+  const { data: userFrames = [] } = useFrame({
+    canvasId: canvasId,
+    userId: user?.id,
+  });
+  const { data: guildFrames = [] } = useFrame({
+    canvasId: canvasId,
+    guildIds: guildIds,
+  });
```
pumbas600, line 56: I think it might be worth having slightly different models returned from our two get endpoints for guild frames and user frames. The guildframe should always have an owning guild, whereas a user frame should always have an owning user. The types don't reflect that currently and so some of your usages look a little strange, e.g.:

`<Heading>{frames[0]?.ownerGuild?.name}</Heading>`

To me, this looks like it's possible for there to be no owner guild, and as a result, an empty heading element

pumbas600: What do you think?

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+    canvasHeight / cropHeight,
+  );
+  cropWidth *= fitScale;
+  cropHeight *= fitScale;
+
+  let cropX = centerX - cropWidth / 2;
+  let cropY = centerY - cropHeight / 2;
+
+  cropX = Math.max(0, Math.min(cropX, canvasWidth - cropWidth));
+  cropY = Math.max(0, Math.min(cropY, canvasHeight - cropHeight));
```
pumbas600: We have a `clamp` utility you can use here :)

pumbas600 (changes requested):
Looking pretty good! I just noticed a few more things after taking a closer look at some of the frontend code

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
S-Ou: I think that's doable

packages/frontend/src/app/Main.tsx
Rocked03: this is actually quite a pain to figure out, it's not cooperating with me

i'm just going to make an issue instead and it can be a later problem

Rocked03: https://github.com/project-blurple/Canvas-Web/issues/395

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -31,10 +32,17 @@ frameRouter.get("/user/:userId/:canvasId", async (req, res) => {
...
 frameRouter.get("/guilds/:canvasId", async (req, res) => {
   try {
-    const guildIds: string[] = (req.query.guildIds as string[]) ?? [];
+    const queryResult = await FrameGuildIdsQueryModel.safeParseAsync(req.query);
+    if (!queryResult.success) {
+      throw new BadRequestError(
+        "Invalid query parameters. Expected guildIds as a string or string array",
+        queryResult.error.issues,
+      );
+    }
```
jaskfla, line 41: Is there an unsafe parse?

```ts
try {
 const { data: { guildIds } } = parseAsync();
} catch (e) {
  if (e instanceof ErrorFromModel) throw new BadRequestError();
  throw e
}
```

Pretty minor—I don’t feel too strongly about this

packages/backend/src/services/frameService.ts
```diff
@@ -1,9 +1,39 @@
...
+      name: true,
+    },
+  },
+  name: true,
+  x_0: true,
+  y_0: true,
+  x_1: true,
+  y_1: true,
+  style_id: true,
+} satisfies Prisma.frameSelect;
```
jaskfla: I assume this is never meant to be mutated?

```suggestion
} as const satisfies Prisma.frameSelect;
```

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
```diff
@@ -96,39 +121,74 @@ export function FramePreviewCard({
...
+        height={MOBILE_THUMB_HEIGHT}
+        aria-label={`${frame.name} frame preview`}
+      />
+      <ThumbnailCanvas
+        className="desktop-thumb"
+        ref={desktopCanvasRef}
+        width={DESKTOP_THUMB_WIDTH}
+        height={DESKTOP_THUMB_HEIGHT}
+        aria-label={`${frame.name} frame preview`}
       />
```
jaskfla: Do these both need to be rendered? 🤨

**Update:** Oh, I see; you’re using CSS to hide/show it. Maybe a micro-optimisation but we could just prevent React from rendering it altogether. I don’t really have a strong opinion either way here

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
```diff
@@ -2,29 +2,53 @@ import { Frame } from "@blurple-canvas-web/types";
...
+  @media (min-width: 50rem) {
+    &.mobile-thumb {
+      display: none;
+    }
+
+    &.desktop-thumb {
+      aspect-ratio: 16 / 9;
+      display: block;
+    }
+  }
```
jaskfla: Bit more idiomatic for this codebase:

```tsx
const DesktopThumbnailCanvas = styled(ThumbnailCanvas)``

const MobileThumbnailCanvas = styled(ThumbnailCanvas)``
```

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -21,6 +23,16 @@ const FramesContainer = styled("div")`
   display: flex;
   flex-direction: column;
   gap: 0.5rem;
+
+  @media (max-width: 767px) {
```
jaskfla: Magic number? As long as it’s been tested I’m fine with it

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -48,54 +60,50 @@ export default function FramesTab({ active, canvasId }: FramesTabProps) {
...
+            {userFrames.length !== 0 ?
+              userFrames.map((frame) => (
+                <FramePreviewCard
+                  key={frame.id}
+                  frame={frame}
+                  sourceImage={sourceImage}
+                  onClick={() => setSelectedFrame(frame)}
+                />
+              ))
+            : <p>You have no frames</p>}
```
jaskfla, line 134: Completely optional but I’d personally be inclined to put this into `<FramePreviewList items={frames} />` that renders a `<ul>` of `<li><button></li>`s

Also: are “preview" and "thumbnail" the same thing? If so, let’s settle on one. Otherwise ignore this

packages/frontend/src/components/action-panel/tabs/SelectedFrameInfoCard.tsx
```diff
@@ -25,18 +24,19 @@ const OwnerInfo = styled("p")`
 export default function FrameInfoCard({ frame }: { frame?: Frame }) {
   if (!frame) return <Wrapper>No frame selected</Wrapper>;
 
+  const ownerInfo =
+    frame.ownerGuild ? { icon: <Users />, label: frame.ownerGuild.name }
+    : frame.ownerUser ? { icon: <User />, label: frame.ownerUser.username }
```
jaskfla: If icon is decorative, remember to pop `aria-hidden` on it

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
jaskfla: Inoffensive and deployable imo

We can polish in subsequent PR 🙏

jaskfla (commented):
Quick review just because I’m squeezing this in a narrow window of spare time; but looks like it’s really coming together

Some nice spots from @pumbas600 too. I’ll come back to give it an official stamp when I can look more closely

packages/backend/src/routes/api/v1/frame.ts
pumbas600: There is, but this is the pattern we have in other places of the API

packages/frontend/src/components/action-panel/tabs/FramePreviewCard.tsx
pumbas600: @Rocked03 did the `<picture>` approach not work?

S-Ou: For Jasper- using CSS at Josh's suggestion, because then we need to calculate viewport width in js
For Josh- kinda but I never really got it working properly, and doing it this way means I can hook into the same theme.breakpoints that we use elsewhere

packages/backend/src/services/frameService.ts
```diff
@@ -0,0 +1,138 @@
...
+    throw new Error(`Frame ${frame.id} is missing a valid guild owner`);
+  }
+
+  return {
+    ...frame,
+    isGuildOwned: true,
+    ownerUser: undefined,
+    ownerGuild: frame.ownerGuild,
+  };
+}
```
pumbas600: This feels like a use for a custom [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)?

E.g.,
```ts
function isGuildFrame(frame: Frame): frame is GuildFrame {
  return frame.ownerGuild && !frame.ownerUser && frame.isGuildOwned
}

function asGuildFrame(frame: Frame): GuildFrame {
  if (!isGuildFrame(frame)) {
    throw new Error(`Frame ${frame.id} is missing a valid guild owner`);
  }

  return frame
}
```

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -0,0 +1,160 @@
...
+    },
+  );
+
+  const frameUrl =
+    (selectedFrame &&
+      createPixelUrl({
+        canvasId: canvasId,
+        frameId: selectedFrame.id,
+      })) ??
+    "";
```
pumbas600: I think this is much clearer:

```suggestion
  const frameUrl =
    selectedFrame ?
      createPixelUrl({
        canvasId: canvasId,
        frameId: selectedFrame.id,
      }))
    : "";
```

packages/types/src/frame.d.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export type GuildFrame = Omit<
+  Frame,
+  "isGuildOwned" | "ownerUser" | "ownerGuild"
+> & {
+  isGuildOwned: true;
+  ownerUser?: undefined;
+  ownerGuild: DiscordGuildRecord;
+};
```
pumbas600: I you restructure this slightly, you can make it so that checking `isGuildOwned` on a `Frame` automatically infers whether it's a `UserFrame` or a `GuildFrame`:

```ts
interface BaseFrame {
  // ... Shared properties
} 

export interface UserFrame extends BaseFrame {
  isGuildOwner: false;
  ownerUser: DiscordUserProfile;
}

export interface GuildFrame extends BaseFrame {
  isGuildOwned: true;
  ownerGuild: DiscordGuildRecord;
}

export type Frame = UserFrame | GuildFrame;
```

packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -0,0 +1,97 @@
...
+    queryKey: ["frame", "id", frameId],
+    queryFn: getFrame,
+    enabled: Boolean(frameId),
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    placeholderData: null,
+  });
+}
+
+export const useFrameFromId = useFrameById;
```
pumbas600: Out of curiosity, why are you creating this :eyes:

packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    const response = await axios.get<FrameRequest.GuildFramesResBody>(
+      `${config.apiUrl}/api/v1/frame/guilds/${canvasId}`,
+      {
+        params: {
+          guildIds: guildIds,
+        },
+        paramsSerializer: {
+          indexes: null,
+        },
```
pumbas600, line 81: Can you leave a comment describing what this `paramsSerializer` is doing?

pumbas600 (dismissed):

packages/backend/src/services/frameService.ts
pumbas600: Note that it may not be necessary with [this change](https://github.com/project-blurple/Canvas-Web/pull/373#discussion_r3072039041).

packages/frontend/src/hooks/queries/useFrame.tsx
Rocked03: Used for search params - the frame ID only specifies that its a frame and not which user or guild it belongs to

Rocked03: oh, but `useFrameFromId` is redundant lol i'll remove it

packages/types/src/frame.ts
```diff
@@ -0,0 +1,37 @@
...
+export interface BaseFrame {
+  id: string;
+  canvasId: number;
+  name: string;
+  x0: number;
+  y0: number;
+  x1: number;
+  y1: number;
+}
+
```
pumbas600, line 22: You should make a `FrameOwnerType` "enum" like:

```ts
// In a utility file...
export type ValuesOf<T> = T[keyof T];

export const FrameOwnerType = {
  User: 'user',
  // ...
} as const;

export type FrameOwnerType = ValuesOf<typeof FrameOwnerType>;
```

Also see: https://youtu.be/jjMbPt_H3RQ?si=neVOSJjYDCVCZd0j

packages/backend/src/services/frameService.ts
```diff
@@ -27,39 +34,83 @@ const frameSelect = {
...
+
+function isUserOwnedFrame(frame: Frame): frame is UserOwnedFrame {
+  return frame.owner.type === FrameOwnerType.User;
+}
+
+function isGuildOwnedFrame(frame: Frame): frame is GuildOwnedFrame {
+  return frame.owner.type === FrameOwnerType.Guild;
+}
```
jaskfla: I don’t mind this, but this seems like it doesn’t need a utility function to me

If it’s for the type guard, then we can type the owner more narrowly, but I’m also fine with deferring this tweak

packages/backend/src/services/frameService.ts
```diff
@@ -27,39 +34,83 @@ const frameSelect = {
...
+    throw new Error(`Frame ${frame.id} is missing a valid user owner`);
+  }
+  return frame;
+}
+
+function asGuildFrame(frame: Frame): GuildOwnedFrame {
+  if (!isGuildOwnedFrame(frame)) {
+    throw new Error(`Frame ${frame.id} is missing a valid guild owner`);
+  }
+  return frame;
```
jaskfla: Looks like this could be an assertion, instead of returning

```ts
function assertIsUserFrame(frame: Frame): asserts frame is UserOwnedFrame {}
```

I’d also opt to defer this tweak for another PR

packages/frontend/src/components/action-panel/tabs/FramePreviewList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  ${({ theme }) => theme.breakpoints.down("md")} {
+    display: grid;
+    gap: 1rem;
+    grid-template-columns: repeat(2, minmax(0, 1fr));
+  }
+`;
+
+const FrameListItem = styled("li")`
+  list-style: none;
+`;
```
jaskfla: No need for this. We have a global style that puts `list-style-type: none` on `ul`s

Just put `role="list"` on `FrameList`:

```tsx
const FrameList = styled("ul").attrs({ role: 'list' })`
```

packages/frontend/src/components/action-panel/tabs/FrameThumbCard.tsx
```diff
@@ -175,20 +169,18 @@ export function FramePreviewCard({
...
-    <CardBody onClick={onClick}>
-      <ThumbnailCanvas
-        className="mobile-thumb"
+    <CardBody>
+      <MobileThumbnailCanvas
         ref={mobileCanvasRef}
         width={MOBILE_THUMB_WIDTH}
         height={MOBILE_THUMB_HEIGHT}
-        aria-label={`${frame.name} frame preview`}
+        aria-label={`${frame.name} frame thumbnail`}
```
jaskfla: To be honest we could hide these from the a11y tree; it isn’t adding that much meaning. The name is probably the best we’ll manage

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -70,62 +75,115 @@ export default function FramesTab({ active, canvasId }: FramesTabProps) {
...
+            const firstFrame = frames[0];
+            if (!firstFrame) {
+              return null;
+            }
+
+            return (
+              <FramesContainer key={ownerId}>
+                <Heading>{firstFrame.owner.guild.name}</Heading>
+                <FramePreviewList
+                  items={[...frames].sort((a, b) =>
```
jaskfla: `toSorted()` is baseline widely available:

```suggestion
                  items={frames.toSorted((a, b) =>
```

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -70,62 +75,115 @@ export default function FramesTab({ active, canvasId }: FramesTabProps) {
...
       {selectedFrame && (
         <ActionPanelTabBody>
           <FrameInfoCard frame={selectedFrame} />
           <BotCommandCard command="/frame create" />
+          {selectedFrame.owner.type !== "system" && (
+            <ActionPanelTooltip
+              title="Copied"
+              onClose={closeTooltip}
+              open={tooltipIsOpen}
+            >
```
jaskfla, line 174: I haven’t pulled the code and tried it, but if this works like I’m imagining: nice touch 👌

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -473,6 +511,38 @@ export default function CanvasView() {
...
+      const shouldApplyForCurrentCanvas =
+        params.canvasId === null || params.canvasId === canvas.id;
+      if (!shouldApplyForCurrentCanvas) return;
+
+      // Wait for frame lookup before deciding whether initial frame view can be applied.
+      if (isInitialFrameFromSearchParamsLoading) return;
+
+      if (
+        initialFrameFromSearchParams &&
+        initialFrameFromSearchParams.canvasId === canvas.id
```
jaskfla: Is it save to use this?

```suggestion
        initialFrameFromSearchParams?.canvasId === canvas.id
```

If there’s a case where it might be `undefined === undefined`, then this is an unsafe suggestion

packages/frontend/src/contexts/SelectedFrameContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+}) => {
+  const [selectedFrame, setSelectedFrame] =
+    useState<SelectedFrameContextType["frame"]>(null);
+
+  return (
+    <SelectedFrameContext.Provider
+      value={{
+        frame: selectedFrame,
+        setFrame: setSelectedFrame,
+      }}
```
jaskfla: I take it consumers are always expected to destructure the return value? Because `value` as an object won’t be referentially stable between renders

Since it’s a get/set, could honestly consider `value={[frame, setFrame]}`

packages/frontend/src/hooks/queries/useFrame.tsx
jaskfla, line 1: Remember to pass path segments through `encodeURIComponent`

packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -10,71 +10,87 @@ import { useQuery } from "@tanstack/react-query";
...
-            indexes: null,
-          },
+    const response = await axios.get<FrameRequest.GuildFramesResBody>(
+      `${config.apiUrl}/api/v1/frame/guilds/${canvasId}`,
+      {
+        params: {
+          guildIds: guildIds,
+        },
+        paramsSerializer: {
+          // This is needed to send the guildIds as repeated query parameters (e.g., guildIds=1&guildIds=2) instead of a comma-separated list (e.g., guildIds=1,2)
```
jaskfla, line 79: Oh, weird! I would’ve thought it would send `&guildIds[]=1&guildIds[]=2`

Thanks for commenting

jaskfla (approved):
One or two tiny nitpicks left and a handful of optional things to leave for a later PR

Otherwise looking clean. Nice stuff

PR #388, opened 2026-04-11
refactor(frontend): retire dynamically coloured button `filter` hack 

jaskfla:
What we were trying to do is precisely what [`contrast-color()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color) is good at (and arguably _for_`)

Keeping the older hack (which I still like) around because Chromium browser support isn’t quite there yet (though not far off!)

### DIFF
--- a/packages/frontend/src/components/button/DynamicButton.tsx
+++ b/packages/frontend/src/components/button/DynamicButton.tsx
@@ -17,7 +17,6 @@ const StyledButton = styled(ButtonBase, {
     background-color: var(--dynamic-bg-color);
 
     :hover,
-    :focus,
     :focus-visible {
       ${({ backgroundColorStr }) =>
         backgroundColorStr &&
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/button/DynamicButton.tsx
```diff
@@ -38,21 +37,28 @@ const StyledButton = styled(ButtonBase, {
 const DynamicButtonContent = styled("span")`
```
jaskfla, line 37: Pretty sure I’m responsible for this, but looking at this today I have no idea what this `<span>` is doing. Seems like we could get by with just the `<button>`

I’d also be keen to refactor `backgroundColorStr` into a `data-` attribute and access it via `background-colour: attr(data-foo type(<color>)`, but that takes a proper sit-down sesh for me to figure out all the interaction details

Rocked03 (approved):
this seems pretty sane to me

PR #389, opened 2026-04-11
feat(frontend): use colour name instead of “this color”

jaskfla:
Also removes a useless `<div>`, and nicer colour inheritance.


Rocked03 (approved):
sweet ty

PR #392, opened 2026-04-11
feat(prisma): upgrade Prisma to v7 and move to testcontainer test setup

stijnvdkolk:
Upgrades Prisma from v5 to v7 and ditches prismock in favour of running tests against a real Postgres (via Testcontainers). Every test runs inside a transaction that gets rolled back, so you get real DB behaviour with full test isolation.

### Prisma v7

- `prisma-client-js` -> `prisma-client` with `@prisma/adapter-pg`
- New `prisma.config.ts` for configuration.
- Generated client now lives in `src/client/generated/` (gitignored), model types re-exported from `@/client`
- Services import types from `@/client` instead of `@prisma/client`
- Dropped `@@id` from views (errors in v7)

### Tests

- Removed prismock and the old `singleton.ts` that swapped `PrismaClient` for `PrismockClient`
- `vitest.globalSetup.ts` starts a Postgres container and runs `prisma db push` to make sure the database matches the schema.
- `vitest.setup.ts` opens an interactive transaction before each test and rolls it back after. Nested `$transaction` calls in app code are turned into savepoints so they work.
- Seed functions are now async (they weren't awaiting anything before, as prismock was synchronous)
- Seed order respects FK constraints
- Removed `jest.config.js` as it was not used


packages/backend/src/services/pixelService.test.ts
```diff
@@ -195,12 +201,15 @@ describe("Get Cooldown Tests", () => {
...
-    seedCanvases();
-    seedColors();
-    seedPixels();
+    await seedEvents();
+    await seedUsers();
+    await seedDiscordProfiles();
+    await seedGuilds();
+    await seedCanvases();
+    await seedColors();
+    await seedPixels();
```
pumbas600: Perhaps at this point it's worth just using `await seedAll()`?

packages/backend/src/test/vitest.setup.ts
```diff
@@ -0,0 +1,80 @@
...
+          throw err;
+        }
+      };
+    }
+
+    if (prop === "$connect" || prop === "$disconnect") {
+      return () => Promise.resolve();
+    }
+
+    const value = transactionClient![prop as keyof Prisma.TransactionClient];
```
pumbas600: This `!` isn't necessary :)

```suggestion
    const value = transactionClient[prop as keyof Prisma.TransactionClient];
```

packages/backend/src/test/vitest.setup.ts
```diff
@@ -0,0 +1,80 @@
...
+    return typeof value === "function" ? value.bind(transactionClient) : value;
```
pumbas600, line 44: By assigning the transaction client to a local variant after checking it's not null, you can avoid needing to use `!` in the $transaction callback:

```suggestion
    const tx = transactionClient;

    if (prop === "$transaction") {
      return async (arg: unknown) => {
        const id = `sp_${++savePointCounter}`;
        await tx.$executeRawUnsafe(`SAVEPOINT ${id};`);
        try {
          const result =
            Array.isArray(arg) ?
              await Promise.all(arg)
            : await (arg as (tx: unknown) => Promise<unknown>)(tx);
          await tx.$executeRawUnsafe(`RELEASE SAVEPOINT ${id};`);
          return result;
        } catch (err) {
          await tx.$executeRawUnsafe(`ROLLBACK TO SAVEPOINT ${id};`);
          throw err;
        }
      };
    }

    if (prop === "$connect" || prop === "$disconnect") {
      return () => Promise.resolve();
    }

    const value = tx[prop as keyof Prisma.TransactionClient];
    return typeof value === "function" ? value.bind(tx) : value;
```

packages/backend/src/test/vitest.globalSetup.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const execAsync = promisify(exec);
+let container: StartedPostgreSqlContainer;
+export async function setup() {
+  container = await new PostgreSqlContainer("postgres:13.3-alpine").start();
+
+  process.env.DATABASE_URL = container.getConnectionUri();
+  await execAsync(
+    `DATABASE_URL=${process.env.DATABASE_URL} npx prisma db push`,
+  );
```
pumbas600: This method of setting the `DATABASE_URL` environment variable doesn't work on Windows unfortunately 😔 

Maybe you could use [dotenv-cli](https://www.npmjs.com/package/dotenv-cli)? We already have `dotenv`

pumbas600 (changes requested):
This is amazing work, thank you for doing this 🙏 

Unfortunately though, I can't run the tests on windows because of how you're trying to set the database url environment variable

packages/backend/src/test/vitest.globalSetup.ts
stijnvdkolk: Ah true, I forgot about Windows, oops, will think about the best way to fix, maybe dotenv-cli will work

packages/backend/src/services/pixelService.test.ts
stijnvdkolk: Guess we can do that yea haha

packages/backend/src/test/vitest.globalSetup.ts
stijnvdkolk: Made it work without using an external dependency

packages/backend/src/test/vitest.globalSetup.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const execAsync = promisify(exec);
+let container: StartedPostgreSqlContainer;
+export async function setup() {
+  container = await new PostgreSqlContainer("postgres:13.3-alpine").start();
+
+  process.env.DATABASE_URL = container.getConnectionUri();
+  await execAsync("npx prisma db push", {
+    env: { ...process.env, DATABASE_URL: process.env.DATABASE_URL },
+  });
```
pumbas600, line 17: This is a much better idea!

packages/backend/src/routes/api/v1/pixel.test.ts
pumbas600, line 1: Out of curiosity, are these tests timing out for you? I think it's related to the express app thats being created, because all the other ones seem fine.

I even tried extending the timeout using `--hook-timeout=30000 --test-timeout=50000` with no success

stijnvdkolk: image
They work for me 😅

pumbas600: image

stijnvdkolk: And they also seem to work on the GitHub CI, so it's a bit confusing why they are timing out for you locally...

pumbas600: Aha! It turns out I just needed to enable the place pixel endpoint (With the `WEB_PLACING_ENABLED` environment variable)

pumbas600 (approved):
This is awesome, thank you!

If you're happy with the changes I made, I think this looks great @stijnvdkolk 🥳

packages/backend/src/routes/api/v1/pixel.test.ts
stijnvdkolk: Haha, glad you got it solved!

Issue #385, opened 2026-04-11
Admin tooling


Rocked03:
- View canvas info
- Create new canvases
- Paste pixels?

Rocked03:
Probably would go hand-in-hand with #371

stijnvdkolk:
Looks like all sub issues got closed, so moved to done 🚀

Issue #377, opened 2026-04-10
Create & edit frames


Rocked03:

Rocked03:
Need a way to select the frame's bounding box, some sort of area selection functionality? Would be good to have this as a generic thing too so it can be used elsewhere (moderation etc)

Rocked03:
Also, can create frames for yourself, but need to hit Discord API to check if they're authorised to make frames on behalf of a server

Must have admin or manage server perms, which are checked iif the bot is in that server

Issue #371, opened 2026-04-10
Moderation tooling


Rocked03:
Pixels can only be batch overwritten with bot commands and manually manipulating the DB

Would be good to have some GUI tools, although this is a large feature that would need some thinking through before going ahead with anything

- Add/remove from blacklist
- Moderate history
- Modify pixels

Rocked03:
Probably would go hand-in-hand with #385

Rocked03:
Perhaps modifying pixels themselves might be a bit heavy-handed, but manipulating history?

Maybe being able revert + rollback history authored by a specific user over a selected area (or entire canvas) + a specified time (or entire history)

Make sure pixels get correctly reverted if so

User filter as a search/dropdown of all history authors? dropdown prioritises users on the blacklist that still have authored history

Rocked03:
Concept:
- Moderation search
- Filters
  - User
  - Timeframe
  - Bounding box (will use the same frame selection tech as frame creation)
- Result is a highlighted set of pixel history
- Action options
  - Remove selected history + rollback to latest pixel
  - Blacklist user

PR #396, opened 2026-04-12
feat: auth discord permissions

Rocked03:
This is a setup PR with backend functionality that will be useful for #377, #385, and #371, so these are getting implemented first so development can happen in parallel easier

- `api/v1/guilds/:guildId/permissions` - returns whether the authenticated user has `administrator` and `manage_server` permissions in the given guild. Throws an error if the bot is not in the guild.
- `isCanvasAdmin()`, `isCanvasModerator` - configured via env, checks if the user has the specified admin/moderator role. This is what will determine whether they can use admin and moderation tooling.

The backend `.env` now requires a `DISCORD_BOT_TOKEN` for API requests against guilds.


packages/backend/src/config/index.ts
```diff
@@ -33,6 +33,9 @@ const config = {
   discord: {
     clientId: requiredEnv("DISCORD_CLIENT_ID"),
     clientSecret: requiredEnv("DISCORD_CLIENT_SECRET"),
+    managementGuild: process.env.DISCORD_MANAGEMENT_GUILD_ID,
+    adminRole: process.env.DISCORD_ADMIN_ROLE_ID,
+    moderatorRole: process.env.DISCORD_MODERATOR_ROLE_ID,
```
jaskfla: I think nicer to keep the `discord` prefix even for the keys here:

1. consistent with the env var name;
2. kinda acts as a namespace
3. then we can alphabetically sort without worrying about this becoming a mess

(and please do alphabetically sort)

packages/backend/src/config/processEnv.d.ts
jaskfla, line 1: Personally I’d also alpha sort here

If you want some logical grouping (could be nice to keep `PORT` and `NODE_ENV` separate, pop in an empty line

packages/backend/src/routes/api/v1/discord.ts
jaskfla, line 1: Do we have an existing pattern of wrapping basically the entire handler in a single `try`–`catch` block? Looking at this as unfamiliar code, it’s hard to tell which lines actually throw

packages/backend/src/services/discordGuildService.ts
```diff
@@ -0,0 +1,189 @@
...
+      `Discord API request failed with status ${response.status}: ${endpoint}`,
+    );
+  }
+
+  return (await response.json()) as T;
+}
+
+function asBearerToken(accessToken: string): string {
+  return `Bearer ${accessToken}`;
+}
```
jaskfla, line 76: Could be inlined, but I don’t mind too much

packages/backend/src/services/discordGuildService.ts
```diff
@@ -0,0 +1,189 @@
...
+    administrator,
+    manage_guild: manageGuild,
+  };
+}
+
+export async function userHasRoleInGuild(
+  guildId: string,
+  roleId: string,
+  accessToken: string,
+): Promise<boolean> {
```
jaskfla: Super clean function—love it. Maybe nicer with named params? (Optional)

packages/backend/src/services/discordGuildService.ts
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  return Object.fromEntries(
+    guilds.map((guild) => {
+      const permissions = BigInt(guild.permissions ?? "0");
+      const administrator =
+        (permissions & ADMINISTRATOR_PERMISSION) === ADMINISTRATOR_PERMISSION;
+      const manageGuild =
+        administrator ||
+        (permissions & MANAGE_GUILD_PERMISSION) === MANAGE_GUILD_PERMISSION;
```
jaskfla: This masking logic looks duplicated?

packages/backend/src/services/discordGuildService.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+}
+
+export function ensureCanvasModerator(user: DiscordUserProfile): void {
+  if (!user.isCanvasModerator) {
+    throw new ForbiddenError(
+      "You do not have permission to perform this action",
+    );
+  }
+}
```
jaskfla: These are `assert`s, not `ensure`s

packages/backend/src/utils/discordRouteUtils.ts
```diff
@@ -0,0 +1,9 @@
+import BadRequestError from "@/errors/BadRequestError";
+
+const DiscordSnowflakePattern = /^\d+$/;
```
jaskfla: _Any_ number of digits? 👀 

Also, casing:

```suggestion
const discordSnowflakePattern = /^\d+$/;
```

packages/backend/src/utils/discordRouteUtils.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export function validateSnowflake(value: string, fieldName: string): void {
```
jaskfla:
```suggestion
export function assertIsSnowflake(value: string, fieldName: string): void {
```

packages/frontend/src/contexts/AuthProvider.tsx
```diff
@@ -31,10 +33,38 @@ interface AuthProviderProps {
 
 export function AuthProvider({ children, profile }: AuthProviderProps) {
   const [user, setUser] = useState(profile);
+  const { data: userData } = useUserData(user);
+
+  useEffect(() => {
+    if (!user) {
+      return;
+    }
+  }, [user]);
```
jaskfla: Uh… leftover debug Effect?

packages/frontend/src/contexts/AuthProvider.tsx
```diff
@@ -31,10 +33,38 @@ interface AuthProviderProps {
...
+      if (!currentUser || currentUser.guilds) {
+        return currentUser;
+      }
+
+      return {
+        ...currentUser,
+        guilds: userData.guilds,
+      };
+    });
+  }, [user, userData]);
```
jaskfla: I’ll be honest, I haven’t traced this. Would be helpful to name the function, or find a way to do it declaratively

My strong suspicion is that [this Effect doesn’t need to exist](https://react.dev/learn/you-might-not-need-an-effect)

If you don’t find an Effectless solution within 5–10 minutes just re-ping me and I can take a closer look

packages/frontend/src/contexts/AuthProvider.tsx
```diff
@@ -31,10 +33,38 @@ interface AuthProviderProps {
...
 
   const signOut = useCallback<AuthContextType["signOut"]>(() => {
     // Delete the session cookie
-    axios.post(`${config.apiUrl}/api/v1/discord/logout`).catch(console.error);
+    axios
+      .post(`${config.apiUrl}/api/v1/discord/logout`, undefined, {
+        withCredentials: true,
+      })
+      .catch(console.error);
```
jaskfla, line 51: Huh, curious that this isn’t in a `useMutation`. Out of scope, though, so dw

packages/frontend/src/hooks/queries/useUserData.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export function useUserData(user: DiscordUserProfile | null) {
+  const getUserData = async () => {
+    const response = await axios.get<UserDataResponse>(
+      `${config.apiUrl}/api/v1/discord/guilds/permissions-map`,
+      { withCredentials: true },
+    );
+    return response.data;
+  };
+
```
jaskfla: By my read we can define this at the top level? Very much a micro-optimisation, but this function gets recreated every render

(That said, I think this is an anti-pattern in the codebase I’m responsible for 😅)

jaskfla (approved):
Looks solid! Nothing from me that will surprise you 😆

jaskfla (changes requested):
Sorry, I misclicked 😓

packages/backend/src/routes/api/v1/discord.ts
Rocked03: yes

i agree tho, but that can be a refactor for another time

packages/backend/src/services/discordGuildService.ts
```diff
@@ -34,6 +34,9 @@ interface DiscordRequestOptions {
   authorization: string;
 }
 
+type CanvasAdminUser = DiscordUserProfile & { isCanvasAdmin: true };
+type CanvasModeratorUser = DiscordUserProfile & { isCanvasModerator: true };
```
jaskfla: Flagging that we should turn these into `interface`s later

Can descope from here

jaskfla (approved):

packages/backend/src/services/discordGuildService.ts
Rocked03: that's easy to do, have updated

PR #397, opened 2026-04-13
feat: disable bot placing endpoint with flag

Rocked03:
We have an endpoint that listens out for pixel placements from the Discord bot, this adds an .env flag that disables the endpoint (in the case that the bot isn't running)


packages/backend/src/config/index.ts
```diff
@@ -40,6 +40,8 @@ const config = {
    */
   webGuildId: 0,
   webPlacingEnabled: process.env.WEB_PLACING_ENABLED === "true",
+  // Keep bot placing enabled by default unless explicitly disabled.
+  botPlacingEnabled: process.env.BOT_PLACING_ENABLED !== "false",
```
pumbas600, line 44: From our conversation today, it sounds like bot placing should be disabled by default if we're not intending on keeping it around going forwards?

pumbas600 (approved):
Looks good, just a small query on the default value

packages/backend/src/config/index.ts
Rocked03: i'm leaving it like this for now

PR #409, opened 2026-04-14
fix: cache now clears upon canvas unlock

### LINKED ISSUE #408: Backend serves static locked cache if canvas unlocked
In the case that:
1. A canvas starts unlocked
2. The canvas gets locked
3. The backend saves the locked snapshot in the static cache, as intended
4. Further queries to the backend will serve the cached content rather than a new query
5. The canvas then gets unlocked (for whatever reason)
6. The backend continues to serve the static cached content instead of fetching new data

Rocked03:
Fixes #408

Backend keeps fetching from static locked cache even if canvas gets unlocked

### DIFF
--- a/packages/backend/src/services/canvasService.ts
+++ b/packages/backend/src/services/canvasService.ts
@@ -289,11 +289,11 @@ function saveCanvasToFileSystem(canvas: canvas, pixels: PixelColor[]): string {
   return path;
 }
 
-function clearCanvasFromFileSystem(canvasId: number): void {
+async function clearCanvasFromFileSystem(canvasId: number): Promise<void> {
   const cachedCanvas = CANVAS_CACHE[canvasId];
 
   if (cachedCanvas?.isLocked) {
-    fs.rmSync(cachedCanvas.canvasPath);
+    await fs.promises.rm(cachedCanvas.canvasPath);
     console.debug(`Cleared canvas ${canvasId} from file system`);
   }
 }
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/canvasService.ts
```diff
@@ -295,7 +289,16 @@ function saveCanvasToFileSystem(canvas: canvas, pixels: PixelColor[]): string {
   return path;
 }
 
-async function getAndCacheCanvas(canvasId: number): Promise<CachedCanvas> {
+function clearCanvasFromFileSystem(canvasId: number): void {
+  const cachedCanvas = CANVAS_CACHE[canvasId];
+
+  if (cachedCanvas?.isLocked) {
+    fs.rmSync(cachedCanvas.canvasPath);
```
jaskfla: Can `clearCanvasFromFileSystem` be async? Better to go with `fs.rm()` I reckon

packages/backend/src/services/canvasService.ts
```diff
@@ -304,6 +307,21 @@ async function getAndCacheCanvas(canvasId: number): Promise<CachedCanvas> {
     throw new NotFoundError(`There is no canvas with ID ${canvasId}`);
   }
 
+  var cachedCanvas = CANVAS_CACHE[canvasId];
```
jaskfla: `var`?? Looks like this can be a `const`

packages/backend/src/services/canvasService.ts
```diff
@@ -304,6 +307,21 @@ async function getAndCacheCanvas(canvasId: number): Promise<CachedCanvas> {
...
+  if (cachedCanvas) {
+    if (cachedCanvas.isLocked !== canvas.locked) {
+      console.debug(
+        `Canvas ${canvasId} lock status has changed. Updating cache...`,
+      );
+      clearCanvasFromFileSystem(canvasId);
+    } else {
+      console.debug(`Cache hit for canvas ${canvasId}`);
+      return CANVAS_CACHE[canvasId];
```
jaskfla: May as well use `cachedCanvas` reference here

jaskfla (approved):

PR #410, opened 2026-04-14
fix: replace dotenv with dotenvx to support env string templating

### LINKED ISSUE #407: Prisma DB string templating no longer works
`{other env var here}` shapes in the DB string are currently read literally instead of substituting the actual env value

stijnvdkolk:
This should fix the env templating. dotenv itself does not support it out of the box, but dotenvx, the successor of dotenv (made by the same person) does support it out of the box.

Closes #407

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -39,6 +39,7 @@
     "clean": "rm -rf node_modules/ build/"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


pumbas600 (approved):
Awesome thank you! I was just talking with @Rocked03 about how this got broken :)

PR #416, opened 2026-04-16
fix(frontend): bungled font stack

jaskfla:
My bad, sorry. Must’ve used `git commit` with `-a` flag with some unstashed changes that shouldn’t’ve been in my working copy.


Rocked03 (approved):
rrr how dare you >:(

PR #417, opened 2026-04-16
refactor: split CanvasContext out to CanvasViewContext

### LINKED ISSUE #384: Refactor CanvasView-related states to separate context
#373 and #369 are both introducing more states to the `CanvasContext` (`containerRef`, `zoom`, `selectedFrame`) that should be moved to a separate context

Rocked03:
`CanvasContext` was getting rather bloated, so this PR splits out all the CanvasView-related states into `CanvasViewContext`

also a small reticle fix


pumbas600 (approved):
🫡

PR #419, opened 2026-04-17
feat: create and edit frames

### LINKED ISSUE #377: Create & edit frames

Rocked03:
It's a hefty one...

Added functionality to create, edit, and delete frames from the frames tab.

- Includes a new menu to edit frame properties (name, owner)
- Includes a whole new "`SelectedBounds`" system that creates an interactive overlay on the `CanvasView` that can be resized to define frame bounds. Set up in a way that it can be used elsewhere as well.
- Styling is pretty WIP, but can be tidied up a lot better in another PR
  - styling still needed for
    - the frame edit menu itself
    - sign in prompt
    - discard changes dialog
    - confirm delete dialog

image

Closes #377


packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -125,6 +136,7 @@ function Tab({
         if (event.key === "Enter" || event.key === " ") onSwitchTab(tabKey);
       }}
       tabIndex={0}
+      disabled={isLocked}
     >
```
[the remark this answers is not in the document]
S-Ou: this caused me issues to fix so i'm electing to ignore it

packages/backend/src/models/paramModels.ts
```diff
@@ -32,6 +32,40 @@ export const FrameGuildIdsQueryModel = z.object({
...
+    }
+
+    if (y1 <= y0) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        path: ["y1"],
+        message: "y1 must be greater than y0",
+      });
+    }
+  });
```
jaskfla, line 64: I feel like we can normalise this?

Like, given any two points, we can trivially figure out how to turn that rectangle into $\left\( x_0, y_0 \right\), \left\( x_1, y_1 \right\)$ where $x_0 < x_1, y_0 < y_1$ and store that

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  editFrame: vi.fn(),
+  getFrameById: vi.fn(),
+  getFramesByGuildIds: vi.fn(),
+  getFramesByUserId: vi.fn(),
+}));
+
+import { createFrame, deleteFrame, editFrame } from "@/services/frameService";
+import { frameRouter } from "./frame";
+
+type EndpointCase = {
```
jaskfla: `interface`

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -0,0 +1,152 @@
...
+import { mockAuth } from "@/test/mockAuth";
+
+vi.mock("@/services/frameService", () => ({
+  createFrame: vi.fn(),
+  deleteFrame: vi.fn(),
+  editFrame: vi.fn(),
+  getFrameById: vi.fn(),
+  getFramesByGuildIds: vi.fn(),
+  getFramesByUserId: vi.fn(),
+}));
```
jaskfla, line 26: Curious placement. Significant?

If so, worth a quick inline comment

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -0,0 +1,152 @@
...
+type EndpointCase = {
+  name: string;
+  path: string;
+  body: Record<string, unknown>;
+  successStatus: number;
+  successMessage: string;
+  serviceName: "create" | "edit" | "delete";
+};
+
+const endpointCases: EndpointCase[] = [
```
jaskfla: If it doesn’t cause issues, would be nice to use `as const satisfies EndpointCase[]` here; but if that causes issues just ignore

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,98 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
     ApiError.sendError(res, error);
   }
 });
+
+frameRouter.post("/:frameId/edit", async (req, res) => {
```
jaskfla: Uh, why not `PUT /:frameId` (or `PATCH`)? 🤨

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,98 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+  try {
+    if (!req.user || !req.session.discordAccessToken) {
+      throw new ApiError("Unauthorized", 401);
+    }
+
+    const frameId = await parseFrameId(req.params);
+
+    const bodyQueryResult = await FrameDataParamModel.safeParseAsync(req.body);
```
jaskfla: Think we can pop these in `Promise.all`

Same in the creation endpoint

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,98 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+    if (!bodyQueryResult.success) {
+      throw new BadRequestError(
+        "Invalid body parameters",
+        bodyQueryResult.error.issues,
+      );
+    }
+
+    await editFrame(
+      req.user as DiscordUserProfile,
```
jaskfla: Curious that this isn’t inferred. Something must be wrong with upstream typing?

If it’s a small fix, would be nice to tighten up and avoid this line of debt—but descope if it’s nontrivial

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,98 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+      bodyQueryResult.data.x1,
+      bodyQueryResult.data.y1,
+    );
+    res.status(200).json({ message: "Frame edited successfully" });
+  } catch (error) {
+    ApiError.sendError(res, error);
+  }
+});
+
+frameRouter.post("/:frameId/delete", async (req, res) => {
```
jaskfla: `DELETE /:frameID`

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,98 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+      req.session.discordAccessToken,
+      frameId,
+    );
+    res.status(200).json({ message: "Frame deleted successfully" });
+  } catch (error) {
+    ApiError.sendError(res, error);
+  }
+});
+
+frameRouter.post("/create", async (req, res) => {
```
jaskfla: No need for `/create`. `POST` already has the right semantics

packages/backend/src/services/frameService.ts
```diff
@@ -159,3 +161,171 @@ export async function getFramesByGuildIds(
...
+      throw new ForbiddenError(
+        "You do not have permission to modify frames for this guild",
+      );
+    }
+  } else {
+    if (ownerId !== user.id) {
+      throw new ForbiddenError("You are not the owner of this frame");
+    }
+  }
+}
```
jaskfla, line 184: Optional, but I find the shallower indent from an early return a touch more eye-pleasing

```suggestion
  if (isGuildOwned) {
    const permissions = await getGuildPermissionsForUser(ownerId, accessToken);

    if (!permissions.administrator && !permissions.manage_guild) {
      throw new ForbiddenError(
        "You do not have permission to modify frames for this guild",
      );
    }
    
    return; // or return true, but then also return true at the end if so
  }

  if (ownerId !== user.id) {
    throw new ForbiddenError("You are not the owner of this frame");
  }
}
```

packages/backend/src/services/frameService.ts
```diff
@@ -159,3 +161,171 @@ export async function getFramesByGuildIds(
...
+  await assertUserHasPermissionsForFrame(
+    user,
+    accessToken,
+    isGuildOwned,
+    ownerId,
+  );
+
+  await assertCoordsAreWithinCanvas(canvasId, x0, y0, x1, y1);
+
+  while (true) {
```
jaskfla, line 301: Is this an existing thing? This feels unhinged to me, but I’m fine with it if it’s a remnant of existing code

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -41,7 +41,7 @@ const TabBar = styled("ul")`
   list-style-type: none;
 `;
 
-const TabStyle = styled("li")<{ active?: boolean }>`
+const TabStyle = styled("li")<{ active?: boolean; disabled?: boolean }>`
```
jaskfla: 1. This is not a transient prop, and will be incorrectly forwarded to the DOM as an invalid attribute of `<li>`. We _would_ fix this by configuring it with `shouldForwardProps`, but;
2. Just use `aria-disabled` and use CSS selector `[aria-disabled=true]`

(Point 1 also applies to the `active` prop, but I need to fix the entire semantics for this tabbing as discussed offline 🫣. For now let’s just not add to the debt)

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -108,11 +117,13 @@ function Tab({
   tabKey,
   label,
   currentTab,
+  isLocked,
```
jaskfla: Completely optional, for for a prop on an FC I feel `locked` fits in more neatly with other boolean attributes like `required`, `disabled`, `hidden`

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -134,12 +146,15 @@ function Tab({
 export default function ActionPanel() {
   const [currentTab, setCurrentTab] = useState(TABS.PLACE);
   const [tempColor, setTempColor] = useState<PaletteColor | null>(null);
+  const [tabsLocked, setTabsLocked] = useState(false);
```
jaskfla: …on the other hand, for a JS state variable I feel `is`/`are`/`has` as a bool indicator is more idiomatic

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -163,26 +178,32 @@ export default function ActionPanel() {
           label="Place"
           currentTab={currentTab}
           onSwitchTab={onSwitchTab}
+          isLocked={tabsLocked && currentTab !== TABS.PLACE}
```
jaskfla: …why? 🤨

Are you running into weird rendering issues?

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -1,190 +1,49 @@
...
-              </DynamicButton>
-            </ActionPanelTooltip>
-          )}
-        </ActionPanelTabBody>
-      )}
+      {activePanel === FramePanelState.Info ?
+        <FrameInfoPanel setActivePanel={setActivePanel} />
+      : (
+        activePanel === FramePanelState.Edit ||
+        activePanel === FramePanelState.Create
```
jaskfla: Maybe define a map object?

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -1,190 +1,49 @@
...
 
-const FramesContainer = styled("div")`
-  display: flex;
-  flex-direction: column;
-`;
+export const FramePanelState = {
+  Info: "info",
+  Create: "create",
+  Edit: "edit",
+};
```
jaskfla: “state“ to me implies a `[value, setValue]` tuple

Sounds like this is a modal tab panel. Perhaps `FramePanelMode` is more appropriate?

packages/frontend/src/components/button/DynamicButton.tsx
```diff
@@ -90,6 +90,7 @@ export default function DynamicButton({
       backgroundColorStr={backgroundColorStr}
       onClick={clickHandler}
       onKeyUp={keyUpHandler}
+      disabled={disabled}
```
jaskfla: Alphasort? 🥺

packages/frontend/src/components/button/PlacePixelButton.tsx
```diff
@@ -123,12 +123,12 @@ export default function PlacePixelButton({ isVerbose }: PlacePixelButtonProps) {
   const { x, y } = adjustedCoords;
   const nbsp = "\u00A0";
 
-  const placePixelMessege =
```
jaskfla, line 141: Nice catch 😆

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -369,10 +369,20 @@ export default function CanvasView() {
...
-    useCanvasViewContext();
+  const {
+    containerRef,
+    coords,
+    isReticleVisible,
+    offset,
+    zoom,
+    setCoords,
+    setOffset,
+    setZoom,
```
jaskfla, line 384: Alphasort? 🥺

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -369,10 +369,20 @@ export default function CanvasView() {
   const canvasPanAndZoomRef = useRef<HTMLDivElement>(null);
 
   const { color } = useSelectedColorContext();
-  const [frame, setFrame] = useSelectedFrameContext();
+  const { frame, setFrame } = useSelectedFrameContext();
```
jaskfla, line 373: This change feels backward to me. Is there other stuff in the selected frame context?

packages/frontend/src/components/canvas/SelectedBoundsOverlay.tsx
```diff
@@ -0,0 +1,518 @@
...
+}: {
+  canvasWidth: number;
+  canvasHeight: number;
+  canEdit: boolean;
+  minHeight: number;
+  minWidth: number;
+  selectedBounds: ViewBounds | null;
+  reticleScale: number;
+  reticleSize: number;
+  setSelectedBounds: (value: ViewBounds) => void;
```
jaskfla:
```suggestion
  setSelectedBounds: React.Dispatch<SetStateAction<ViewBounds>>
```

packages/frontend/src/components/canvas/SelectedBoundsOverlay.tsx
```diff
@@ -0,0 +1,518 @@
...
+    startClientY: number;
+  } | null>(null);
+
+  const overlayCutoutPath = useMemo(() => {
+    if (!selectedBounds) return null;
+
+    const left = Math.max(0, Math.min(canvasWidth, selectedBounds.left));
+    const top = Math.max(0, Math.min(canvasHeight, selectedBounds.top));
+    const right = Math.max(left, Math.min(canvasWidth, selectedBounds.right));
+    const bottom = Math.max(top, Math.min(canvasHeight, selectedBounds.bottom));
```
jaskfla: These are `clamp`s, right?

packages/frontend/src/components/canvas/SelectedBoundsOverlay.tsx
```diff
@@ -0,0 +1,518 @@
...
+          <path
+            d={overlayCutoutPath}
+            fill="#808080"
+            fillRule="evenodd"
+            fillOpacity={0.25}
+          />
+        </OverlayDesaturateShade>
+      )}
+      {selectedBoundsEdges.map((edge) => {
+        const edgeThickness = 14 / zoom;
```
jaskfla: Magic number? 🤨

Making it a const or commenting it might make it clearer to read

packages/frontend/src/components/canvas/SelectedBoundsOverlay.tsx
```diff
@@ -0,0 +1,518 @@
...
+
+        return (
+          <CornerHitTarget
+            key={`selected-bounds-edge-hit-${edge.key}`}
```
jaskfla: Not a fan of just `edge.key`? They only need to be unique amongst siblings

packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
```diff
@@ -180,7 +180,10 @@ export default function PlacePixelTab({
         )}
         {canPlacePixel && <PlacePixelButton isVerbose={!isLarge} />}
         {isJoinServerShown && (
-          <DynamicAnchorButton color={selectedColor} href={serverInvite}>
+          <DynamicAnchorButton
+            color={selectedColor?.rgba || null}
```
jaskfla: Does passing `undefined` cause issues?

packages/frontend/src/components/canvas/SelectedBoundsOverlay.tsx
```diff
@@ -0,0 +1,518 @@
...
+
+    event.preventDefault();
+    event.stopPropagation();
+
+    event.currentTarget.releasePointerCapture(event.pointerId);
+    dragStateRef.current = null;
+  }
+
+  return (
+    <>
```
jaskfla, line 608: This Fragment is a bit overwhelming. How much of it could be extracted into separate modules?

packages/frontend/src/components/canvas/SelectedBoundsOverlay.tsx
```diff
@@ -0,0 +1,518 @@
...
+            scale: reticleScale,
+            transform: `translate(${corner.offset.x}px, ${corner.offset.y}px)`,
+          }}
+        >
+          <OverlayReticle
+            src="/images/boundsCorner.png"
+            alt=""
+            aria-hidden
+            style={{
+              transform: `rotate(${getCornerRotation(corner.key as CornerKey)}deg)`,
```
jaskfla: Nice

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+  Autocomplete,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+  InputLabel,
+  TextField,
+  ToggleButton,
+  ToggleButtonGroup,
```
jaskfla, line 16: Confirming that all of these need MUI? My hunch is that most don’t

- `Autocomplete` can be an `<input>` with a [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist)
- `Dialog` maybe look complex enough to warrant MUI, but could it be a native `<dialog>`?
- `TextField` should definitely be a plain `<input>`
- `InputLabel` can be `<label>`
- `ToggleButton` should just be `<input type="radio">`

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+`;
+
+const PreviewContainer = styled("div")`
+  display: flex;
+  flex-direction: column;
+  gap: 0.5rem;
+`;
+
+const EditPreviewCanvas = styled(PreviewCanvas)`
+  height: auto;
```
jaskfla: Is this not the default?

```suggestion
  height: unset;
```

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+`;
+
+const ButtonRow = styled("div")`
+  background: transparent;
+  display: flex;
+  gap: 0.5rem;
+  padding: 0;
+  width: 100%;
```
jaskfla:
```suggestion
  display: flex;
  gap: 0.5rem;
  inline-size: 100%;
```

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+  > * {
+    flex: 1 1 0;
+    min-width: 0;
+  }
+`;
+
+type GuildEntry = [string, GuildData];
+type GuildOption = {
```
jaskfla: `interface`

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+  guildId: string;
+  guild: GuildData;
+  group: string;
+};
+
+function areBoundsEqual(a: ViewBounds | null, b: ViewBounds | null) {
+  if (!a && !b) return true;
+  if (!a || !b) return false;
```
jaskfla, line 91: We explicitly want `(null, null)` to evaluate to `true`?

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+    },
+    [[], []],
+  );
+}
+
+export default function FrameEditPanel({
+  setActivePanel,
+  isCreateMode,
+}: {
+  setActivePanel: (panel: FramePanelState) => void;
```
jaskfla: `React.Dispatch`

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+    if (selectedFrame) {
+      setFrameBounds(normalizeFrameBounds(selectedFrame));
+    } else {
+      setBoundsToCurrentView(0.75);
+    }
+
+    setCanEdit(true);
+
+    didInitBoundsRef.current = true;
```
jaskfla: Named function? 🥺

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+  }, [selectedFrame, setFrameBounds, setBoundsToCurrentView, setCanEdit]);
+
+  const [selectedOwner, setSelectedOwner] = useState<FrameOwnerType>(
+    selectedFrame ? selectedFrame.owner.type : FrameOwnerType.User,
```
jaskfla:
```suggestion
    selectedFrame?.owner.type ?? FrameOwnerType.User,
```

Ditto in the next statement

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,609 @@
...
+    frameBounds,
+    frameName,
+    selectedOwner,
+    selectedGuildId,
+    isDirtyTrackingReady,
+  ]);
+
+  const managedGuildEntries = Object.entries(user?.guilds ?? {})
+    .filter(([, guild]) => guild.administrator || guild.manageGuild)
+    .toSorted(([, a], [, b]) => (b.memberCount ?? 0) - (a.memberCount ?? 0));
```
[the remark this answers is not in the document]
jaskfla: I’m fine with targeting baseline widely available

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+      group: withoutFramesGroup,
+    }));
+
+    return [...withFrames, ...withoutFrames];
+  }, [managedGuildEntries, guildFrames]);
+
+  const selectedGuildOption =
+    guildOptions.find((option) => option.guildId === selectedGuildId) ?? null;
+
+  const hasUnsavedChanges = useMemo(() => {
```
jaskfla: `isDirty`

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+  const refreshFrameQueries = async () => {
+    if (user) {
+      await queryClient.invalidateQueries({
+        queryKey: ["frame", "user", canvas.id, user.id],
+      });
+    }
+    await queryClient.invalidateQueries({
+      queryKey: ["frame", "guild", canvas.id],
+    });
```
jaskfla: `Promise.all`

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -0,0 +1,610 @@
...
+    }
+
+    closeEditor();
+  };
+
+  const handleDeleteButtonAction = () => {
+    setIsDeleteConfirmOpen(true);
+  };
+
+  const handleSaveAction = async () => {
```
jaskfla, line 368: `useMutation`, with cache invalidation in place of `refreshFrameQueries`

jaskfla (changes requested):
Alphabetically, I’ve only reviewed down to `packages/frontend/src/components/frames/FrameEditPanel.tsx`, but this review is getting a bit out of hand

Gonna let these comments through for now

- Haven’t checked out the branch, but looks like there’s some clever stuff in the canvas overlay;
- couple semantic issues with the HTTP methods in the backend API
- handful of refactors that would really improve readability;
- would also like to see TanStack Query for the consumers of the new endpoints so we can get away from imperative code in the render cycle as much as possible

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const guildId = frame.owner.guild.guild_id;
+    const userGuildData = user.guilds?.[guildId];
+    if (userGuildData !== undefined) {
+      if (userGuildData.administrator || userGuildData.manageGuild) {
+        return true;
+      }
+    }
+  }
+
+  return false;
```
jaskfla: Assuming this evaluates to a boolean:

```suggestion
    return userGuildData !== undefined && (userGuildData.administrator || userGuildData.manageGuild)
```

Otherwise just smoosh it through `Boolean()`

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+        frameId: selectedFrame.id,
+      })
+    : "";
+
+  const userHasPermsToEditSelectedFrame =
+    selectedFrame && user && userCanEditFrame(user, selectedFrame);
+
+  return (
+    <>
+      <ScrollBlock>
```
jaskfla: What does `ScrollBlock` mean? At a glance, hard to infer what it does; but also I get that naming things is hard

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+              <DynamicButton
+                color={hexStringToPixelColor(selectedFrame.id)}
+                onAction={() => {
+                  setTooltipIsOpen(true);
+                  navigator.clipboard.writeText(frameUrl);
+                }}
+              >
+                Copy frame link
+              </DynamicButton>
+            </ActionPanelTooltip>
```
jaskfla: This could be its own component that wraps `DynamicButton`

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+          )}
+        </ActionPanelTabBody>
+      : user ?
```
jaskfla: Is this a nested ternary? Probably worth extracting into a component or IIFE or _something_

JSX life 😔

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  useCanvasContext,
+  useSelectedFrameContext,
+} from "@/contexts";
+import { useGuildFrames, useUserFrames } from "@/hooks/queries/useFrame";
+import { useCanvasImage } from "@/hooks/useCanvasImage";
+import { Heading } from "../action-panel/ActionPanel";
+import { ActionPanelTabBody } from "../action-panel/tabs/ActionPanelTabBody";
+import { FramePreviewList } from "../action-panel/tabs/FramePreviewList";
+
+const FramesContainer = styled("div")`
```
jaskfla: Now that [container](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) means something in CSS, I’d prefer to use “wrapper”  here

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  const groupedByOwnerId = guildFrames.reduce<
+    Record<string, GuildOwnedFrame[]>
+  >((acc, frame) => {
+    const ownerId = frame.owner.guild.guild_id;
+    acc[ownerId] ??= [];
+    acc[ownerId].push(frame);
+    return acc;
+  }, {});
```
jaskfla: This is just `Object.groupBy(guildFrames, f => f.owner.guild.guild_id)` right?

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+      const firstFrameA = framesA[0];
+      const firstFrameB = framesB[0];
+      if (!firstFrameA || !firstFrameB) {
+        return 0;
+      }
+
+      const ownerGuildA = firstFrameA.owner.guild.name;
+      const ownerGuildB = firstFrameB.owner.guild.name;
+      return ownerGuildA.localeCompare(ownerGuildB);
+    },
```
jaskfla: Extract into `function sortByOwningGuildName(a, b)` (or something like that)

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  );
+
+  const inbuiltFullCanvasFrame: SystemOwnedFrame = {
```
jaskfla: Could this be `as const satisfies SystemOwnedFrame`?

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+    x1: canvas.width,
+    y1: canvas.height,
+    owner: {
+      type: "system",
+      name: "Blurple Canvas",
+    },
+  };
+
+  return (
+    <ActionPanelTabBody>
```
jaskfla, line 92: Ah, here comes some regret calling it the **action panel**

Conventionally, we would call the body of a tab the **tab panel**, but **action panel tab panel** shan’t do 😆

packages/frontend/src/components/frames/FrameEditPanel.tsx
S-Ou: technically... yes? it shouldn't ever happen but this is for evaluating if data has changed and if null is still null then the data hasn't changed, so this is intended behaviour still

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <FramesContainer>
+        <Heading>Your Frames</Heading>
```
jaskfla: Sentence case everywhere

```suggestion
        <Heading>Your frames</Heading>
```

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+        {user ?
```
jaskfla, line 95: Hmmmm, not in love with the nested ternary but maybe I can accept this one

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,119 @@
...
+      </FramesContainer>
+      <FramesContainer>
+        <Heading>Blurple Canvas</Heading>
+        <FramePreviewList
+          items={[inbuiltFullCanvasFrame]}
+          sourceImage={sourceImage}
+          onSelectFrame={setSelectedFrame}
+        />
+      </FramesContainer>
+      {sortedGuildFrameMap.map(([ownerId, frames]) => {
```
jaskfla, line 116: Worth timing (`console.time` and `console.timeEnd`) how long this takes. Could be a non-issue, but this is sorting a list inside a map, so let’s check that it’s quick

Remember to throttle your CPU to something reasonable because you probably have quite a beefy system

packages/frontend/src/components/frames/FrameEditPanel.tsx
S-Ou: I'm going to defer this to the future Frame styling PR

packages/frontend/src/components/frames/FramePreview.tsx
jaskfla, line 1: Nice symbol naming in this module

packages/frontend/src/components/canvas/CanvasView.tsx
S-Ou: Not every usage of the context need both the getter and setter, it has better parity with the selectedColorContext

packages/frontend/src/components/frames/FramePreview.tsx
```diff
@@ -0,0 +1,90 @@
...
+export type SourceRect = {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+};
+
+export const PreviewCanvas = styled("canvas")`
+  border-radius: 0.375rem;
+  display: block;
```
jaskfla: Double check the cascade; I’ve already set this in the `normalize` CSS layer in `core.css`

I’d be surprised if this line is actually necessary

packages/frontend/src/components/action-panel/ActionPanel.tsx
S-Ou: When editing a frame, this locks the tabs so you can't switch mid-edit. I could alternatively link it up to the same confirm dialog as the back button, but these tabs sit so high up in the hierarchy that that feels like a nightmare to wire up... and just simply disabling the tab buttons feels like a pretty reasonable UI decision

packages/backend/src/services/frameService.ts
S-Ou: It's an existing thing, and I think it's neat to leave it intact

packages/frontend/src/contexts/CanvasViewContext.tsx
```diff
@@ -57,16 +63,26 @@ export const CanvasViewProvider = ({ children }: CanvasViewProviderProps) => {
     return null;
   }, [canvas.startCoordinates, selectedCoords]);
 
+  // biome-ignore lint/correctness/useExhaustiveDependencies: Trigger on canvas change
+  useEffect(
+    function resetCoordsOnCanvasChange() {
+      setSelectedCoords(null);
+    },
+    [canvas.id],
+  );
```
jaskfla: Don’t need an effect for this. Keep the previous render cycle’s `canvas.id` in some internal state, and compare to see if it’s changed between renders:

```js
const [prevCanvasId, setPrevCanvasId] = useState(initCanvasId);

if (prevCanvasId !== canvas.Id) {
  setSelectedCoords(null)
  setPrevCanvasId(canvasId) // for next render cycle
}
```

This potentially saves an entire render cycle, because this Effect sets state, and anything that depends on `selectedCoords` will then re-render. Get rid of the Effect and all those changes are evaluated in one go

packages/frontend/src/contexts/SelectedBoundsContext.tsx
```diff
@@ -0,0 +1,178 @@
...
+  setSelectedBounds: () => {},
+  setBoundsToCurrentView: () => {},
+});
+
+interface CurrentViewProps {
+  canvas: CanvasInfo;
+  containerRef: RefObject<HTMLDivElement | null>;
+  offset: Point;
+  zoom: number;
+}
```
jaskfla: Props or params? 🤨

packages/frontend/src/contexts/SelectedBoundsContext.tsx
```diff
@@ -0,0 +1,178 @@
...
+  children,
+}: SelectedBoundsProviderProps) => {
+  const [selectedBounds, setSelectedBounds] =
+    useState<SelectedBoundsContextType["selectedBounds"]>(null);
+  const [canEdit, setCanEdit] =
+    useState<SelectedBoundsContextType["canEdit"]>(false);
+  const [minHeight, setMinHeight] =
+    useState<SelectedBoundsContextType["minHeight"]>(5);
+  const [minWidth, setMinWidth] =
+    useState<SelectedBoundsContextType["minWidth"]>(5);
```
jaskfla, line 130: Nice use of derived types

packages/frontend/src/contexts/SelectedBoundsContext.tsx
```diff
@@ -0,0 +1,178 @@
...
+    setMinHeight(height);
+  }
+
+  const { canvas } = useCanvasContext();
+  const { containerRef, offset, zoom } = useCanvasViewContext();
+
+  function clearSelectedBounds() {
+    setSelectedBounds(null);
+    setCanEdit(false);
+  }
```
jaskfla: Any functions in the context `value` should be wrapped in `useCallback`, otherwise they get reinstantiated every render

Any consumer that uses it will then be re-rendered any time this context renders

In general, try to make sure that any context or hook returns referentially stable values

packages/frontend/src/util/index.ts
```diff
@@ -73,3 +87,35 @@ export function getUserGuildFlags(
 ): Record<string, GuildData> {
   return user.guilds ?? {};
 }
+
+export function normalizeFrameBounds(frame: Frame): ViewBounds {
```
jaskfla: Completely optional

```suggestion
export function normalizeFrameBounds({ x0, x1, y0, y1 }: Frame): ViewBounds {
```

packages/frontend/src/util/index.ts
```diff
@@ -73,3 +87,35 @@ export function getUserGuildFlags(
...
+  }
+
+  const normalized = hex.startsWith("#") ? hex : `#${hex}`;
+  if (!/^#([0-9A-Fa-f]{6})$/.test(normalized)) {
+    return null;
+  }
+
+  const r = parseInt(normalized.slice(1, 3), 16);
+  const g = parseInt(normalized.slice(3, 5), 16);
+  const b = parseInt(normalized.slice(5, 7), 16);
```
jaskfla: [Use `Number.parseInt`](https://biomejs.dev/linter/rules/use-number-namespace/)

jaskfla (changes requested):
First review done (finally!)

A couple comments I’ve omitted in the line-by-line comments:

- still a couple `SetStateAction`s that have been manually typed
- couple `type`s that could be `interface`s
- a lot of state we’re keeping in contexts should be moved to the URL; but very much not for this PR

packages/frontend/src/components/frames/FrameList.tsx
S-Ou: well, that's on you

packages/frontend/src/components/frames/FrameInfoPanel.tsx
S-Ou: It's a block that allows the content inside to overflow via scroll. Means that the other parts of the action panel can stay at their fixed heights while the contents of the ScrollBlock get the scrollbar treatment

packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
S-Ou: had to add an `?` but it shouldn't now

packages/frontend/src/components/frames/FrameEditPanel.tsx
S-Ou: I've had to keep the padding: 0 otherwise it inherits it

packages/backend/src/routes/api/v1/frame.test.ts
S-Ou: no idea, this test stuff was purely ai generated

packages/frontend/src/components/frames/FrameList.tsx
Rocked03: image

packages/backend/src/utils/index.ts
```diff
@@ -3,3 +3,17 @@
...
+  x1: number,
+  y1: number,
+): { x0: number; y0: number; x1: number; y1: number } {
+  return {
+    x0: Math.min(x0, x1),
+    y0: Math.min(y0, y1),
+    x1: Math.max(x0, x1),
+    y1: Math.max(y0, y1),
+  };
+}
```
jaskfla: - Pretty sure we have US spelling everywhere (except probably comments), so `normalizeBounds`
- args is a bit awkward here

Two options:

```ts
([x0, y0], [x1, y1]) => [[x0, y0], [x1, y1]]
// or
({ x0, y0, x1, y1 }) => { x0, y0, x1, y1 }
```

Second one probably closer in spirit to what you’ve got; and I _think_ closer to how we represent points elsewhere?

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -125,6 +133,7 @@ function Tab({
         if (event.key === "Enter" || event.key === " ") onSwitchTab(tabKey);
       }}
       tabIndex={0}
+      aria-disabled={locked}
```
jaskfla: Why not `disabled`?

packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
```diff
@@ -39,7 +39,7 @@ export const TabBlock = styled(Block, {
   grid-template-rows: auto 1fr auto;
 `;
 
-export const ScrollBlock = styled(Block)`
+export const ScrollableBlock = styled(Block)`
```
jaskfla: `ScrollView`, really

In this CSS-adjacent context, we should reserve "block" to mean the display type imo

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -1,190 +1,47 @@
...
-  display: flex;
-  flex-direction: column;
-`;
+export const FramePanelMode = {
+  Info: "info",
+  Create: "create",
+  Edit: "edit",
+};
+
+export type FramePanelMode = ValueOf<typeof FramePanelMode>;
```
jaskfla, line 18: Pretty sure this evaluates to `string` unless we add `as const` in line 16, right?

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+  if (frame.owner.type === "guild") {
+    const guildId = frame.owner.guild.guild_id;
+    const userGuildData = user.guilds?.[guildId];
+    return (
+      userGuildData !== undefined &&
+      (userGuildData.administrator || userGuildData.manageGuild)
+    );
+  }
+
+  return false;
```
jaskfla: If we use a switch statement, TS should know that this is unreachable

packages/frontend/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "lib": ["dom", "dom.iterable", "ES2023"],
+    "lib": ["dom", "dom.iterable", "ES2024"],
```
jaskfla: To avoid merge conflict

```suggestion
    "lib": ["dom", "dom.iterable", "esnext"],
```

packages/types/src/index.ts
```diff
@@ -11,3 +11,4 @@ export type * from "./pixelHistory";
 export type * from "./pixelInfo";
 export type * from "./point";
 export type * from "./statistics";
+export * from "./utils";
```
jaskfla: Also to avoid merge conflict

```suggestion
export type * from "./utils";
```

packages/frontend/src/components/action-panel/ActionPanel.tsx
Rocked03: this is the one where you suggested switching away from `disabled` to `aria-disabled` because `button` wouldn't like it

Rocked03: oh well this is redundant now that main is merged in

jaskfla: That explains it!

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -78,12 +77,12 @@ frameRouter.put("/:frameId/edit", async (req, res) => {
...
-      bodyQueryResult.data.y0,
-      bodyQueryResult.data.x1,
-      bodyQueryResult.data.y1,
-    );
+    const { x0, y0, x1, y1 } = normalizeBounds({
+      x0: bodyQueryResult.data.x0,
+      y0: bodyQueryResult.data.y0,
+      x1: bodyQueryResult.data.x1,
+      y1: bodyQueryResult.data.y1,
+    });
```
jaskfla: I know the result also has a `name` property, but that’s fine

```suggestion
    const { x0, y0, x1, y1 } = normalizeBounds(bodyQueryResult.data);
```

Ditto line 136

packages/backend/src/utils/index.ts
```diff
@@ -4,12 +4,17 @@ BigInt.prototype.toJSON = function (): string {
...
+  x0,
+  y0,
+  x1,
+  y1,
+}: {
+  x0: number;
+  y0: number;
+  x1: number;
+  y1: number;
+}): { x0: number; y0: number; x1: number; y1: number } {
```
jaskfla: Optional: define a type and reuse it for the arg and return types

packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
```diff
@@ -24,23 +24,29 @@ export const ActionPanelTabBody = styled("div")`
...
+export function TabPanel(
+  props: React.ComponentPropsWithRef<typeof StyledView>,
+) {
+  return <StyledView role="tabpanel" {...props} />;
+}
+
+export const ScrollableView = styled(View)`
   align-self: stretch;
   grid-column: 1 / -1;
 `;
```
jaskfla, line 46: String of misleading names here:

- `<View>` has overflow `auto`, so _that_ is a `ScrollView`
- `<StyledView>`[^foot] probably shouldn’t be exported (assuming you  haven’t added a usage that isn’t semantically a tabpanel
- `<ScrollableView>` should be called `FullWidthScrollView`

[^foot]: Moot point: I remain annoyed that Emotion refuses to add an option to set default props, because with **styled-components** `TabPanel` could be defined with the styles from `StyledView` and the default `role` prop in one go

Out of scope: I’d like to remove the `active` prop and just the `hidden` HTML attribute

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -24,7 +24,11 @@ interface FramesTabProps extends React.ComponentPropsWithRef<
   setTabsLocked: (locked: boolean) => void;
 }
 
-export default function FramesTab({ active, setTabsLocked }: FramesTabProps) {
+export default function FramesTab({
+  active,
```
jaskfla, line 22: We’re just forwarding this prop; no need to restructure it out. Just delegate it with `...props`

packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
jaskfla, line 1: Out of scope: flagging for myself that the components in this folder can and should be more idiomatically named

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -492,10 +491,10 @@ export default function FrameEditPanel({
               }}
               disabled={!isCreateMode} // Can't change owner after frame is created
             >
-              <ToggleButton value={FrameOwnerType.User}>You</ToggleButton>
-              <ToggleButton value={FrameOwnerType.Guild}>Server</ToggleButton>
+              <ToggleButton value={"user"}>You</ToggleButton>
+              <ToggleButton value={"guild"}>Server</ToggleButton>
```
jaskfla:
```suggestion
              <ToggleButton value="user">You</ToggleButton>
              <ToggleButton value="guild">Server</ToggleButton>
```

jaskfla (approved):
Comment in `ActionPanelTabBody.tsx` is a request, but pre-approving. The rest are less critical suggestions

Come together nicely—thanks for your patience 🙏

.github/workflows/ci.yml
```diff
@@ -79,3 +79,25 @@ jobs:
...
+        uses: pnpm/action-setup@v5
+
+      - name: Set up Node
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: "package.json"
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
jaskfla, line 100: Oh, weird that we don’t validate lockfile in its own step

I might sort this out if I feel like it some day. Super noncritical

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      name: "Frame name",
+      ownerId: "1",
+      isGuildOwned: false,
+      x0: 0,
+      y0: 0,
+      x1: 10,
+      y1: 10,
+    },
+    successStatus: 201,
+    successMessage: "Frame created successfully",
```
jaskfla: Sorry, one more request: remove ”successfully“ from all of these

“Frame created“ communicates all the information with less fluff

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    successStatus: 200,
+    successMessage: "Frame edited successfully",
+    serviceName: "edit",
+  },
+  {
+    name: "delete",
+    method: "delete",
+    path: "/api/v1/frame/abc123/delete",
+    body: {},
+    successStatus: 200,
```
jaskfla:
```suggestion
    successStatus: 204,
```

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  app.use("/api/v1/frame", frameRouter);
+  return app;
+};
+
+const sendMutationRequest = ({
+  app,
+  method,
+  path,
+  body,
+}: {
```
jaskfla: I would suggest `sendMutationRequest(path, { app, body, method })`

Should almost never have a function with more than 3 params imo—it’s just impossible to tell at a glance what the params are when they’re just based on order

Not too bad here since it’s just a test suite, though

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,108 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+frameRouter.delete("/:frameId/delete", async (req, res) => {
+  try {
+    if (!req.user || !req.session.discordAccessToken) {
+      throw new ApiError("Unauthorized", 401);
+    }
+
+    const frameId = await parseFrameId(req.params);
+
+    await deleteFrame(req.user, req.session.discordAccessToken, frameId);
+    res.status(200).json({ message: "Frame deleted successfully" });
```
jaskfla: `204` with no message is fine; any `2xx` code implies success—no need to send the word “success”

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,108 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+  } catch (error) {
+    ApiError.sendError(res, error);
+  }
+});
+
+frameRouter.post("/", async (req, res) => {
+  try {
+    if (!req.user || !req.session.discordAccessToken) {
+      throw new ApiError("Unauthorized", 401);
+    }
```
jaskfla, line 117: Out of scope: Would be nice for us to add middleware to handle auth

cc @pumbas600

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -49,3 +59,108 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+      canvasId,
+      bodyQueryResult.data.name,
+      ownerQueryResult.data.ownerId,
+      ownerQueryResult.data.isGuildOwned,
+      x0,
+      y0,
+      x1,
+      y1,
+    );
+    res.status(201).json({ message: "Frame created successfully" });
```
jaskfla: Convention is to send the created record, no?

packages/backend/src/services/frameService.ts
```diff
@@ -158,3 +160,171 @@ export async function getFramesByGuildIds(
...
+    if (!permissions.administrator && !permissions.manage_guild) {
+      throw new ForbiddenError(
+        "You do not have permission to modify frames for this guild",
+      );
+    }
+  } else {
+    if (ownerId !== user.id) {
+      throw new ForbiddenError("You are not the owner of this frame");
+    }
+  }
```
jaskfla:
```suggestion
  }
  
  if (ownerId !== user.id) {
    throw new ForbiddenError("You are not the owner of this frame");
  }
```

packages/backend/src/services/frameService.ts
```diff
@@ -158,3 +160,171 @@ export async function getFramesByGuildIds(
...
+          y_0: y0,
+          x_1: x1,
+          y_1: y1,
+        },
+      });
+      return;
+    } catch (error) {
+      if (
+        error instanceof Prisma.PrismaClientKnownRequestError &&
+        error.code === "P2002"
```
jaskfla: Does Prisma export `P2002` with a symbol of sorts? This is very magic-string-y. Something like `Prisma.ErrorCode.UniqueConstraintViolation`. I assume you got this code from _somewher_ 🤷

packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
```diff
@@ -33,3 +36,37 @@ export default function ActionPanelTooltip({
     </Tooltip>
   );
 }
+
+export function TooltipDynamicButton({
```
jaskfla, line 44: ⚠️ This isn’t `styled`-able

I’d suggest making its props extend `React.ComponentPropsWithoutRef<typeof DynamicButton>` and delegating props the button

(`…WithoutRef` because I’m _pretty_ sure that tooltips use refs, but I’m not actually that deeply familiar with the inner workings)

packages/frontend/src/components/action-panel/tabs/ActionPanelTooltip.tsx
```diff
@@ -1,8 +1,11 @@
+import { PixelColor } from "@blurple-canvas-web/types";
 import { Tooltip, TooltipProps } from "@mui/material";
```
jaskfla: Super minor, but MUI recommends default imports. We already have enough trouble with dev server performance, so let’s try squeeze a bit more out

```suggestion
import Tooltip, { TooltipProps } from "@mui/material/Tooltip";
```

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -1,196 +1,55 @@
...
-  const frameUrl =
-    selectedFrame ?
-      createPixelUrl({
-        canvasId: canvasId,
-        frameId: selectedFrame.id,
-      })
-    : "";
+  useEffect(() => {
+    setTabsLocked(activePanel !== FramePanelMode.Info);
+  }, [activePanel, setTabsLocked]);
```
jaskfla: Don’t need an Effect for this

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -0,0 +1,113 @@
...
+                }}
+              >
+                Create frame
+              </DynamicButton>
+            </ActionPanelTabBody>
+          );
+        }
+
+        return null;
+      })()}
```
jaskfla: This IIFE warrants its own FC. Looking at it, you could even encapsulate…

```tsx
  const { user } = useAuthContext();
  const { frame: selectedFrame } = useSelectedFrameContext();
```

…within it

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const groupedByOwnerId = Object.groupBy(
+    guildFrames,
+    (f) => f.owner.guild.guild_id,
+  );
+
+  const sortedGuildFrameMap = Object.entries(groupedByOwnerId)
+    .filter(
+      (entry): entry is [string, GuildOwnedFrame[]] => entry[1] !== undefined,
+    )
+    .sort(sortByOwnerGuildName);
```
jaskfla: Bit of a micro-optimisation, but worth it because it also improves readability:

Here we’re calling `useGuildFrames` for the sole purpose of transforming it into `sortedGuildFrameMap`

Extract the `Object.entries(Object.groupBy()).filter().sort()` logic into a function at module scope (for referential stability) with signature `(data: GuildFrames) => SortedGuildFrameMap`, and pass it as `options.select` to the `UseQueryOptions` in `useGuildFrames`

As part of that, add a second parameter with type `UseQueryOptions<FrameRequest.GuildFramesResBody>` to `useGuildFrames`

(I’m sure co-pilot will do a great job at this if you give it the last three paragraphs)

packages/frontend/src/components/frames/FramePreview.tsx
```diff
@@ -0,0 +1,89 @@
+import { Frame } from "@blurple-canvas-web/types";
+import { styled } from "@mui/material/styles";
+import { clamp, normalizeFrameBounds } from "@/util";
+
+export const FRAME_FILL_RATIO = 0.9;
+
+export type SourceRect = {
```
jaskfla: `interface`

packages/frontend/src/contexts/SelectedBoundsContext.tsx
```diff
@@ -0,0 +1,182 @@
...
+          getCurrentViewBounds({
+            canvas,
+            containerRef,
+            offset,
+            zoom,
+          }),
+          fillRatio,
+          canvas,
+        ),
+      );
```
jaskfla: Nitpick but this would be a lot easier to read as three lines

packages/frontend/src/contexts/SelectedBoundsContext.tsx
```diff
@@ -0,0 +1,182 @@
...
+      value={{
+        canEdit,
+        minHeight,
+        minWidth,
+        selectedBounds,
+        clearSelectedBounds,
+        setSelectedBounds,
+        setBoundsToCurrentView,
+        setCanEdit,
+        setMinimumBounds,
```
jaskfla, line 174: Looks like all of these are referentially stable—nice!

packages/frontend/src/util/index.ts
```diff
@@ -1,8 +1,22 @@
...
+  DiscordUserProfile,
+  Frame,
+  GuildData,
+  PixelColor,
+} from "@blurple-canvas-web/types";
 import { DateTime } from "luxon";
 
 export { default as createPixelUrl } from "./searchParams";
 
+export type ViewBounds = {
```
jaskfla: `interface`

packages/frontend/src/util/index.ts
```diff
@@ -73,3 +87,35 @@ export function getUserGuildFlags(
...
+
+export function hexStringToPixelColor(hex: string | null): PixelColor | null {
+  if (hex === null) {
+    return null;
+  }
+
+  const normalized = hex.startsWith("#") ? hex : `#${hex}`;
+  if (!/^#([0-9A-Fa-f]{6})$/.test(normalized)) {
+    return null;
+  }
```
jaskfla, line 114:
```ts
  if (hex === null || !/^#?([0-9A-Fa-f]{6})$/.test(hex)) {
    //                    └─ This ‘?’ handles the `normalised` logic
    return null;
  }
```

and below, the channels can be sliced with negative indexing so it’s `#`-agnostic: `r = hex.slice(-6, -4)`

packages/frontend/src/util/index.ts
```diff
@@ -73,3 +87,35 @@ export function getUserGuildFlags(
...
+    left,
+    right,
+    top,
+    bottom,
+    width: right - left,
+    height: bottom - top,
+  };
+}
+
+export function hexStringToPixelColor(hex: string | null): PixelColor | null {
```
jaskfla, line 107: Unhinged and this was never the right call (even if wasn’t “too complex” for the TS compiler), but technically TS syntax would allow us to use this instead of a string 🤓

```ts
type HexDigit = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | 'a' | 'b' | 'c' | 'd' | 'e' | 'f';
type RgbHexStr = `${'#' | ''}${HexDigit}${HexDigit}${HexDigit}${HexDigit}${HexDigit}${HexDigit}`;
```

jaskfla (changes requested):
Revoking previous pre-approval, sorry. Code gets easier to review as it gets cleaned up so spotting things that flew under the radar previously!

packages/frontend/src/util/index.ts
S-Ou: yikes no thanks

that would just mean we'd have to do the regex validation elsewhere

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const guildIds = Object.keys(user?.guilds ?? {});
+  const { data: guildFrames = [] } = useGuildFrames({
+    canvasId: canvas.id,
+    guildIds: guildIds,
+  });
+
+  const groupedByOwnerId = Object.groupBy(
+    guildFrames,
+    (f) => f.owner.guild.guild_id,
```
S-Ou:
```suggestion
  const { data: sortedGuildFrameMap = [] } = useGuildFrames(
    {
      canvasId: canvas.id,
      guildIds: guildIds,
    },
    {
      select: selectSortedGuildFrameMap,
    },
```

packages/backend/src/routes/api/v1/frame.ts
Rocked03: I've made this an issue, would be useful. Currently working on stuff in my Admin PR that just do things in the endpoint rather than a middleware

https://github.com/project-blurple/Canvas-Web/issues/457

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -17,6 +21,9 @@ const FramesWrapper = styled("div")`
   flex-direction: column;
 `;
 
+type GuildFrames = FrameRequest.GuildFramesResBody;
+type SortedGuildFrameMap = [string, GuildOwnedFrame[]][];
```
jaskfla: This isn’t a map 👀

```suggestion
type SortedGuildFrameEntries = [string, GuildOwnedFrame[]][];
```

packages/frontend/src/util/index.ts
jaskfla: Yeah there’s definitely such thing as _too much TypeScript_, and this would be that

jaskfla (approved):
Looking solid. I don’t feel the need to look any more closely. Re-approving to unblock so you can merge in your own time once you’re satisfied

PR #422, opened 2026-04-18
fix: missing csrf middleware & clear text transmission of sensitive cookie

Rocked03:
Fixing https://github.com/project-blurple/Canvas-Web/security/code-scanning/4 and https://github.com/project-blurple/Canvas-Web/security/code-scanning/5

### DIFF
--- a/packages/backend/src/middleware/auth.ts
+++ b/packages/backend/src/middleware/auth.ts
@@ -126,6 +126,8 @@ export function initializeAuth(app: Express) {
     session({
       cookie: {
         maxAge: 24 * 60 * 60 * 1000, // 1 day (in ms)
+        sameSite: "lax",
+        secure: config.environment !== "development",
       },
       // having a random secret would mess with persistent sessions
       secret: config.expressSessionSecret,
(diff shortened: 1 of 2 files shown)


Rocked03:
Copilot wrote this up because I don't really have much clue what this is all about, so wouldn't mind another set of eyes on this 😅

packages/backend/src/middleware/auth.ts
```diff
@@ -1,18 +1,79 @@
...
+function ensureCsrfProtection(req: Request, res: Response, next: NextFunction) {
+  if (!req.session.csrfToken) {
+    req.session.csrfToken = randomBytes(32).toString("hex");
+  }
+
+  res.cookie(csrfCookieName, req.session.csrfToken, {
+    httpOnly: false,
+    sameSite: "lax",
+    secure: config.environment !== "development",
+    path: "/",
```
pumbas600, line 48: There's a few security vulnerabilities here—The CSRF token should always be `httpOnly: true`, as the frontend isn't meant to be able to access it. `sameSite` should also be strict.

This will break in our local environment though. Ideally, we'll want to setup our local development environment to closely mirror production: Running on HTTPS with the backend on the "same" domain... I'm not 100% sure the best way to do this, but I suspect we'll want to have Caddy running locally as a reverse proxy like we do in production.

pumbas600 (changes requested):

packages/backend/src/middleware/auth.ts
Rocked03: That part fixed a medium severity thing but the rest of it is a high severity thing that's blocking builds

Rocked03:
I'm going to close this for now, and can revisit later if it becomes an issue

PR #424, opened 2026-04-19
feat: database seeding

### LINKED ISSUE #381: Setup seed script which can be used to initialise a local database
A simple script that can populate a local database with enough data to get the project running locally.

Rocked03:
With the help of @volcanofr :)

Have had to change up the frame SQL fetching because the doubled `owner_id` relation in the schema was causing issues with the seeding


packages/backend/prisma.config.ts
```diff
@@ -1,11 +1,16 @@
...
 
+dotenvx.config();
+
 export default defineConfig({
   schema: "prisma/schema.prisma",
   views: {
     path: "prisma/views",
   },
+  migrations: {
+    seed: "ts-node --esm prisma/seed.ts",
```
jaskfla: I admittedly haven’t yet looked at `seed.ts`, but could we get away with `node --experimental-strip-types`?

Would be nice to lessen our dependence on `ts-node` over time rather than increase

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,220 @@
...
+  participationSeedData,
+  pixelSeedDataBatches,
+  userSeedData,
+} from "./seedData/index.ts";
+
+const prisma = new PrismaClient({
+  adapter: new PrismaPg(process.env.DATABASE_URL ?? ""),
+});
+
+const seedStartedAt = Date.now();
```
jaskfla: `Date` interface for timing has a couple edge case issues (the details of which I forget)

Can probably leave out of this PR, but would be good to use [`Performance`](https://developer.mozilla.org/en-US/docs/Web/API/Performance) here

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,220 @@
...
+  console.log(`[+${elapsedMs}ms] ${message}`);
+}
+
+async function runSeedingStep(
+  step: string,
+  action?: () => Promise<void>,
+): Promise<void> {
+  const startedAt = Date.now();
+  logWithTiming(`Seeding ${step}...`);
+  if (action) await action();
```
jaskfla:
```suggestion
  await action?.();
```

packages/backend/src/seed/index.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  if (seedings.length === 0) {
+    logWithTiming("No seedings to run");
+    return;
+  }
+
+  logWithTiming(`Seedings to run: ${seedings.join(", ")}`);
+
+  const order: Seeding[] = [
```
jaskfla: Flagging (for myself) to dig up code that lets us dynamically figure out valid topological order

(Also not for this PR)

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,220 @@
...
+    "frame",
+    "canvas",
+    "user",
+    "event",
+    "discord_guild_record",
+    "discord_user_profile",
+    "color",
+  ];
+  await runSeedingStep("cleanup", async () => {
+    await prisma.$transaction([
```
jaskfla: Appreciate that we’re awaiting promises all the way up/down the stack

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,220 @@
...
+      ...seedings
+        .sort((a, b) => order.indexOf(a) - order.indexOf(b))
+        .map((seeding) => prisma[seeding].deleteMany()),
+    ]);
```
jaskfla: I don’t think this is doing what is intended:

- this is spreading the result of `seedings.sort().map()`; but
- I suspect you actually want `$transaction(seedings.toSorted().map())`

Otherwise this spread into a new array is completely redundant

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,220 @@
...
+  logWithTiming("Database seed completed");
+}
+
+main()
+  .then(async () => await prisma.$disconnect())
+  .catch(async (e) => {
+    console.error(e);
+    await prisma.$disconnect();
+    process.exit(1);
+  });
```
jaskfla: We can `await` at top level

packages/backend/prisma/seedData/guilds.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  const colorIds = colorSeedData
+    .filter((color) => !color.global)
+    .filter(
+      (color) =>
+        !participations.some(
+          // filtering out the ones already hardcoded above
+          (participation) => participation.color_id === color.id,
+        ),
+    )
```
jaskfla: `.filter().filter()` chain could probably be `.filter(a && b)`

jaskfla (commented):
Exciting stuff—just a cursory glance over, not a real review

scripts/anonymize_user_ids.py
jaskfla, line 1: Oh, and if you wouldn’t mind configuring `settings.json` to format Python with [Black](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter)

jaskfla: And if we’re interested in trying some newfangled cool shit, there’s always [Mojo](https://www.modular.com/open-source/mojo)

jaskfla:
@Rocked03 Should this be tested in CI?

Rocked03:
> @Rocked03 Should this be tested in CI?

I suppose it'd probably be good to? Although it's not necessarily a small seeding...

packages/backend/prisma/seed-data/colors.ts
```diff
@@ -0,0 +1,337 @@
+interface ColorSeedData {
```
pumbas600: Is it possible to derive these types using the ones provided by Prisma?

package.json
```diff
@@ -31,7 +31,9 @@
     "check": "pnpm --recursive check",
     "check:fix": "pnpm --recursive check:fix",
     "prepare": "husky",
-    "clean": "pnpm --recursive clean && rm -rf node_modules/"
+    "clean": "pnpm --recursive clean && rm -rf node_modules/",
+    "prisma:push": "pnpm --filter=backend prisma:push",
+    "prisma:seed": "pnpm --filter=backend prisma:seed --"
```
pumbas600, line 36: Out of curiosity, why is the `--` needed at the end here?

pumbas600: Same in the backend `package.json`?

packages/backend/prisma/schema.prisma
```diff
@@ -91,21 +89,17 @@ model event {
...
 model frame {
-  id             String                @id
+  id             String  @id
   canvas_id      Int
-  canvas         canvas                @relation(fields: [canvas_id], references: [id])
+  canvas         canvas  @relation(fields: [canvas_id], references: [id])
   owner_id       BigInt
-  owner_user     discord_user_profile? @relation(fields: [owner_id], references: [user_id], map: "user")
-  owner_guild    discord_guild_record? @relation(fields: [owner_id], references: [guild_id], map: "guild")
   is_guild_owned Boolean
```
pumbas600, line 96: I'm not sure how I feel about this change. From my experience, I think having the foreign key is better practice

.github/workflows/ci.yml
```diff
@@ -42,6 +42,50 @@ jobs:
       - name: Check linting
         run: pnpm lint
 
+  prisma_seed:
```
jaskfla: Maybe in its own workflow?

.vscode/settings.json
```diff
@@ -64,6 +64,9 @@
     "**/dist": true,
     "**/node_modules": true
   },
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter"
+  },
   "[prisma]": {
     "editor.defaultFormatter": "Prisma.prisma"
   }
```
jaskfla: `py` before `pr`? 🧐

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,224 @@
...
+    "frame",
+    "guild",
+    "history",
+    "info",
+    "participation",
+    "pixel",
+    "user",
+  ] as const;
+  type Seeding = (typeof allSeedings)[number];
+  let seedings: Seeding[] = [...allSeedings];
```
jaskfla: Should this be a set? Lotta `includes()` checks

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+    seedings = filteredSeedings;
+  }
+
+  if (seedings.length === 0) {
+    logWithTiming("No seedings to run");
+    return;
+  }
+
+  logWithTiming(`Seedings to run: ${seedings.join(", ")}`);
```
jaskfla: Completely optional here, but nice opportunity for a reminder about `Intl.ListFormat`

```suggestion
  const formatter = new Intl.ListFormat();
  logWithTiming(`Seedings to run: ${formatter.format(seedings.join)}`);
```

In this case no arguments needed to constructor, but it defaults to the equivalent of `new Intl.ListFormat(undefined, { type: conjunction }`

packages/backend/prisma/seed.ts
```diff
@@ -0,0 +1,224 @@
...
+  type Seeding = (typeof allSeedings)[number];
+  let seedings: Seeding[] = [...allSeedings];
+
+  if (!OVERWRITE) {
+    const filteredSeedings: Seeding[] = [];
+    for (const seeding of seedings) {
+      const count = (await prisma[seeding].count()) as
+        | number
+        | undefined
+        | null;
```
pumbas600: Out of curiosity, why do we need this `as` cast?

pumbas600 (changes requested):
I've not had a chance to pull this down and take a look locally, but I just had a few questions :)

packages/backend/prisma/seed-data/colors.ts
Rocked03: I tried but since seeding mostly happens outside of the build, it was just acting funky

package.json
Rocked03: so that we can pass `--overwrite=true` through, and in order to add params like that you need to add the `--`

packages/backend/prisma/schema.prisma
Rocked03: I'm not a huge fan either but seeding just simply wouldn't resolve a foreign relation that didn't exist, ie for a frame with owner id `123`, it expected both a `discord_user_profile` AND `discord_guild_record` entry with ID `123`

packages/backend/src/services/frameService.ts
```diff
@@ -47,7 +34,84 @@ async function findFrameForType(frameId: string) {
...
+  guild_id: bigint;
+  name: string;
+};
+
+type OwnerLookup = {
+  usersById: Map<bigint, UserOwnerRecord>;
+  guildsById: Map<bigint, GuildOwnerRecord>;
+};
+
+function collectOwnerIds(frames: FrameDbRecord[]) {
```
jaskfla: Collect, or partition?

```suggestion
function partitionOwnerIds(frames: FrameDbRecord[]) {
```

packages/backend/src/services/frameService.ts
```diff
@@ -47,7 +34,84 @@ async function findFrameForType(frameId: string) {
...
+            in: userIds,
+          },
+        },
+        select: {
+          user_id: true,
+          username: true,
+          profile_picture_url: true,
+        },
+      })
+    : Promise.resolve([] as UserOwnerRecord[]),
```
jaskfla: Technically redundant `Promise.resolve`, but no strong opinion here

```suggestion
    : [] as UserOwnerRecord[],
```

packages/backend/prisma/seed.ts
S-Ou: no idea, this is from volcano's code. i've removed it and it seems fine

package.json
jaskfla: Should that be the consumer’s responsibility, so that we retain the _option_ to add more flags?

```sh
pnpm run prisms:seed -- params to forward
```

I don’t have enough familiarity with this code to have an opinion either way[^foot]

[^foot]: (But I _do_ have an option that the `clean` script which I suspect I wrote would be improved as `rm -rf -- node_modules/` 😅

Rocked03: I don't have any strong opinion here, the intent was really just to get `pnpm prisma:seed --overwrite=true` to work

packages/backend/prisma/seed-data/colors.ts
jaskfla: wrt `rgba`, Prisma will never give us a tuple AFAIK

Which, I mean, fair enough—pretty sure psql doesn’t have a comparable data structure

packages/backend/prisma/seedData/colors.ts
```diff
@@ -0,0 +1,337 @@
...
+  code: string;
+  emoji_name: string;
+  emoji_id: bigint;
+  global: boolean;
+  name: string;
+  rgba: [number, number, number, number];
+}
+
+// These are all the colors available in the 2024 event
+export const colorSeedData: ColorSeedData[] = [
```
jaskfla: This type overrides makes the `as const` useless

Needs to be `as const satisfies ColorSeedData[]` and lose this type or it just widens back out

packages/backend/src/services/frameService.ts
```diff
@@ -153,8 +229,10 @@ export async function getFramesByGuildIds(
     select: frameSelect,
   });
 
-  return frames.map((frame) => {
-    const mapped = frameFromDb(frame);
+  const owners = await loadOwnerLookup(frames);
+
+  return frames.map((frame: FrameDbRecord) => {
+    const mapped = frameFromDb(frame, owners);
     asGuildFrame(mapped);
```
jaskfla, line 230: Is this an `assert` type guard? Might need renaming to `assertIsGuildFrame`

Rocked03: not quite, it's doing a conversion

packages/backend/src/seed/data/colors.ts
jaskfla, line 1: We use `kebab-case` dirnames, right?

packages/backend/prisma/seed-data/events.ts
```diff
@@ -0,0 +1,73 @@
...
+  title: "Canvas Dev",
+  canvas_admin: [708540954302218311n],
+  current_event_id: 2034,
+  cached_canvas_ids: [2024, 2034],
+  admin_server_id: 412754940885467146n,
+  current_emoji_server_id: 412754940885467146n, // This is for the bot, not used by the web app
+  host_server_id: 412754940885467146n,
+  default_canvas_id: 2034,
+  all_colors_global: false,
+};
```
jaskfla: Couple more instances of possible `as const`s in this file

```suggestion
export const infoSeedData = {
  title: "Canvas Dev",
  canvas_admin: [708540954302218311n],
  current_event_id: 2034,
  cached_canvas_ids: [2024, 2034],
  admin_server_id: 412754940885467146n,
  current_emoji_server_id: 412754940885467146n, // This is for the bot, not used by the web app
  host_server_id: 412754940885467146n,
  default_canvas_id: 2034,
  all_colors_global: false,
} as const satisfies InfoSeedData;
```

packages/backend/prisma/seed-data/frames.ts
```diff
@@ -0,0 +1,101 @@
...
+  id: string;
+  canvas_id: number;
+  owner_id: bigint;
+  is_guild_owned: boolean;
+  name: string;
+  x_0: number;
+  x_1: number;
+  y_0: number;
+  y_1: number;
+  style_id: number; // Used by the bot, not currently supported in the web app
```
jaskfla: Let it flow through to consumers

```suggestion
  /** @privateRemarks Used by the bot, not currently supported in the web app */
  style_id: number;
```

package.json
pumbas600: Ah that's a good point, I'm happy to keep it as-is

packages/backend/prisma/schema.prisma
pumbas600: Oh right of course! 

Yes, you should have two separate columns in your database, one for the `owner_user_id` and another the `owner_guild_id`. I think that's a better way of addressing this issue

That would actually also make `is_guild_owned` redundant too. I'm not sure how much of this you want to do in this PR though

Rocked03: out of scope for this PR yeah, i've made an issue instead https://github.com/project-blurple/Canvas-Web/issues/426, feel free to add/change details in it

README.md
```diff
@@ -87,6 +87,18 @@ The **[backend](/packages/backend/.env.example)** and **[frontend](/packages/fro
...
+**Set up the database.** Make sure you have a PostgreSQL database running, and that the `DATABASE_URL` environment variable in `packages/backend/.env` is set correctly. Then, run this command from the monorepo root to create the database schema:
+
+```sh
+pnpm prisma:push
+```
+
+You can also seed the database with test data:
+
+```sh
+pnpm prisma:seed
```
hwan513, line 108: Should also mention that it requires a backend build to be able to run the generation script. Should also be in the package specific readme as well probably.

Was able to get it working with prisma dev.
Something along the lines of: 
```
pnpm install
cd packages/backend
pnpm prisma dev
# copy the output env generated by the URL
# DATABASE_URL="postgres://postgres:postgres@localhost:51214/template1?sslmode=disable&connection_limit=10&connect_timeout=0&max_idle_connection_lifetime=0&pool_timeout=0&socket_timeout=0"
pnpm run build
pnpm run prisma:push
pnpm run prisma:seed
```

Rocked03: oh, thanks for that-- i've added it to the readme

packages/backend/prisma/apply-views.ts
```diff
@@ -0,0 +1,62 @@
+import { spawnSync } from "node:child_process";
+import { readFile } from "node:fs/promises";
+import { fileURLToPath } from "node:url";
+
+const viewDefinitions = [
+  ["public", "most_frequent_color"],
+  ["public", "color_place_frequency"],
```
volcanofr, line 6: `percentile_cont` does not works for me. Or at least, `median_time_diff` is not correctly defined.
I've simply removed it locally to make it work.

Rocked03: What version of PostgreSQL are you running? `percentile_cont` was introduced in Postgres 9.4.

packages/backend/prisma/apply-views.ts
```diff
@@ -0,0 +1,62 @@
+import { spawnSync } from "node:child_process";
+import { readFile } from "node:fs/promises";
+import { fileURLToPath } from "node:url";
+
+const viewDefinitions = [
```
volcanofr, line 4: it would be nice to automatically detect files from `./views` (and be able to exclude them, from a specific directory or with a prefix/suffix)

packages/backend/prisma/apply-views.ts
Rocked03: That's a slight oversight on my part, I've added to the README specifying that we require v9.4+

volcanofr: `SELECT version();` gave:
> PostgreSQL 17.5 on x86_64-windows, compiled by msvc-19.44.35209, 64-bit

packages/backend/prisma/apply-views.ts
Rocked03: I'm not sure if this is the right approach? Views are hardcoded in, so there isn't really any benefit to dynamically loading them in. Also, we need to pay attention to the orders that the views are loaded, as some reference others, and both dynamic loading and excluding views might break this.

packages/backend/prisma/apply-views.ts
Rocked03: Can you please share your error? I can't repro this

packages/backend/prisma/apply-views.ts
volcanofr: Yeah you're right, I forgot the order...

packages/backend/prisma/apply-views.ts
volcanofr: Can't repro anymore 💀, was it just my PC randomly kill processes?
And now have `median_time_diff` nice!

packages/backend/package.json
```diff
@@ -36,7 +36,9 @@
     "lint:fix": "biome check --formatter-enabled=false --write .",
     "check": "pnpm format && pnpm lint",
     "check:fix": "pnpm format:fix && pnpm lint:fix",
-    "clean": "rm -rf node_modules/ build/"
+    "clean": "rm -rf node_modules/ build/",
+    "prisma:push": "prisma db push && node prisma/apply-views.mjs",
```
pumbas600: Could we use `node --experimental-strip-types` like we are with `seed.ts` and make `apply-views.mjs` a ts file?

packages/backend/src/seed/index.ts
pumbas600, line 1: What do you think about moving this into a seed folder in `src`? I've done a quick test locally and I believe you can then access the prisma types. That'll save you needing to essentially write every model twice: once in prisma, and again in these seed files.

packages/backend/README.md
```diff
@@ -6,6 +6,32 @@
...
+
+Set `DATABASE_URL` in `.env` to a running PostgreSQL instance.
+
+Our schema and views require PostgreSQL 9.4 or newer.
+
+If you want Prisma to spin up a local development database for you, run:
+
+```sh
+pnpm prisma dev
+# Copy the DATABASE_URL shown in the output into .env
```
pumbas600: Oh cool. I didn't know about this!

packages/backend/prisma/apply-views.mjs
```diff
@@ -0,0 +1,62 @@
...
+      `DROP VIEW IF EXISTS ${quoteIdentifier(schema)}.${quoteIdentifier(viewName)} CASCADE;`,
+      `CREATE VIEW ${quoteIdentifier(schema)}.${quoteIdentifier(viewName)} AS`,
+      selectQuery.trim(),
+      ";",
+    ].join("\n");
+
+    const result = spawnSync(
+      process.execPath,
+      [prismaCliPath, "db", "execute", "--stdin"],
+      {
```
pumbas600: It feels a little bit fragile to be referencing the prisma cli path directly. Could we do:

```suggestion
    const result = spawnSync(
      "pnpm",
      ["prisma, "db", "execute", "--stdin"],
      {
```

packages/backend/prisma/views/public/leaderboard_guild.sql
```diff
@@ -13,7 +12,14 @@ SELECT
 FROM
   history
 WHERE
-  history.user_id NOT IN (SELECT user_id FROM blacklist)
```
pumbas600, line 16: Out of curiosity, why did you change this?

pumbas600 (dismissed):
Looking really good! I tested it out locally and it worked like a charm!

I just have a few suggestions which should hopefully make maintainance easier going forwards :)

packages/backend/README.md
S-Ou: hm, i think copilot hallucinated this 😭

packages/backend/prisma/views/public/leaderboard_guild.sql
Rocked03: the views were acting up a bit and this was an attempt to fix it that i never reverted

packages/backend/src/seed/index.ts
Rocked03: i've refactored it, needed to ignore some import warnings but it works

packages/backend/src/seed/index.ts
```diff
@@ -0,0 +1,284 @@
...
+        return prisma.history.deleteMany();
+      case "info":
+        return prisma.info.deleteMany();
+      case "participation":
+        return prisma.participation.deleteMany();
+      case "pixel":
+        return prisma.pixel.deleteMany();
+      case "user":
+        return prisma.user.deleteMany();
+    }
```
jaskfla: Probably doesn’t need to be a function

```suggestion
    return prisma[seeding].deleteMany();
```

packages/backend/src/seed/index.ts
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  if (!OVERWRITE) {
+    for (const seeding of allSeedings) {
+      const count = await countRecords(seeding);
+      if (count && count >= 1) {
+        seedings.delete(seeding);
+      }
+    }
```
jaskfla: Optional, but I’d be in favour of `Promise.all(allSettings.map())` here

packages/backend/src/seed/data/pixels.ts
```diff
@@ -0,0 +1,186 @@
...
+  "data",
+  "pixelData2024.csv",
+);
+const historySeedDataPath = path.join(
+  process.cwd(),
+  "src",
+  "seed",
+  "data",
+  "historyData2024.csv",
+);
```
jaskfla: `process.cwd(), 'src/seed/data'` is duplicated between these two

packages/backend/README.md
pumbas600: It seemed to work for me!

packages/backend/src/seed/data/pixels.ts
```diff
@@ -0,0 +1,186 @@
...
+
+function parseHistorySeedData(line: string): Prisma.historyCreateManyInput {
+  const [userId, x, y, colorId, timestamp] = line.split(",");
+
+  return {
+    user_id: BigInt(userId),
+    canvas_id: 2024,
+    x: Number(x),
+    y: Number(y),
+    color_id: Number(colorId),
```
jaskfla: These are always integers, right?

```suggestion
    x: Number.parseInt(x, 10),
    y: Number.parseInt(y, 10),
    color_id: Number.parseInt(colorId, 10),
```

packages/backend/src/seed/index.ts
```diff
@@ -0,0 +1,284 @@
...
+      await prisma.guild.createMany({ data: guildSeedData() });
+    });
+  }
+
+  // ====== Color data ======
+
+  // COLOR
+  if (seedings.has("color")) {
+    await runSeedingStep("color", async () => {
+      await prisma.color.createMany({ data: [...colorSeedData] });
```
jaskfla: Why copy the array? Does `createMany` have side effects?

packages/backend/src/seed/index.ts
```diff
@@ -0,0 +1,284 @@
...
+  const userData = discordUserProfileSeedData();
+
+  // === DISCORD_USER_PROFILE ===
+  if (seedings.has("discord_user_profile")) {
+    await runSeedingStep("discord_user_profile", async () => {
+      await prisma.discord_user_profile.createMany({
+        data: userData,
+      });
+    });
+  }
```
jaskfla: This is duplicated quite a lot. Feels like we could have something like

```ts
for (const s of seedings) {
	const data =
		typeof seedData[s] === 'function' ? seedData[s]() : seedData[s];
	await runSeedingStep(
		s,
		async () => void (await prisma[s].createMany({ data }))
	);
}
```

…where `seedData` is a map (could be a `Map` if you like, but I reckon a `const satisfies Records<Seeding, Array | () => Array>`

packages/backend/src/seed/index.ts
```diff
@@ -0,0 +1,284 @@
...
+
+main()
+  .then(async () => {
+    await prisma.$disconnect();
+  })
+  .catch(async (e) => {
+    console.error(e);
+    await prisma.$disconnect();
+    process.exit(1);
+  });
```
jaskfla: `try`–`catch` with `async–await`

No need to be averse to top-level awaits anymore

packages/backend/src/services/frameService.ts
```diff
@@ -47,7 +34,80 @@ async function findFrameForType(frameId: string) {
...
+function partitionOwnerIds(frames: FrameDbRecord[]) {
+  const userIds = new Set<bigint>();
+  const guildIds = new Set<bigint>();
+
+  for (const frame of frames) {
+    if (frame.is_guild_owned) {
+      guildIds.add(frame.owner_id);
+    } else {
+      userIds.add(frame.owner_id);
+    }
```
jaskfla: Optional:

```suggestion
    (frame.is_guild_owned ? guildIds : userIds).add(frame.owner_id);
```

```suggestion
    const partition = frame.is_guild_owned ? guildIds : userIds;
    partition.add(frame.owner_id);
```

packages/backend/src/seed/index.ts
jaskfla: Looking at it, could change `runSeedingStep` signature to `runSeedinStep(s, seedData[s])`

Rocked03: not all the seedings shared the same flow so i've had to do a slightly different refactor, but it does still tidy it up a decent bit

Rocked03:
Have discussed off-platform, will force-merge this PR

PR #432, opened 2026-04-21
fix: zoom drift 🏎️

### LINKED ISSUE #427: Zoom gets miscalibrated when switching between canvases
Can't reliably repro, but tends to happen when you switch canvas

https://github.com/user-attachments/assets/f62a27de-bb89-414d-b3cb-c19cad581475

hwan513:
Stupid stupid code, I wonder who wrote this code. Oh wait, it was me.

Issue was caused in the previous version by the most recent `containerRef.current` being in the `useEffect`, so when the cleanup ran it would run on `undefined // containerRef.current` sometimes, causing the component to have 2 event handlers. Fixed by capturing the current container as a const in the `useEffect`.

Resolves #427.


pumbas600 (approved):
Of course 🤦 

I would 100% expect this to be picked up by some kind of React eslint rule :(

Good solve though!

PR #435, opened 2026-04-22
feat: psql automated backup script

### LINKED ISSUE #394: SQL Backup Script
Since the hosting server gets deconstructed and reconstructed every year, the data needs to be backed up. This is currently done manually, but it'd be helpful to have a helper script that streamlines this process

Also would be good if we could automate backups for security

Rocked03:
Just a draft for now, not sure if this is the right direction. 

Addresses #394, which is basically an interest in having some automated SQL backup solution


scripts/backup/backup-full.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# Creates a full database backup (schema + all data).
+set -euo pipefail
```
jaskfla: I’d prefer not to use this

Just `set -ex`

scripts/backup/backup-full.sh
```diff
@@ -0,0 +1,33 @@
...
+
+: "${PGHOST:?PGHOST is required}"
+: "${PGPORT:=5432}"
+: "${PGUSER:?PGUSER is required}"
+: "${PGPASSWORD:?PGPASSWORD is required}"
+: "${PGDATABASE:?PGDATABASE is required}"
+
+BACKUP_DIR="${BACKUP_DIR:-/backups}"
+RETENTION_DAYS="${RETENTION_DAYS:-14}"
```
jaskfla: RHS of assignment is already a double-quoted context. Optional, but I personally have a slight preference not to use redundant quotes

scripts/backup/backup-full.sh
```diff
@@ -0,0 +1,33 @@
...
+
+mkdir -p "${BACKUP_DIR}"
```
jaskfla: Similarly with redundant braces

```suggestion
mkdir -p "$BACKUP_DIR"
```

scripts/backup/backup-full.sh
```diff
@@ -0,0 +1,33 @@
...
+
+ts="$(date +%F_%H-%M-%S)"
+out="${BACKUP_DIR}/db-full-${ts}.sql"
+
+pg_dump \
```
jaskfla: Good call with the line breaks

scripts/backup/backup-full.sh
```diff
@@ -0,0 +1,33 @@
...
+  --format=plain \
+  --encoding=UTF8 \
+  --no-owner \
+  --no-privileges \
+  --file="${out}"
+
+gzip -f "${out}"
+echo "$(date -Is) Backup created: ${out}.gz"
+
+find "${BACKUP_DIR}" -type f -name "db-*.sql.gz" -mtime +"${RETENTION_DAYS}" -delete || true
```
jaskfla: Flagging for myself to look at this more closely and see if there’s a less, uh, Bashy way to handle nonzero exit code here

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,103 @@
...
+RUN_UNLOCKED_ON_START="${RUN_UNLOCKED_ON_START:-${RUN_DATA_ON_START:-true}}"
+RUN_FULL_ON_START="${RUN_FULL_ON_START:-false}"
+RUN_SCHEMA_WITH_FULL="${RUN_SCHEMA_WITH_FULL:-true}"
+
+mkdir -p "${BACKUP_DIR}" "${STATE_DIR}"
+export TZ="${TZ:-UTC}"
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+if ! [[ "${UNLOCKED_INTERVAL_MINUTES}" =~ ^[0-9]+$ ]] || [ "${UNLOCKED_INTERVAL_MINUTES}" -le 0 ]; then
```
jaskfla: Presuming we’ll only ever run these scripts in Unix, let’s just commit using Bash and just use `(( UNLOCKED_INTERVAL_MINUTES < 0 ))`

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,103 @@
...
+  "${SCRIPT_DIR}/backup-full.sh"
+}
+
+run_schema_backup() {
+  "${SCRIPT_DIR}/backup-schema.sh"
+}
+
+run_unlocked_backup() {
+  "${SCRIPT_DIR}/backup-unlocked.sh"
+}
```
jaskfla, line 152: This feels slightly odd?

Why not just run the script directly? Or even just define them in here as functions in one big `.sh` file

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,103 @@
...
+      full_pid=$!
+      run_schema_backup &
+      schema_pid=$!
+      wait "${full_pid}"
+      wait "${schema_pid}"
+    else
+      run_full_backup
+    fi
+    printf '%s\n' "${now_epoch}" > "${STATE_DIR}/last_full_epoch"
+    printf '%s\n' "${now_epoch}" > "${STATE_DIR}/last_unlocked_epoch"
```
jaskfla: Why `echo` sometimes and `printf` other times?

scripts/backup/db-backup.sh
jaskfla, line 1: `SCREAMING_SNAKES` for env vars and `inside_voice_snakes` for local variables please 🙏

scripts/backup/backup-unlocked.sh
```diff
@@ -0,0 +1,78 @@
...
+  echo "SET statement_timeout = 0;"
+  echo "SET lock_timeout = 0;"
+  echo "SET idle_in_transaction_session_timeout = 0;"
+  echo "SET client_encoding = 'UTF8';"
+  echo "SET standard_conforming_strings = on;"
+  echo "SELECT pg_catalog.set_config('search_path', '', false);"
+  echo "SET check_function_bodies = false;"
+  echo "SET xmloption = content;"
+  echo "SET client_min_messages = warning;"
+  echo "SET row_security = off;"
```
jaskfla: All but line 23 can use single quotes

jaskfla (changes requested):

scripts/backup/db-backup.sh
```diff
@@ -8,96 +8,197 @@ set -euo pipefail
...
+export TZ=${TZ:-UTC}
+
+if ! [[ "$UNLOCKED_INTERVAL_MINUTES" =~ ^[0-9]+$ ]] || (( UNLOCKED_INTERVAL_MINUTES < 0 )); then
+  echo 'UNLOCKED_INTERVAL_MINUTES must be a positive integer.' >&2
   exit 1
 fi
 
-if ! [[ "${FULL_INTERVAL_MINUTES}" =~ ^[0-9]+$ ]] || [ "${FULL_INTERVAL_MINUTES}" -le 0 ]; then
-  echo "FULL_INTERVAL_MINUTES must be a positive integer." >&2
+if ! [[ "$FULL_INTERVAL_MINUTES" =~ ^[0-9]+$ ]] || (( FULL_INTERVAL_MINUTES < 0 )); then
```
jaskfla, line 29: The first test here should be redundant. I would assume `(( ))` chokes if given something that can’t be parsed as an integer

scripts/backup/db-backup.sh
```diff
@@ -8,96 +8,197 @@ set -euo pipefail
...
+      psql -h "$PGHOST" -p "$PGPORT" -U "$PGUSER" -d "$PGDATABASE" -v ON_ERROR_STOP=1 -c "COPY (SELECT $cols FROM public.$table WHERE $where_clause) TO STDOUT"
+      echo '\.'
+    } >> "$backup_file"
+  }
+
+  append_copy_block 'canvas' "id IN ($unlocked_canvas_ids)"
+  append_copy_block 'pixel' "canvas_id IN ($unlocked_canvas_ids)"
+  append_copy_block 'history' "canvas_id IN ($unlocked_canvas_ids)"
+  append_copy_block 'cooldown' "canvas_id IN ($unlocked_canvas_ids)"
+  append_copy_block 'frame' "canvas_id IN ($unlocked_canvas_ids)"
```
jaskfla:
```suggestion
  append_copy_block canvas "id IN ($unlocked_canvas_ids)"
  append_copy_block pixel "canvas_id IN ($unlocked_canvas_ids)"
  append_copy_block history "canvas_id IN ($unlocked_canvas_ids)"
  append_copy_block cooldown "canvas_id IN ($unlocked_canvas_ids)"
  append_copy_block frame "canvas_id IN ($unlocked_canvas_ids)"
```

scripts/backup/db-backup.sh
```diff
@@ -8,96 +8,197 @@ set -euo pipefail
...
-  now_epoch="$(date +%s)"
+  now_epoch=$(date +%s)
 
-  last_full_epoch="$(cat "${STATE_DIR}/last_full_epoch" 2>/dev/null || echo 0)"
+  last_full_epoch=$(cat "$STATE_DIR/last_full_epoch" 2>/dev/null || echo 0)
   full_interval_seconds=$((FULL_INTERVAL_MINUTES * 60))
-  if [ $((now_epoch - last_full_epoch)) -ge "${full_interval_seconds}" ]; then
-    if [ "${RUN_SCHEMA_WITH_FULL}" = "true" ]; then
+  if [ $((now_epoch - last_full_epoch)) -ge "$full_interval_seconds" ]; then
+    if [ "$RUN_SCHEMA_WITH_FULL" = 'true' ]; then
```
jaskfla: I’d encourage using `[[ ]]` everywhere over `[ ]`

jaskfla (approved):
Re-reviewed for readability. As discussed offline, fine with a bit of unnecessary verbosity in this script

Couple optional comments, but pre-approving 🛫

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,204 @@
...
+  append_copy_block 'cooldown' "canvas_id IN ($unlocked_canvas_ids)"
+  append_copy_block 'frame' "canvas_id IN ($unlocked_canvas_ids)"
+
+  gzip -f "$backup_file"
+  echo "$(date -Is) Backup created: $backup_file.gz"
+
+  find "$BACKUP_DIR" -type f -name 'db-*.sql.gz' -mtime +"$RETENTION_DAYS" -delete || true
+}
+
+echo "Starting backup loop with UNLOCKED_INTERVAL_MINUTES=$UNLOCKED_INTERVAL_MINUTES, FULL_INTERVAL_MINUTES=$FULL_INTERVAL_MINUTES, RUN_SCHEMA_WITH_FULL=$RUN_SCHEMA_WITH_FULL, TZ=$TZ"
```
jaskfla, line 154: Also optional, put everything from here onwards in `main() {}` and run `main` at the bottom of the script

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,204 @@
...
+    else
+      run_full_backup
+    fi
+    echo "$now_epoch" > "$STATE_DIR/last_full_epoch"
+    echo "$now_epoch" > "$STATE_DIR/last_unlocked_epoch"
+  fi
+
+  last_unlocked_epoch=$(cat "$STATE_DIR/last_unlocked_epoch" 2>/dev/null || echo 0)
+  unlocked_interval_seconds=$((UNLOCKED_INTERVAL_MINUTES * 60))
+  if [ $((now_epoch - last_unlocked_epoch)) -ge "$unlocked_interval_seconds" ]; then
```
jaskfla: Another `(( ))` candidate

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,204 @@
...
+      wait "$schema_pid"
+    else
+      run_full_backup
+    fi
+    echo "$now_epoch" > "$STATE_DIR/last_full_epoch"
+    echo "$now_epoch" > "$STATE_DIR/last_unlocked_epoch"
+  fi
+
+  last_unlocked_epoch=$(cat "$STATE_DIR/last_unlocked_epoch" 2>/dev/null || echo 0)
+  unlocked_interval_seconds=$((UNLOCKED_INTERVAL_MINUTES * 60))
```
jaskfla, line 197: These will only ever be integers, right? Could use `declare -i last_unlocked_epic=...` to enforce that

scripts/backup/db-backup.sh
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env bash
+# Continuous PostgreSQL backup runner for containerized deployment.
+set -euo pipefail
```
jaskfla: Other comment is now outdated, but reflagging this as a (very common) antipattern

```suggestion
set -ex
```

scripts/backup/db-backup.sh
S-Ou: copilot seems to disagree that it's redundant, ie when the variable is unset it doesn't throw

I'll leave it as is since it doesn't hurt to be explicit

scripts/backup/db-backup.sh
S-Ou: Feels unnecessary, it's the user's fault if they want to start adding decimal minutes

scripts/backup/db-backup.sh
S-Ou: I'll pass but can consider if we ever need to significantly refactor this script

jaskfla: Yeah, it really only makes a meaningful difference if a script is expected to be run as it streams (e.g. `curl | bash`)

scripts/backup/db-backup.sh
jaskfla: Not actually decimals I had in mind, but typos

PR #436, opened 2026-04-22
refactor(frontend): replace audio context with `useLocalStorage`

### LINKED ISSUE #278: Audio cue on cooldown expiry

jaskfla:


Rocked03 (approved):
i've got no strong thoughts but just gave it a quick pass and useLocalStorage makes sense to me

this isn't merging into main anyways so we can do a more thorough review in the PR that does merge into main

packages/frontend/src/app/settings/useLocalStorage.ts
```diff
@@ -0,0 +1,88 @@
...
+    },
+    [key, queryClient],
+  );
+
+  const { mutate } = useMutation({
+    mutationKey: ["localStorage", key],
+    mutationFn: set,
+  });
+
+  return [data, mutate];
```
jaskfla, line 85: One could easily imagine returning `[get, set, remove]`; but for now there’s no business logic[^foot] that would `removeItem()`

[^foot]: Would it be _hobby logic_ for us in this context? 🤔

Rocked03: Project Blurple can be an unregistered non-profit business 😎

packages/frontend/src/app/settings/useLocalStorage.ts
jaskfla, line 1: - `localStorage` stores only strings
- Using `JSON.stringify` and `JSON.parse` to serialise/deserialise
- Hence only supporting JSON data types (no `undefined`, `Date`…)

packages/frontend/src/app/settings/Settings.tsx
```diff
@@ -46,14 +46,49 @@ const Aside = styled("aside")`
...
+      onChange={(e) => setValue(e.target.checked)}
+    />
+  );
+}
 
+function CooldownExpiryJingleSetting() {
+  const [value, setValue] = useLocalStorage("cooldown-jingle");
+  return (
+    <CheckboxSetting
+      aria-busy={value === undefined}
```
Rocked03, line 71: Are these all going to be `value === undefined`? Why not just move that into `CheckboxSetting` and set it to be `checked === undefined`?

packages/frontend/src/app/settings/useLocalStorage.ts
```diff
@@ -0,0 +1,86 @@
...
+  | number
+  | string
+  | JsonDataType[]
+  | { [key: string]: JsonDataType };
+
+type Satisfies<T extends BaseT, BaseT> = T;
+
+interface LocalStorageValueTypes extends Satisfies<
+  {
+    "cooldown-jingle": boolean;
```
Rocked03, line 23: perhaps we could prefix these to match the settings categories? `audio/cooldown-jingle`, `audio/sound-fx`, `accessibility/webfonts` etc?

packages/frontend/src/app/settings/useLocalStorage.ts
```diff
@@ -0,0 +1,86 @@
...
+  const queryClient = useQueryClient();
+  const set = useCallback(
+    async (newValue: LocalStorageValueTypes[Key]) => {
+      try {
+        window.localStorage.setItem(key, JSON.stringify(newValue));
+        await queryClient.invalidateQueries({
+          queryKey: ["localStorage", key],
+        });
+      } catch (e) {
+        console.error(e); // TODO: Reveal this error to user
```
Rocked03, line 74: I think I originally came across this by recommendation from somebody (probably Jasper), but I've used it a number of times and it's shockingly good at its job https://sonner.emilkowal.ski/

packages/frontend/src/app/settings/useLocalStorage.ts
Rocked03: I'm anticipating that this shouldn't be an issue, settings are typically a set of predefined choices and I can't think of anything that we might implement that would need anything more

packages/frontend/src/app/settings/Settings.tsx
jaskfla: Opinionated, but it’s not the responsibility of an `<input>` to _figure out_ if the data it represents is loading

It just so happens that `useLocalStorage`’s contract is such that `undefined` means loading; but one could imagine that if the state came from `useQuery` then we would pass it `isFetching`

Clarity over brevity

packages/frontend/src/app/settings/useLocalStorage.ts
jaskfla: Ahahaha, yes, that sounds like me. Fucking great library, that one

Not sure how heavy a dependency it is. If we go with toasts, for sure I would vote for Sonner; but honestly it could just pop a message inline on the settings page and save the dependency

Will leave out of scope for this PR, though

packages/frontend/src/app/settings/Settings.tsx
Rocked03: Shared state is all, but I'm not fussed either way

packages/frontend/src/app/settings/useLocalStorage.ts
jaskfla: I would go as far as saying that anything more complex probably _shouldn’t_ be using localStorage

packages/frontend/src/app/settings/useLocalStorage.ts
jaskfla: Good call; I’ll do this

PR #438, opened 2026-04-22
chore(frontend): more succinct cooldown button label

jaskfla:


packages/frontend/src/components/button/PlacePixelButton.tsx
```diff
@@ -103,7 +112,13 @@ export default function PlacePixelButton({ isVerbose }: PlacePixelButtonProps) {
   if (timeLeft > 0) {
     return (
       <Button variant="contained" disabled>
-        On cooldown: {timeLeft} seconds left
+        On cooldown (
+        <Time>
+          {durationFormat?.format({ seconds: timeLeft }) ?? (
```
Rocked03, line 117: will this cause any unexpected issues if the text width is too wide for the button?

Rocked03 (approved):

packages/frontend/src/components/button/PlacePixelButton.tsx
jaskfla: Exceedingly unlikely with `style: "narrow"`, which is the most compact format

(In fact, with Safari, `en-US` doesn’t even put a space before the “s” 🤮)

PR #440, opened 2026-04-22
fix(frontend): action panel tab semantics

jaskfla:


packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -98,91 +104,98 @@ export const Heading = styled("h2")`
...
   return (
     <Wrapper>
-      <TabBar>
+      <TabBar role="tablist">
         <Tab
-          tabKey={TABS.PLACE}
-          label="Place"
-          currentTab={currentTab}
+          aria-controls={placeTabId}
+          aria-selected={currentTab === "place"}
```
jaskfla, line 158: Some annoying duplication here but this can/will all change when we move state to the URL anyway

packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
```diff
@@ -31,14 +32,20 @@ export const Block = styled("div")`
...
-export const TabBlock = styled(Block, {
+const StyledBlock = styled(Block, {
   shouldForwardProp: (prop) => prop !== "active",
 })<{ active?: boolean }>`
   display: ${({ active }) => (active ? "grid" : "none")};
   gap: 0.5rem;
   grid-template-rows: auto 1fr auto;
 `;
 
+export function Tabpanel(
```
Rocked03:
```suggestion
export function TabPanel(
```
?

Rocked03 (approved):
my beloved attempt at enums 😭 


looks good tho

packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx
jaskfla: I can accept that—I’ll do the refactor

**Update:** 72eae0c435775cd265a9c2993c1b1e914c8b07d0 ✅

jaskfla:
> my beloved attempt at enums 😭

If you feel strongly about it, I’m happy to put it back

The main thing that prompted me to refactor was the fact that it was named `TABS` (plural)) and not `TabKey` (singular). I find that it doesn’t make sense to have a type like `(tabKey: TABS) => void`; it implies that `tabKey` is a collection

…but also pumbas and I do just have a distaste for TS enums

Rocked03:
> > my beloved attempt at enums 😭
> 
> If you feel strongly about it, I’m happy to put it back
> 
> The main thing that prompted me to refactor was the fact that it was named `TABS` (plural)) and not `TabKey` (singular). I find that it doesn’t make sense to have a type like `(tabKey: TABS) => void`; it implies that `tabKey` is a collection
> 
> …but also pumbas and I do just have a distaste for TS enums

Haha it's alg

PR #459, opened 2026-04-25
fix(deps): quarantine period

jaskfla:
pnpm uses minutes(!)


Rocked03 (approved):
wdym, 7 minutes is a totally reasonable quarantine period

PR #460, opened 2026-04-25
chore(deps): transitive dependency vulnerability

jaskfla:


packages/frontend/package.json
```diff
@@ -49,7 +49,7 @@
     "lucide-react": "^1.8.0",
     "luxon": "^3.7.2",
     "next": "^16.2.3",
-    "react": "^19.2.4",
+    "react": "^19.2.5",
```
jaskfla, line 52: Bumped because of react-dom peer dep requirement. Not actually sure how long this issue has been around—maybe 2–3 weeks, maybe somehow(?) new to this PR

Rocked03 (approved):

jaskfla:
Looks like I also accidentally committed Knip as dev dependency in #461

Piggybacked reversion onto this PR. My bad

PR #465, opened 2026-04-25
fix: typescript errors in types package

pumbas600:
We weren't correctly type checking our types package, so there were a number of errors we missed.

### DIFF
--- a/packages/backend/src/services/paletteService.ts
+++ b/packages/backend/src/services/paletteService.ts
@@ -1,7 +1,13 @@
-import { PaletteColor, PixelColor } from "@blurple-canvas-web/types";
+import {
+  PaletteColor,
+  PaletteColorSummary,
+  PixelColor,
+} from "@blurple-canvas-web/types";
 
-import { prisma } from "@/client";
+import { color, prisma } from "@/client";
 import { getCurrentEvent } from "./eventService";
+
+type ColorSummary = Pick<color, "id" | "code" | "name" | "rgba" | "global">;
 
 /**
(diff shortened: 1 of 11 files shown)


packages/types/tsconfig.json
```diff
@@ -4,6 +4,7 @@
 
   // Reference: https://www.typescriptlang.org/tsconfig
   "compilerOptions": {
+    "skipLibCheck": false,
```
jaskfla, line 7: Does this default true? o.O

packages/backend/src/services/paletteService.ts
```diff
@@ -69,3 +75,15 @@ export async function getEventPalette(
...
 }
+
+export function toPaletteColorSummary(
+  color: ColorSummary,
+): PaletteColorSummary {
+  return {
+    id: color.id,
+    code: color.code,
+    name: color.name,
+    rgba: color.rgba as PixelColor,
```
jaskfla, line 86: Ah, yeah, I was wondering why this function exited. I guess this is a reasonable compromise given Prisma’s generated type here

jaskfla (approved):
Thanks 😅

packages/types/tsconfig.json
pumbas600: It must do, because when I turned it off I started getting errors about these 😔

packages/backend/src/services/paletteService.ts
pumbas600: Yup, but I also like that it ensures we're returning explicitly what we're expecting, rather than anything else that may exist on `color`

PR #467, opened 2026-04-25
feat: improve docker builds

pumbas600:
Closes #265 

**What's Changed?**
- Optimises Docker builds (They're only a fraction of their old size).
- Removes final references to Prismock

image

PR #469, opened 2026-04-25
chore(deps): fix dev dependencies declared as regular dependencies

jaskfla:


Rocked03 (approved):
duped in #467, should we close this one? approved anyways if we'd rather make the change separately

jaskfla:
> duped in #467, should we close this one? approved anyways if we'd rather make the change separately

Only partial dupe; and no, these changes will disappear from the diff when this is merged and #467 is rebased onto main.

PR #454, opened 2026-04-24
feat(frontend): admin dashboard - read-only tabs

Rocked03:
- View event info tab
- View color info tab

image

image

- still to come
  - create and edit canvases
  - create and edit notices
  - paste pixels

---
_old plan:_
- ~~Create and edit events~~
- ~~Create and edit canvases~~
- ~~Create and edit colors~~
  - ~~Assign colors to partner servers~~

~~**Does not** include fancy styling (that's saved for a future story)~~

PR #471, opened 2026-04-25
feat(backend): admin endpoints

### LINKED ISSUE #472: Admin endpoints
All the endpoints needed for #385

Rocked03:
Breaking things up by moving the backend changes in #454 over to this standalone PR

- Create and edit events
- Create and edit canvases
- Create and edit colors
  - Assign colors to partner servers


packages/backend/src/services/canvasService.ts
```diff
@@ -348,3 +349,68 @@ async function getOrFetchCacheCanvas(canvasId: number): Promise<CachedCanvas> {
...
+      name,
+      width,
+      height,
+      event_id: currentEventId.id,
+      start_coordinates: startCoordinates,
+      locked: true,
+      all_colors_global: allColorsGlobal,
+      cooldown_length: cooldownLength,
+    },
+  });
```
Rocked03, line 381: Oh I'm an idiot - this needs to also create all the initial rows in the pixel table

packages/backend/examples.rest
jaskfla, line 1: Oh damn I completely forgot we had this—thanks for keeping it alive

packages/backend/src/models/guild.models.ts
jaskfla, line 1: Miscellaneous or guild models? 🧐

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -66,6 +75,60 @@ canvasRouter.get("/:canvasId", async (req, res) => {
...
 
+canvasRouter.post("/", async (req, res) => {
+  try {
+    assertLoggedIn(req);
+    assertCanvasAdmin(req.user);
+
+    const canvasData = await CreateCanvasBodyModel.safeParseAsync(req.body);
+
+    if (!canvasData.success) {
+      res.status(400).json({ message: "Invalid canvas data" });
```
jaskfla: Are we able to forward a parse error from Zod?

From memory should be in the parse result as `canvasData.errors` or something like that (or use try–catch with “unsafe” parse and use `error.message` from the caught error

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -66,6 +75,60 @@ canvasRouter.get("/:canvasId", async (req, res) => {
...
+
+    await createCanvas({
+      name: canvasData.data.name,
+      width: canvasData.data.width,
+      height: canvasData.data.height,
+      startCoordinates: canvasData.data.startCoordinates,
+      cooldownLength: canvasData.data.cooldownLength,
+    });
+
+    res.status(201).json({ message: "Canvas created successfully" });
```
jaskfla: Respond with the created record[^success]

[^success]: And another reminder that “successfully” is almost never needed in a success message

packages/backend/src/routes/api/v1/event.ts
```diff
@@ -17,19 +28,47 @@ eventRouter.get("/current", async (_req, res) => {
...
+
+    await createEvent(eventData.name, eventData.id);
 
-    if (!pathParams.success) {
-      throw new BadRequestError(
-        "Malformed path parameters",
-        pathParams.error.issues,
-      );
+    res.status(201).json({ message: "Event created successfully" });
+  } catch (error) {
```
jaskfla, line 52: Too many things that could throw for my liking

I understand that we have a bit of a pattern of wrapping the entire handler in a try because of `ApiError.sendError`. Considering that, wrap whatever throws `NotFoundError` in a nested try block

packages/backend/src/routes/api/v1/event.ts
```diff
@@ -17,19 +28,47 @@ eventRouter.get("/current", async (_req, res) => {
...
+    if (error instanceof NotFoundError) {
+      return res.status(409).json({ message: error.message });
```
jaskfla: Very minor, but I think this is clearer about what’s actually happening:

```suggestion
      res.status(409).json({ message: error.message });
      return;
```

jaskfla (commented):
Only viewed 16/24 files so far, but submitting partial review here

No major complaints

packages/backend/src/models/guild.models.ts
Rocked03: hm it turned out that way didn't it

packages/backend/examples.rest
```diff
@@ -94,11 +94,34 @@ GET {{basePath}}/canvas/{{canvasId}}/pixel/history
...
 
+### Query pixel history (POST)
+# Requires moderator session
+POST {{basePath}}/canvas/3/pixel/history
+  ?x0=1
+  &y0=1
+  &x1=2
+  &y1=2 HTTP/1.1
+content-type: application/json
+{
```
jaskfla, line 106: Does this work? I could be misremembering but I thought it needed a blank line before the request body

packages/backend/src/models/color.models.ts
```diff
@@ -14,13 +14,7 @@ export async function parseColorId(
...
 ): Promise<PaletteColor["id"]> {
   const result = await ColorIdParamModel.safeParseAsync(params);
-  if (!result.success) {
-    throw new BadRequestError(
-      `${params.colorId} is not a valid color ID`,
-      result.error.issues,
-    );
-  }
-
+  assertZodSuccess(result, `${params.colorId} is not a valid color ID`);
```
jaskfla, line 17: It looks like `assertZodSuccess` destroys the type and makes `result` hold `unknown` as far as LSP is concerned?

Should be pretty straightforward to make it generic and preserve the expected type

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -95,7 +92,7 @@ canvasRouter.post("/", async (req, res) => {
       cooldownLength: canvasData.data.cooldownLength,
     });
 
-    res.status(201).json({ message: "Canvas created successfully" });
+    res.status(201).json({ message: "Canvas created" });
```
jaskfla: Return canvas record here

packages/backend/src/routes/api/v1/event.ts
```diff
@@ -42,16 +43,13 @@ eventRouter.post("/", async (req, res) => {
...
 
-    const eventData = await CreateEventBodyModel.parseAsync(req.body);
+    const eventData = await CreateEventBodyModel.safeParseAsync(req.body);
+    assertZodSuccess(eventData);
 
-    await createEvent(eventData.name, eventData.id);
+    await createEvent(eventData.data.name, eventData.data.id);
 
-    res.status(201).json({ message: "Event created successfully" });
+    res.status(201).json({ message: "Event created" });
```
jaskfla: Return created record

packages/backend/src/routes/api/v1/event.ts
```diff
@@ -63,12 +61,13 @@ eventRouter.put("/:eventId", async (req, res) => {
...
-      EditEventBodyModel.parseAsync(req.body),
+      EditEventBodyModel.safeParseAsync(req.body),
     ]);
+    assertZodSuccess(eventData);
 
-    await editEvent(eventId, eventData.name);
+    await editEvent(eventId, eventData.data.name);
 
-    res.status(200).json({ message: "Event edited successfully" });
+    res.status(200).json({ message: "Event edited" });
```
jaskfla: In the case that PUT edits an existing record: either 200 with edited record as response body, or 204 if we feel that’s redundant

In this case this creates a new record, it “must” respond with 201

packages/backend/src/services/eventService.ts
```diff
@@ -45,7 +46,7 @@ export async function createEvent(
 ): Promise<BlurpleEvent> {
   const existingEvent = await getEventById(id).catch(() => null);
   if (existingEvent) {
-    throw new NotFoundError(`An event with ID ${id} already exists`);
+    throw new ConflictError(`An event with ID ${id} already exists`);
```
jaskfla: Nice spot, but why `catch` and _then_ check whether it’s defined?

Put in a `try–catch`, and in the case of a Prisma unique constraint violation throw `ConflictError`

jaskfla (changes requested):
No qualms with the endpoint logic itself. Really just comes down to two comments, both of which are small but touch many files:

- HTTP semantics could do with a review; and
- devX would improve hugely by making `assertZodSuccess` generic

jaskfla (approved):
🚀

PR #474, opened 2026-04-25
feat(backend): moderation endpoints

### LINKED ISSUE #473: Moderation endpoints
All the endpoints required for #371

Rocked03:
Endpoints for
- Complex history search
- Blacklist


packages/backend/prisma/schema.prisma
jaskfla, line 1: Makes sense to leave the database be, but for the new code in JS-land, I’d push to use the term **blocklist** instead

packages/backend/src/models/blocklistModels.ts
```diff
@@ -0,0 +1,31 @@
...
+    );
+  }
+  const { userId } = result.data;
+  if (!userId) {
+    throw new BadRequestError("userId is required");
+  }
+  if (Array.isArray(userId)) {
+    return userId.map((id) => BigInt(id));
+  }
+  return [BigInt(userId)];
```
jaskfla: Small stylistic thing, but normalising the input into an array early on is a bit easier to replicate as a pattern

```suggestion
  const ids = Array.isArray(userId)) ? userId : [userId]
  return userId.map(BigInt);
```

Less critical in this case because the logic is so simple, but if it were more than just passing it through `BitInt` there would be much more duplicated code

packages/backend/src/models/paramModels.ts
jaskfla, line 1: Optional, but I’d personally prefer either of these naming patterns

- `…/models/models.ts`, or
- `…/models/history.models.ts`.

If you could please also rename `paramModels` to reflect this the new vertical splitting 🙏

packages/backend/src/routes/api/v1/blocklist.test.ts
```diff
@@ -0,0 +1,133 @@
...
+}));
+
+import { blacklistRouter } from "./blacklist";
+
+const createApp = ({ authenticated = false, moderator = false } = {}) => {
+  const app = express();
+  app.use(express.json());
+  app.use(mockAuth);
+  app.use((req, _res, next) => {
+    req.session = {} as typeof req.session;
```
jaskfla, line 23: Think I flagged this somewhere else (different PR?), but I think this should be typeable as `RequestHandler`

packages/backend/src/routes/api/v1/blocklist.test.ts
```diff
@@ -0,0 +1,133 @@
...
+describe("Blacklist route tests", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("returns the blacklist for a moderator", async () => {
+    const blacklist = [
+      {
+        user_id: 9n,
+        date_added: new Date(0),
```
jaskfla: - Ideally make lines 48 and 62 point to the same `Date` instance;
- can also use use `new Date()` (non-epoch)

packages/backend/src/routes/api/v1/blocklist.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      .set("X-TestUserId", "1")
+      .send({
+        userId: ["1", "2"],
+      })
+      .type("json")
+      .expect(200);
+
+    expect(response.body).toStrictEqual({
+      message: "Users added to blacklist",
+    });
```
jaskfla: This should be a `201 Created` with the created record as response

packages/backend/src/services/blacklistService.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function checkIfUserIsBlacklisted(
+  userId: BlacklistEntry["userId"],
+): Promise<boolean> {
+  const blacklistEntry = await prisma.blacklist.findFirst({
+    where: {
+      user_id: userId,
+    },
+  });
```
jaskfla: Does Prima have `exists()` method?

packages/backend/src/services/blacklistService.ts
```diff
@@ -0,0 +1,58 @@
...
+        user_id: userId,
+      })),
+      skipDuplicates: ignoreDuplicates,
+    });
+  } catch (error) {
+    if (
+      error instanceof Prisma.PrismaClientKnownRequestError &&
+      error.code === PrismaErrorCode.UniqueConstraintViolation
+    ) {
+      throw new BadRequestError("User is already in the blacklist");
```
jaskfla: If POST request, respond with `409 Conflict` instead

Even better, use PUT request and just consider this a success. Absolutely no harm in  making this idempotent and always using `skipDuplicates`, right?

packages/backend/src/services/blacklistService.ts
```diff
@@ -0,0 +1,58 @@
...
+export async function getBlacklist() {
+  return await prisma.blacklist.findMany({
+    select: {
+      user_id: true,
+      date_added: true,
+    },
+  });
+}
+
+export async function checkIfUserIsBlacklisted(
```
jaskfla: - `userIsBlocklisted` or `isUserBlocklisted`; or
- `getUserIsBlocklisted` or `getIsUserBlocklisted` to found more function-y

`checkIf` doesn’t imply what the return type is; sounds almost as if it does the check and doesn’t report anything back

Consider how these two lines read:

```ts
if (userIsBlocklisted()) { /* do a thing */ };
if (checkIfUserIsBlocklisted()) { /* do a thing */ };
```

packages/backend/src/services/discordGuildService.ts
```diff
@@ -198,7 +198,7 @@ export function assertCanvasAdmin(
 
 export function assertCanvasModerator(
```
jaskfla:
```suggestion
export function assertIsCanvasModerator(
```

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+import { toPaletteColorSummary } from "./paletteService";
+import { validatePixel } from "./pixelService";
+
+interface GetPixelHistoryParams {
+  canvasId: number;
+  coordinates: Point | [Point, Point];
+  dateRange?: {
+    from?: Date;
+    to?: Date;
+  };
```
jaskfla, line 19: Throwing out that this could arguably be typed as a pair, but I think my preference is what you have here

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
+import type { PixelHistoryWrapper, Point } from "@blurple-canvas-web/types";
+import { prisma } from "@/client";
+import { addUsersToBlacklist } from "./blacklistService";
+import { toPaletteColorSummary } from "./paletteService";
+import { validatePixel } from "./pixelService";
+
+interface GetPixelHistoryParams {
+  canvasId: number;
```
jaskfla: Derive this from the canvas type (same with `userId` below)

```suggestion
  canvasId: Canvas["id"];
```

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+}: GetPixelHistoryParams): Promise<PixelHistoryWrapper> {
+  if (!Array.isArray(coordinates)) {
+    await validatePixel(canvasId, coordinates, false);
+    coordinates = [coordinates, coordinates];
+  } else {
+    await Promise.all([
+      validatePixel(canvasId, coordinates[0], false),
+      validatePixel(canvasId, coordinates[1], false),
+    ]);
+  }
```
jaskfla: Same comment about normalising `coordinates` into an array _first_, then do the `Promise.all(coordinates.map())` logic with no duplication

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+    timestamp: {
+      gte: dateRange?.from,
+      lte: dateRange?.to,
+    },
+    user_id:
+      userIdFilter ?
+        userIdFilter.include ?
+          { in: userIdFilter.ids }
+        : { notIn: userIdFilter.ids }
+      : undefined,
```
jaskfla: Bit tricky to parse. Consider building this with a function/IIFE within function scope?

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  const [pixelHistory, totalEntries] = await Promise.all([
+    prisma.history.findMany({
+      take: 100,
+      orderBy: {
+        timestamp: "desc",
+      },
+      where: whereFilter,
```
jaskfla: If you name it `where` you can simplify this and line 81

```suggestion
      where,
```

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+          }
+        : null,
+    })),
+    totalEntries,
+  };
+}
+
+export async function deletePixelHistoryEntries(
+  canvasId: number,
+  historyIds: bigint[],
```
jaskfla: Derivable

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+    },
+    select: {
+      id: true,
+      user_id: true,
+    },
+  });
+
+  const existingEntryIds = new Set(existingEntries.map((entry) => entry.id));
+
+  const invalidIds = historyIds.filter((id) => !existingEntryIds.has(id));
```
jaskfla:
```suggestion
  const invalids = new Set(historyIds).difference(existingEntryIds);
```

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (invalidIds.length > 0) {
+    throw new Error(
+      `The following history IDs do not exist for canvas ${canvasId}: ${invalidIds
+        .map((id) => id.toString())
+        .join(", ")}`,
```
jaskfla: Nice verbose error message without going overboard 👌

Suggestions:

- Pretty sure `toString` is redundant;
- Use `Intl.ListFormat` instead of `.join()`

```ts
const formatter = new Intl.ListFormat('en-US'); // Defaults to { type: 'conjunction' }
throw new Error(`…${formatter.format(invalids)}`)
```

packages/backend/src/services/blacklistService.ts
```diff
@@ -0,0 +1,58 @@
...
+  const blacklistEntry = await prisma.blacklist.findFirst({
+    where: {
+      user_id: userId,
+    },
+  });
+  return !!blacklistEntry;
+}
+
+export async function addUsersToBlacklist(
+  userIds: BlacklistEntry["userId"][],
```
jaskfla: Consider

```suggestion
  userIds: Iterable< BlacklistEntry["userId"]>,
```

At (least) one of the call sites, you’re turning a `Set` into an `Array` just to pass it into here. I see no harm in accepting both, and normalising in here if needed

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,150 @@
...
+        : null,
+    })),
+    totalEntries,
+  };
+}
+
+export async function deletePixelHistoryEntries(
+  canvasId: number,
+  historyIds: bigint[],
+  blacklistAuthors: boolean = false,
```
jaskfla: Tiiiiny nitpick, but because `blocklist` can be both verb and noun, it sounds like this could be an array of user IDs

I’d suggest `shouldBlockAuthors` (admittedly a mouthful), or `doBlockAuthors` if you want shorter (but this is unclear imo)

packages/backend/src/services/pixelService.ts
```diff
@@ -143,13 +82,7 @@ export async function validateColor(
...
 export async function validateUser(userId: bigint) {
-  const blacklist = await prisma.blacklist.findFirst({
-    where: {
-      user_id: userId,
-    },
-  });
-
-  if (blacklist) {
+  if (await checkIfUserIsBlacklisted(userId)) {
     throw new ForbiddenError("User is blacklisted");
```
jaskfla: Does the requesting client see this error message? Wonder if this message should be more opaque/generic

packages/backend/src/utils/index.ts
```diff
@@ -6,6 +9,7 @@ BigInt.prototype.toJSON = function (): string {
 
 export const PrismaErrorCode = {
   UniqueConstraintViolation: "P2002",
+  RecordNotFound: "P2025",
 } as const;
```
jaskfla, line 13: Ooh, I missed that you had to manually define these. Surprised that you couldn’t find an export from Prisma

In that absence, this was a great call. Much more readable 🙏

packages/backend/src/utils/index.ts
```diff
@@ -23,3 +27,17 @@ export function normalizeBounds({ x0, y0, x1, y1 }: Bounds): Bounds {
     y1: Math.max(y0, y1),
   };
 }
+interface AuthenticatedRequest extends Request {
```
jaskfla, line 31:
```suggestion

interface AuthenticatedRequest extends Request {
```

jaskfla (changes requested):
Very happy with module structure and how responsibilities are divvied up

One (more major) request re terminology, and my usual flurry of readability nitpicks

Nice stuff 🙏

packages/backend/src/models/history.models.ts
```diff
@@ -0,0 +1,56 @@
...
+          code: "custom",
+          message: "startDateTime must be before endDateTime",
+        });
+      }
+
+      if (includeUserIds && excludeUserIds) {
+        ctx.addIssue({
+          code: "custom",
+          message: "Cannot have both includeUserIds and excludeUserIds",
+        });
```
jaskfla, line 60: Why not? I understand that there can be conflict if both include the same user ID, but is it an invalid use case to provide two non-overlapping sets here?

If that _is_ a genuine usage, then we can resolve either by validating that they’re non-overlapping or giving one higher precedence

packages/backend/prisma/schema.prisma
Rocked03: I was thinking that the whole time I wrote this, but I'd rather that all gets updated in one go in a separate PR

packages/backend/src/models/paramModels.ts
Rocked03: Could do. I'm leaving the original body and param models in this file since I already refactored them in the admin endpoints pr

packages/backend/src/routes/api/v1/history.ts
```diff
@@ -0,0 +1,132 @@
...
+    const pixelHistory = await getPixelHistory({ canvasId, coordinates });
+
+    res.status(200).json(pixelHistory);
+  } catch (error) {
+    ApiError.sendError(res, error);
+  }
+});
+
+historyRouter.post<CanvasIdParam>("/", async (req, res) => {
+  // Could become a QUERY endpoint in the future once it becomes supported
```
jaskfla, line 44: Are you implying that this has `GET` semantics?

packages/backend/src/routes/api/v1/history.ts
```diff
@@ -0,0 +1,132 @@
...
+  deletePixelHistoryEntries,
+  getPixelHistory,
+} from "@/services/historyService";
+import { assertLoggedIn } from "@/utils";
+
+export const historyRouter = Router({ mergeParams: true });
+
+historyRouter.get<CanvasIdParam>("/", async (req, res) => {
+  try {
+    // grabbing the canvasId from the path
```
jaskfla: `parseCanvasId` makes this clear enough I reckon

```suggestion
```

packages/backend/src/routes/api/v1/history.ts
```diff
@@ -0,0 +1,132 @@
...
+    const canvasId = await parseCanvasId(req.params);
+
+    // grabbing the x and y from the query
```
jaskfla:
```suggestion
```

packages/backend/src/routes/api/v1/history.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const coordinate0 = {
+      x: queryResult.data.x0,
+      y: queryResult.data.y0,
+    };
+    const coordinate1 = {
+      x: queryResult.data.x1 ?? queryResult.data.x0,
+      y: queryResult.data.y1 ?? queryResult.data.y0,
+    };
+    const coordinates: [Point, Point] = [coordinate0, coordinate1];
```
jaskfla: Let’s decide between `coordinate` and `point` if they mean the same thing[^ring]

[^ring]: One term to rule them all 💍

packages/backend/src/models/history.models.ts
Rocked03: The IDs will be manually provided by the searcher, so why would you need both?

For a set `a, b, c, d, e, f`:
- you can include only `a, b, c`
- you could exclude `e, f`
- but what good does it do if you include `a, b, c` but also exclude `e, f` which aren't even part of the subset you've just requested
- alternatively there's not much point both including and excluding the same value

packages/backend/prisma/schema.prisma
jaskfla: Hm, yeah, that’s okay too; but right now there are less than 20 instances. Wouldn’t pollute this diff too much

Your call either way—feel free to resolve this thread

packages/backend/src/models/history.models.ts
jaskfla: Ahh, gotcha, that makes sense

Rocked03: Note to self: duplicate this, but for color IDs (forgot to add it)

packages/backend/prisma/schema.prisma
Rocked03: Fair point, I can definitely make the change in the front and backend, but I think I'll leave the actual schema as is for now, and then add a note to change this in the story about merging the blocklist into the user table

packages/backend/src/models/history.models.ts
jaskfla: Missed this earlier, but I think naming them `from, to, include, exclude` is cleaner

Rocked03: It'll be exclude/includeColors too though, will still need the distinguishment

packages/backend/src/routes/api/v1/history.ts
S-Ou: I mean, yeah? It's a request purely to get data, except that it uses a request body to pass in a more complex search query

packages/backend/src/services/historyService.ts
S-Ou: Considered that, but then we run validatePixel twice on the same coord and that makes little sense to me

packages/backend/src/services/historyService.ts
S-Ou: hm no... our types store bigints as strings

packages/backend/src/services/blacklistService.ts
S-Ou: not that I can find

packages/backend/src/services/blacklistService.ts
S-Ou: ehh I suppose

packages/backend/src/services/pixelService.ts
S-Ou: Eh I don't feel like we need to be opaque about a user being blocklisted

.vscode/settings.json
```diff
@@ -23,6 +23,7 @@
     }
   ],
   "cSpell.words": [
+    "Blocklisted",
```
jaskfla:
```suggestion
    "blocklisted",
```

packages/backend/src/routes/api/v1/blocklist.test.ts
```diff
@@ -0,0 +1,154 @@
...
+} from "@/services/blocklistService";
+import { mockAuth } from "@/test/mockAuth";
+
+vi.mock("@/services/blocklistService", () => ({
+  addUsersToBlocklist: vi.fn(),
+  getBlocklist: vi.fn(),
+  removeUsersFromBlocklist: vi.fn(),
+}));
+
+import { blocklistRouter } from "./blocklist";
```
jaskfla: I take it order is important here? May be worth a comment

packages/backend/src/routes/api/v1/blocklist.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const app = createApp({ authenticated: true, moderator: true });
+    vi.mocked(removeUsersFromBlocklist).mockResolvedValueOnce(undefined);
+
+    const response = await request(app)
+      .delete("/api/v1/blocklist")
+      .set("X-TestUserId", "1")
+      .send({
+        userId: "9",
+      })
+      .type("json")
```
jaskfla: Is content-type applicable to a 204?

packages/backend/src/services/blocklistService.test.ts
jaskfla, line 1: Really appreciate these colocated `.test.ts` files 🙏

packages/backend/src/services/blocklistService.ts
jaskfla, line 1: Every header comment in this file is redundant. Code is perfectly self-documenting imo

packages/backend/src/services/historyService.test.ts
```diff
@@ -0,0 +1,145 @@
...
+          global: true,
+        },
+        userId: "1",
+        userProfile: {
+          id: "1",
+          username: "test_user_1",
+          profilePictureUrl: "https://example.com/avatar1.png",
+        },
+      });
+    });
```
jaskfla, line 49: This looks veeeeeery tightly coupled to the internals of `seedAll()`

Def appreciate the testing efforts in this PR, but this sort of thing spooks me as being unmaintainable. Perhaps not a great candidate for unit testing, but a more end-to-end-y test?

```ts
// Arrange
const pixel1 = await placePixel();
const pixel2 = await placePixel();
const pixel3 = await placePixel();
const pixel4 = await placePixel();
// assert 201 status codes if we deem fit

// Act
const history = await getPixelHistory()

// Assert
expect(history.totalEntries).toBe(4); // …and so on
```

Very much not for this PR, but something for us to consider. For this PR, I would push to remove this test (and any similar tests)

packages/backend/src/services/historyService.ts
```diff
@@ -15,50 +19,72 @@ interface GetPixelHistoryParams {
...
+    user_id: (() => {
+      if (!userIdFilter) {
+        return undefined;
+      }
+      if (userIdFilter.include) {
+        return { in: userIdFilter.ids };
+      } else {
+        return { notIn: userIdFilter.ids };
+      }
+    })(),
```
jaskfla, line 74: Optional—just a preference thing—I find this much easier/quicker to parse:

```suggestion
    user_id: (() => {
      if (!userIdFilter) return undefined;
      const op = userIdFilter.include ? "in" : "notIn";
      return { [op]: userIdFilter.ids };
    })(),
```

…annoying downside is that TypeScript infers a wider type `{ [x: string]: bigint[] }` instead of `{ in: bigint[]; notIn?: undefined; } | { notIn: bigint[]; in?: undefined; }`, which if we want to re-narrow requires explicitly defining a really ugly union

packages/backend/src/services/historyService.ts
```diff
@@ -123,12 +156,11 @@ export async function deletePixelHistoryEntries(
 
   const existingEntryIds = new Set(existingEntries.map((entry) => entry.id));
 
-  const invalidIds = historyIds.filter((id) => !existingEntryIds.has(id));
-  if (invalidIds.length > 0) {
+  const invalidIds = new Set(historyIds).difference(existingEntryIds);
+  if (invalidIds.size > 0) {
+    const formatter = new Intl.ListFormat("en-US");
```
jaskfla: Shameless micro-optimisation, but I’d suggest instantiating this module level

Not actually fussed about this detail. It’s genuinely negligible (much less than 1 ms) on my laptop

packages/backend/src/services/historyService.ts
```diff
@@ -123,12 +156,11 @@ export async function deletePixelHistoryEntries(
...
     throw new Error(
-      `The following history IDs do not exist for canvas ${canvasId}: ${invalidIds
-        .map((id) => id.toString())
-        .join(", ")}`,
+      `The following history IDs do not exist for canvas ${canvasId}: ${formatter.format([...invalidIds].map((id) => id.toString()))}`,
```
jaskfla, line 157: Surely `.format()` convert to string for us? I’m not sure off the dome, but I would be surprised if not

(`Array.from` versus spread just a preference thing—your call)

```suggestion
      `The following history IDs do not exist for canvas ${canvasId}: ${formatter.format(Array.from(invalidIds))}`,
```

packages/backend/src/services/historyService.ts
```diff
@@ -0,0 +1,180 @@
...
+        color: true,
+        timestamp: true,
+        guild_id: true,
+        user_id: true,
+        discord_user_profile: true,
+      },
+    }),
+    prisma.history.count({
+      where: where,
+    }),
```
jaskfla:
```suggestion
    prisma.history.count({ where }),
```

jaskfla (approved):
Looking solid! One major-ish concern with coupling between seed data and unit test suites; and a handful of nitpicks details.

Nothing that I would consider blocking merge, so pre-approving for you to resolve and merge in your own time

packages/backend/src/services/blocklistService.ts
S-Ou: ¯\_(ツ)_/¯

packages/backend/src/services/historyService.ts
S-Ou: nope, it doesn't take bigint

PR #478, opened 2026-04-26
chore(lint): use `import type`

jaskfla:
Will help Knip diagnose dependency issues

Use `pnpm run lint:fix` or `pnpm lint-staged` when developing locally so you don’t have to worry about manually doing this busywork


Rocked03 (approved):
👍

PR #480, opened 2026-04-26
fix(frontend): `<ColorCodeChip>` semantics

jaskfla:


packages/frontend/src/components/button/PrimitiveButton.tsx
jaskfla, line 1: Struggling with what to name this:

- `PrimitiveButton` (eh, sure, as much as an HTML element can be “primitive”)
- `UnstyledButton` (not strictly true)
- `AppearanceNoneButton` (most technically correct, but kinda gross)

packages/frontend/src/components/button/index.ts
```diff
@@ -1,3 +1,4 @@
 export { Button } from "./Button";
 export { DynamicAnchorButton, default as DynamicButton } from "./DynamicButton";
 export { default as PlacePixelButton } from "./PlacePixelButton";
```
jaskfla, line 3: Out of scope, but this should be colocated with the view that uses it. It’s not really intended to be widely reused

packages/frontend/src/components/ColorCodeChip.tsx
```diff
@@ -29,33 +31,29 @@ const Container = styled("code", {
...
-      {...props}
-    >
-      {colorCode}
-    </Container>
+    <StyledButton onClick={clickHandler} onKeyUp={keyUpHandler} {...props}>
+      <code aria-hidden>{colorCode}</code>
+      <VisuallyHidden>
+        Code {colorCode.split("").join("-")}. Click to copy.
+      </VisuallyHidden>
+    </StyledButton>
```
jaskfla, line 57: Screenshot 2026-04-26 at 16 56 22

packages/frontend/src/components/button/PrimitiveButton.tsx
Rocked03: I don't really have any better suggestion, PrimitiveButton sounds fine

packages/frontend/src/components/ColorCodeChip.tsx
Rocked03: can't say no to a11y

Rocked03 (approved):
no strong feelings here, looks good

PR #482, opened 2026-04-28
feat(backend): limit frame count

### LINKED ISSUE #429: Limit frame count
Limit of 32 frames per owner per canvas

Number TBD, could also have different limits for users vs guilds

Enforce within the endpoint

Also include alert in the GUI when limit is reached

Rocked03:
Limit the count of frames a user or guild can own per canvas. Enforced in the endpoint but limit states are also propagated to the frontend to disable creations. Only restricts creating frames, will not prevent users from editing or deleting frames even if they are over the limit (for whatever reason)

Closes #429


Rocked03:
Yet to test the front-end, will un-draft once I've done that

packages/backend/src/config/index.ts
```diff
@@ -46,6 +46,10 @@ const config = {
   // Keep bot placing enabled by default unless explicitly disabled.
   botPlacingEnabled: process.env.BOT_PLACING_ENABLED !== "false",
   allColorsGlobal: process.env.ALL_COLORS_GLOBAL === "true",
+  frames: {
+    userLimit: Number(process.env.FRAME_LIMIT_USER || 32),
+    guildLimit: Number(process.env.FRAME_LIMIT_GUILD || 32),
```
jaskfla: `Number.parseInt`

packages/backend/src/routes/api/v1/frame.test.ts
```diff
@@ -145,6 +154,36 @@ describe("Frame mutation route tests", () => {
...
+        ownerId: "1",
+        isGuildOwned: false,
+        x0: 0,
+        y0: 0,
+        x1: 10,
+        y1: 10,
+      },
+    }).set("X-TestUserId", "1");
+
+    expect(response.status).toBe(400);
```
jaskfla: I think this is a `422 Unprocessable Content` situation

The request itself is syntactically perfectly valid

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -50,11 +55,22 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
       queryResult.data.guildIds,
       await parseCanvasId(req.params),
     );
-    res.status(200).json(frame);
+
+    const countLimitResults: Record<string, boolean> = {};
+    for (const guildId of queryResult.data.guildIds) {
+      countLimitResults[guildId] =
+        frames.filter((frame) => frame.owner.guild.guild_id === guildId)
+          .length >= config.frames.guildLimit;
```
jaskfla: Micro-optimisation, but this avoids instantiating a filtered array

```ts
frames.reduce((count, frame) => {
  if (frame.owner.guild.guild_id === guildId) count++;
  return count;
}, 0);
```

Would also be more readable to do the do the counting and comparing in two separate lines

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -50,11 +55,22 @@ frameRouter.get("/guilds/:canvasId", async (req, res) => {
...
+    }
+
+    res.status(200).json({
+      data: frames,
+      isAtCountLimit: countLimitResults,
```
jaskfla: Why two different names? 🧐

I’m a bit short on time right now so haven’t thought about suggestions, but I’m sure we can come up with something that better implies it’s a `Record<string, boolean>`. Both these names are a bit ambiguous

packages/backend/src/services/frameService.ts
```diff
@@ -401,3 +402,39 @@ export async function createFrame(
     }
   }
 }
+
+async function getFrameCountForOwner(
+  canvasId: number,
+  ownerId: string,
+  isGuildOwned: boolean,
```
jaskfla: No obvious/natural arg order here imo; I’d suggest using named params

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -214,15 +214,24 @@ export default function FrameEditPanel({
...
     .filter(([, guild]) => guild.administrator || guild.manageGuild)
     .toSorted(([, a], [, b]) => (b.memberCount ?? 0) - (a.memberCount ?? 0));
 
-  const { data: guildFrames = [] } = useGuildFrames({
+  const { data: guildFramesResponse } = useGuildFrames({
     canvasId: canvas.id,
     guildIds: managedGuildEntries.map(([guildId]) => guildId),
   });
 
+  const guildFrames = guildFramesResponse?.data ?? [];
```
jaskfla, line 232: Does this response contain anything meaningful _other_ than a `data` property?

If not, just make `useGuildFrames`’ query function return `.data`—we already do this in a couple places

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -550,7 +564,7 @@ export default function FrameEditPanel({
         <ButtonRow>
           {!isCreateMode ?
             <>
-              <DynamicButton
+              <DynamicButton // Edit mode
```
jaskfla: No need for these comments—`isCreateMode` makes it perfectly clear

packages/frontend/src/components/frames/FrameList.tsx
```diff
@@ -58,10 +61,11 @@ export default function FrameList() {
 
   const sourceImage = useCanvasImage(canvas.id);
 
-  const { data: userFrames = [] } = useUserFrames({
+  const { data: userFramesResponse } = useUserFrames({
     canvasId: canvas.id,
     userId: user?.id,
   });
+  const userFrames = userFramesResponse?.data ?? [];
```
jaskfla, line 68: Another potentially needless `.data`

packages/types/src/api/FrameRequest.d.ts
```diff
@@ -15,7 +15,17 @@ export interface Params {
 
 export type ResBody = Frame[];
 export type FrameByIdResBody = Frame;
-export type UserFramesResBody = UserOwnedFrame[];
-export type GuildFramesResBody = GuildOwnedFrame[];
+
+export type UserFramesResBody = {
```
jaskfla: `interface`

jaskfla (commented):
Just symbol naming nitpicks, I think—no complaints with the logic itself

packages/frontend/src/components/frames/FrameEditPanel.tsx
Rocked03: yeah, `.data` are the frames and `.isAtCountLimit` is how we get the limits

packages/frontend/src/components/frames/FrameEditPanel.tsx
Rocked03: ¯\_(ツ)_/¯

packages/backend/src/routes/api/v1/frame.ts
Rocked03: oh i'm not particularly sold on either, it's a little bit of a wordy thing. Was gonna ask for suggestions myself if I had remembered

the user endpoint has it as a single boolean indicating whether they're at the frame count limit for the requested canvas

the guild endpoint has it as a record <guild id, boolean> since each guild has its own limit

jaskfla: What is the `string` in `Record<string, boolean> ` in this case? Is it `Record<User["id"], boolean>`? Canvas ID?

Haven’t yet had the chance to trace closely enough to understand in human language what this map is

Separately, wWill throw out that we could also consider “quota“ instead of “limit”

Rocked03: the user endpoint is just a single boolean, this Record here is for the guilds

the string is the guild ID

Rocked03:
tested and works as intended :)

jaskfla (approved):
Think this one’s been through enough rounds that I’m happy just to stamp this

Any remaining niggly paper cuts—if any still exist—I trust can be handled down the line 🙂

PR #483, opened 2026-04-28
fix: skip canvas fetch on build

Rocked03:
When the frontend builds, it makes a request to the backend to fetch the default canvas info. This obviously fails because it's just a build and has no access to the backend. It's only a silent non-fatal error, but this fix just skips that erroneous fetch so no more log spam whenever you build the frontend

### DIFF
--- a/packages/frontend/src/app/layout.tsx
+++ b/packages/frontend/src/app/layout.tsx
@@ -55,7 +55,7 @@ async function getServerSideProfile(): Promise<DiscordUserProfile | null> {
 async function getServerSideCanvasInfo(): Promise<CanvasInfo> {
   // Skip during build - data will be fetched fresh on client startup
   if (process.env.npm_lifecycle_event === "build") {
-    return getDefaultCanvasInfo();
+    return defaultCanvasInfo;
   }
 
   try {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/app/layout.tsx
```diff
@@ -62,20 +67,24 @@ async function getServerSideCanvasInfo(): Promise<CanvasInfo> {
...
-      startCoordinates: [1, 1],
-      eventId: 1,
-      webPlacingEnabled: false,
-      allColorsGlobal: false,
-    };
+    return getDefaultCanvasInfo();
   }
 }
 
+function getDefaultCanvasInfo(): CanvasInfo {
```
jaskfla: Any particular reason this is a function and not just an object?

jaskfla (approved):
Thanks for figuring this out. From just reading, I feel like we can just have a `const satisfies CanvasInfo`—no need to lazily get it with a function

packages/frontend/src/app/layout.tsx
Rocked03: was feeling ✨quirky✨

PR #485, opened 2026-04-28
feat(frontend): fullscreen canvas

### LINKED ISSUE #420: Full-screen canvas
Button to make the canvas view fullscreen

Rocked03:
- Added fullscreen functionality to the Canvas View


packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -20,10 +20,18 @@ const Wrapper = styled("div")`
...
   gap: 0.5rem;
   grid-template-rows: auto 1fr;
+  height: 100%;
+  min-height: 0;
   overflow-y: auto; // Fallback property, should appear before overflow-block
   overflow-block: auto;
   padding: var(--padding-width);
 
+  #canvas-container:fullscreen &,
+  #canvas-container:-webkit-full-screen & {
```
jaskfla: Would prefer if we didn’t use magic string for `canvas-container` ID, but not the end of the world

I’d suggest using the generated class from styled-components—or even skipping this qualifier entirely and using `:fullscreen` on its own

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -28,7 +28,9 @@ export default function FramesTab({ setTabsLocked, ...props }: FramesTabProps) {
     FramePanelMode.Info,
   );
 
-  setTabsLocked(activePanel !== FramePanelMode.Info);
+  useEffect(() => {
+    setTabsLocked(activePanel !== FramePanelMode.Info);
+  }, [activePanel, setTabsLocked]);
```
jaskfla, line 33: Is this necessary? I did see that there was an error to do with some sort of `setState`

If this is necessary, just use `useState` to keep the previous active panel rather than an Effect

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -115,6 +131,67 @@ const InviteButton = styled(Button)`
...
 `;
 
+const BaseFullscreenButton = styled(Button, {
+  shouldForwardProp: (prop: string) =>
+    !["$isPanelVisible", "$isFullscreen"].includes(prop),
+})<{ $isPanelVisible?: boolean; $isFullscreen?: boolean }>`
+  inset-inline-end: ${({ $isPanelVisible, $isFullscreen }) =>
+    $isPanelVisible && $isFullscreen ?
+      "calc(min(var(--action-panel-width), calc(100vw - 1rem)))"
+    : "0.5rem"};
```
jaskfla: Instead of `vw`, use `dvi` with `vi` as fallback

Out of scope, but may need to account for `env(safe-area-inset-*)`

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -115,6 +131,67 @@ const InviteButton = styled(Button)`
...
+
+  color: white;
+  position: absolute;
+  text-decoration: none;
+  border-color: transparent;
+  min-width: auto;
+  padding: 0.5rem;
+
+  @media (hover: hover) and (pointer: fine) {
+    :hover {
```
jaskfla: This will target any child with `:hover`. Is that intention?

If not, use `&:hover`

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -115,6 +131,67 @@ const InviteButton = styled(Button)`
...
+      border-color: inherit;
+      box-shadow: 0 0 10px rgba(0 0 0 / 25%);
```
jaskfla:
```suggestion
      box-shadow: 0 0 10px oklch(0 0 0 / 25%);
```

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -432,6 +513,61 @@ export default function CanvasView() {
...
+          webkitRequestFullscreen?: () => Promise<void> | void;
+        })
+      | null;
+    const supportsFullscreen =
+      document.fullscreenEnabled ||
+      !!fullscreenDocument.webkitFullscreenEnabled ||
+      !!fullscreenContainer?.webkitRequestFullscreen;
+
+    setCanUseFullscreen(!isSmall && supportsFullscreen);
+  }, [containerRef.current, isSmall]);
```
jaskfla: Extract this into `useCanUseFullscreen: () => boolean`

Maybe name `useIsFullscreenAvailable`?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -979,9 +1115,64 @@ export default function CanvasView() {
...
+      id="canvas-container"
+      ref={containerRef}
+      onPointerDown={handlePointerDown}
+    >
+      {canUseFullscreen && (
+        <FullscreenButton
+          $isFullscreen={isFullscreen}
+          $isPanelVisible={isFullscreenPanelVisible}
+          aria-label={isFullscreen ? "Exit fullscreen" : "Enter fullscreen"}
+          aria-pressed={isFullscreen}
```
jaskfla, line 1136: Stoked that you’re accounting for this

- Would prefer `VisuallyHidden` node instead of `aria-label` (services like Google Translate ignore `aria-label`s, and in dev you can press <kbd>Alt</kbd> to force `VisuallyHidden` to render visibly)
- `aria-pressed` doesn’t _quite_ make sense semantically here. Not really a toggle button. The label itself does great 👌

packages/frontend/src/contexts/ActionPanelContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+      currentTab,
+      isFullscreenPanelVisible,
+      setAreTabsLocked,
+      setCurrentTab,
+      setFullscreenPanelVisible,
+      setTempColor,
+      tempColor,
+    }),
+    [areTabsLocked, currentTab, isFullscreenPanelVisible, tempColor],
+  );
```
jaskfla: No need to memo-ise `value`, but do need to make sure all its properties are stable

jaskfla (commented):
Just a quick glance—not a proper review

packages/frontend/src/components/action-panel/ActionPanel.tsx
S-Ou: Unfortunately tried that already:
- using `${CanvasContainer}:fullscreen` requires the Emotion compiler
- `:fullscreen` on its own doesn't work because the Action Panel isn't the ref being fullscreen'd

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
Rocked03: Yeah, there was an error, can't remember exactly what but this was the fix

S-Ou: tbh I think I'll let you have a crack at this... I've looked at it a bit and I don't really see any clean way to get to what you're suggesting

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -135,9 +138,10 @@ const BaseFullscreenButton = styled(Button, {
   shouldForwardProp: (prop: string) =>
     !["$isPanelVisible", "$isFullscreen"].includes(prop),
 })<{ $isPanelVisible?: boolean; $isFullscreen?: boolean }>`
+  inset-inline-end: calc(min(var(--action-panel-width), calc(100vi - 1rem)));
   inset-inline-end: ${({ $isPanelVisible, $isFullscreen }) =>
     $isPanelVisible && $isFullscreen ?
-      "calc(min(var(--action-panel-width), calc(100vw - 1rem)))"
+      "calc(min(var(--action-panel-width), calc(100dvi - 1rem)))"
```
jaskfla: If `dvi` is unsupported then the fallback will _always_ apply, regardless of these two prop values

Couple solutions for you to consider, all of which are acceptable.

```css
${(p) =>
  p.$isPanelVisible && p.$isFullscreen ?
    css`
      inset-inline-end: calc(100vi);
      inset-inline-end: calc(100dvi);
    `
  : css`
      inset-inline-end: 0.5rem;
    `}
```

```css
--full-inline-size: 100vi;
--full-inline-size: 100dvi;
inset-inline-end: ${(p) =>
  p.$isPanelVisible && p.$isFullscreen ?
    "calc(var(--full-inline-size))"
  : "0.5rem"};

```

```css
inset-inline-end: 0.5rem;
inset-inline-end: ${(p) =>
  p.$isPanelVisible &&
  p.$isFullscreen &&
  css`
    inset-inline-end: calc(100vi);
    inset-inline-end: calc(100dvi);
  `};
```

In most cases, I would actually prefer…

```jss
const BaseFullscreenButton = styled(Button)`
  inset-inline-end: 0.5rem;
`;

<BaseFullscreenButton style={isFullscreen && isFullscreenPanelVisible ? { insetInlineEnd } : undefined} />
```

…but in this case React’s stupid style object doesn’t let us set a fallback

packages/frontend/src/contexts/ActionPanelContext.tsx
jaskfla, line 1: More state stuff for me to refactor later, eh? 😆

dw I can see how this isn’t straightforward so we can sweep it ~~under the rug~~ into a context for now

packages/frontend/src/components/action-panel/ActionPanel.tsx
jaskfla: Ah, bummer—thanks for explaining 🙏

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -20,10 +20,18 @@ const Wrapper = styled("div")`
...
+    height: auto;
+    max-height: 100%;
```
jaskfla, line 33: Can use `block-size` here

Why `auto`, though? Isn’t that the default?

packages/frontend/src/components/action-panel/ActionPanel.tsx
```diff
@@ -20,10 +20,18 @@ const Wrapper = styled("div")`
   display: grid;
   gap: 0.5rem;
   grid-template-rows: auto 1fr;
+  height: 100%;
+  min-height: 0;
   overflow-y: auto; // Fallback property, should appear before overflow-block
   overflow-block: auto;
   padding: var(--padding-width);
 
+  #canvas-container:fullscreen &,
```
jaskfla: I realise this is my fault, but if you wouldn’t mind: rename `<CanvasContainer id="canvas-container">` to `<CanvasWrapper id="canvas-wrapper">` 🫣

(Or “canvas root” if you prefer that, but we have more wrappers than roots in this codebase iirc)

This “container” is  [not a container](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries)

…also maybe use a class instead of ID?

```js
export const CANVAS_WRAPPER_CLASS_NAME = "canvas-wrapper"`;
```

jaskfla (approved):
Couple notes that will need addressing, but happy to pre-approve from here

I might go in a tweak some button styling in future, but after this round I think this PR is ready for release 🛫

packages/frontend/src/components/action-panel/ActionPanel.tsx
S-Ou: nah, the parent sets it as 100% so it takes the full vertical width, whereas this means that the height of the panel in fullscreen mode fits the content (if it doesn't take the full height)

packages/frontend/src/contexts/ActionPanelContext.tsx
Rocked03: 🙃

packages/frontend/src/components/action-panel/ActionPanel.tsx
S-Ou: Have renamed, but I've kept it as an ID since there should reasonably only be one instance of the canvas wrapper on the page at any one time

PR #492, opened 2026-04-29
fix: frontend environment variables

pumbas600:
**What's Changed?**
- Separates out server component configuration from client component configuration.
- Adds helper to enforce server component configuration is only used server side.
- Fixes missing `NEXT_PUBLIC` prefix to show bot commands environment variable.


packages/frontend/src/config/serverConfig.ts
Rocked03, line 1: if everything else is `serverConfig` then should this be `serverConfig.ts`? 

additionally perhaps `index.ts` -> `clientConfig.ts` as well? although this would change a lot of the imports so perhaps not

Rocked03 (approved):
looks good to me, just one thought but nothing pressing

packages/frontend/src/config/serverConfig.ts
pumbas600: I like it! It did touch a lot of places, but it's a safe change :)

PR #494, opened 2026-04-29
feat: notices

### LINKED ISSUE #405: Notices banner
Visual banner on the page where we can add info about 
- Upcoming events ("Canvas 2026 starts on XX May!")
- Issues ("We know X is broken and we're working on it")
- any other PSAs we might need to make

Rocked03:
- Added notices
  - Displayed as a visual banner on the canvas 
  - Full endpoint support
    - Create/edit/delete UI will be added at a later stage (as an extension of #385) 

image

*that date is not the actual start date, just for testing :)


.vscode/settings.json
```diff
@@ -41,6 +42,7 @@
     "Prismock",
     "SOFTENG",
     "Tāmaki",
+    "Timestamptz",
```
jaskfla:
```suggestion
    "timestamptz",
```

packages/backend/prisma/schema.prisma
```diff
@@ -191,6 +192,19 @@ model session {
   expiresAt DateTime @db.Timestamptz(6)
 }
 
+model notice {
+  id                 Int      @id @default(autoincrement())
+  type               String   @default("info") // info | warning | error
```
jaskfla: enum?

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateTable
+CREATE TABLE "notice" (
+    "id" SERIAL NOT NULL,
+    "type" TEXT NOT NULL DEFAULT 'info',
+    "header" TEXT,
+    "content" TEXT,
+    "priority" INTEGER NOT NULL DEFAULT 0,
+    "active" BOOLEAN NOT NULL DEFAULT false,
+    "persist_on_dismiss" BOOLEAN NOT NULL DEFAULT false,
```
jaskfla: What does this mean?

packages/backend/src/services/noticeService.ts
```diff
@@ -0,0 +1,98 @@
...
+export async function updateNotice({
+  noticeId,
+  type,
+  header,
+  content,
+  priority,
+  active,
+  persistOnDismiss,
+  canvasId,
+}: UpdateNoticeBody): Promise<Notice> {
```
jaskfla, line 116: Signature might be nicer as

```js
updateNotice(id, data)
```

packages/frontend/package.json
```diff
@@ -48,6 +48,7 @@
     "js-cookie": "^3.0.5",
     "lucide-react": "^1.8.0",
     "luxon": "^3.7.2",
+    "markdown-to-jsx": "^9.7.16",
```
jaskfla, line 51: Worth 4.2 MB? 😬

packages/frontend/src/components/notices/NoticeBanner.tsx
jaskfla, line 1: Flagging for myself. Lots we can dedupe here, but I’m just short on time right now so will skip reviewing this file for now

packages/frontend/src/components/notices/Notices.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export default function Notices() {
+  const { data: notices = [] } = useNotices();
+  const { canvas } = useCanvasContext();
+
+  const [persistedDismissed = [], setPersistedDismissed] =
+    useLocalStorage("notices/dismissed");
+
+  const [transientDismissed, setTransientDismissed] = useState<Set<string>>(
+    () => new Set(),
```
jaskfla:
```suggestion
    new Set(),
```

packages/frontend/src/components/notices/Notices.tsx
```diff
@@ -0,0 +1,88 @@
...
+  );
+
+  const persistedSet = useMemo(
+    () => new Set<string>(persistedDismissed ?? []),
```
jaskfla: Does `Set` constructor not accept `undefined`?

packages/frontend/src/styles/core.css
jaskfla, line 1: Is `font-variation-settings` set on `*`? If not, I’ll need to fix 😬

There’s some really non-ergonomic CSS quirks we have to deal with when using `font-variation-settings`

packages/frontend/src/util/text.ts
```diff
@@ -0,0 +1,76 @@
...
+    [24 * 60 * 60 * 1000, "day"],
+    [60 * 60 * 1000, "hour"],
+    [60 * 1000, "minute"],
+    [1000, "second"],
+  ];
+
+  for (const [ms, unit] of units) {
+    if (absDiff >= ms) {
+      const value = Math.round(diff / ms);
+      return new Intl.RelativeTimeFormat(undefined, { numeric: "auto" }).format(
```
jaskfla, line 22: 😄

jaskfla (commented):
Just putting through partial review so keep some momentum

packages/backend/prisma/schema.prisma
Rocked03: syntactically, we could

but the internet seems to quite strongly advise avoiding postgres enums unless we're desperate

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
Rocked03: copying from discord
> `persistOnDismiss`
If true, the notification will still persist and show on reload even if dismissed
If false then it doesn't come back
Quite open for a better name, this was the best I could think of when I did it

packages/frontend/package.json
Rocked03: imo yeah - it does an impressively good job at wrapping text and applying all the right formatting jsx to fit markdown

packages/frontend/src/styles/core.css
Rocked03: I don't think so :/

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateTable
+CREATE TABLE "notice" (
+    "id" SERIAL NOT NULL,
+    "type" TEXT NOT NULL DEFAULT 'info',
+    "header" TEXT,
+    "content" TEXT,
+    "priority" INTEGER NOT NULL DEFAULT 0,
+    "active" BOOLEAN NOT NULL DEFAULT false,
```
jaskfla: Maybe `deleted_at TIMESTAMPTZ`?

packages/backend/prisma/schema.prisma
jaskfla: Yeah fair enough. Inclined to agree about pg enum, though wonder what we can do to keep the ts type narrow

Can kick that can down the road, though

packages/frontend/package.json
jaskfla: Wait, what’s text wrapping got to do with this? Isn’t that just done by the browser?

My gut is that this dependency is overkill for us to show some banners. I’d push for for storing HTML instead of Markdown in the db and use the [HTML Sanitizer API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration) to take care of injection risks

Unfortunately Safari support isn’t there yet, but we can follow after pretty conventional email APIs that basically look like `{ html, plaintext }` where we fall back to a plaintext version if we can’t safely render the HTML

Critically, this leaves open the option for us to switch to Markdown, since HTML is valid Markdown. But if we commit to Markdown now, we cannot switch to HTML without migrating data

packages/frontend/src/app/Main.tsx
```diff
@@ -40,7 +40,7 @@ const Wrapper = styled("main")`
 export default function Main() {
   return (
     <Wrapper>
-      <CanvasView />
+      <CanvasView showNotices />
```
jaskfla: Premature prop? If it’s only ever `true`, then can just unconditionally render `<Notices>`

packages/frontend/src/util/text.ts
```diff
@@ -0,0 +1,76 @@
+// Datetime regex: <t:timestamp:format> or <t:timestamp>
+// Discord's format
+const DATETIME_REGEX = /<t:(\d{1,}):?([tTdDfFR])?>/g;
```
jaskfla:
```suggestion
const DATETIME_REGEX = /<t:(\d+):?([tTdDfFR])?>/g;
```

packages/frontend/src/util/text.ts
jaskfla, line 1: Re file name: `text` feels too generic—would `timestamp` be appropriate? Or even `discord-timestamp` since this is emulating their behaviour

Re module: Bit conflicted, because Discord’s timestamp does more than we need. Personally much prefer to represent datetime with an ISO string, but if you do anticipate needing the flags for display various formats then def nicer to stick with an existing format where someone else has done all the thinking for us

packages/frontend/src/app/Main.tsx
Rocked03: Not quite, this is prep for the mod dashboard which also uses the canvas view but needs to hide some irrelevant stuff

packages/frontend/package.json
Rocked03: Discussed off-platform-- going to keep `markdown-to-jsx` since html isn't a particularly viable option for what will eventually be user input (notice creation will be built into the future admin dashboard)

packages/frontend/src/util/text.ts
Rocked03: I made it `text` since it then leaves it open for expansion in the future. I'm not entirely sure what yet, but it'd be neat to build in other text functionality here if we ever choose to 

and arguably, it's too early to say whether Discord's timestamp format is too much yet. It's nice to have the options, and it's a system that already works quite neatly on Discord

packages/frontend/src/app/Main.tsx
jaskfla: Is that coming in this PR? Cleaner to leave it out if mod dashboard is coming in a different PR

packages/frontend/src/styles/core.css
jaskfla: Oh, it’s set in `html` immediately above. This change should be reverted

If there’s a specificity thing going on we can fix it locally, but it’s overkill to make this change in the global stylesheet

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
jaskfla: Is this attribute now dead because of a46d50fab20d8bbb150215f9bdd5de8e4a6c01e2?

packages/frontend/src/util/text.ts
jaskfla: Yeah simmered on it and simpler just to use Discord’s format

packages/backend/src/services/noticeService.ts
```diff
@@ -0,0 +1,93 @@
+import type { Notice, NoticeType } from "@blurple-canvas-web/types";
+import { prisma } from "@/client";
+import type { CreateNoticeBody } from "@/models/notice.models";
+
+type NoticeDbRecord = NonNullable<
+  Awaited<ReturnType<typeof prisma.notice.findFirst>>
+>;
```
jaskfla: Surely Prisma exports a an equivalent type

packages/frontend/src/styles/core.css
Rocked03: oh, because it's set in the html above it doesn't get rerendered with the new variable

as least, best I could tell, but without this addition, the slant wasn't being applied

packages/frontend/src/app/Main.tsx
Rocked03: No, #500... can omit in this PR

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
Rocked03: no - i've still left it because i feel like some notices should still only have "dismiss" and not "don't show again", so this controls it

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
jaskfla: Think we resolved offline to reject this suggestion

packages/frontend/src/styles/core.css
jaskfla: When you say “re-render” do you mean React render lifecycle or browser repaint?

Regardless, I can have a looksie at this. There must just be something else in the cascade that needs overriding, but not from root

packages/frontend/src/components/notices/NoticeBanner.tsx
jaskfla: Might actually just leave be and I can do a redesign PR later

packages/backend/src/models/notice.models.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface NoticeIdParam {
+  noticeId: number;
+}
+
+export async function parseNoticeId(
+  params: NoticeIdParam,
+): Promise<NoticeIdParam["noticeId"]> {
+  const result = await NoticeIdParamModel.safeParseAsync(params);
+  if (!result.success) {
+    throw new BadRequestError(
```
jaskfla, line 17: This should be `Not Acceptable`, not `Bad Request`

packages/backend/prisma/migrations/20260430120000_add_notice_table/migration.sql
jaskfla: Ah, finally managed to place my finger on why this name feels wrong:

- “on” makes it feel like it responds to an event; and
- to a lesser degree, “persist” makes it sound like data is being written somewhere

I still prefer `stubborn`, but that’s a bit non-obvious. Maybe something along the lines of `always_notify` or `notify_after_user_dismissal`

jaskfla (approved):
Pre-approving, but still think we would do well to rename `persist_on_dismiss`

packages/backend/src/models/notice.models.ts
S-Ou: gonna do in separate PR, there's a few here with that issue

PR #497, opened 2026-04-30
fix: improve colour picker accessibility

jaskfla:
Still a fair bit of room to improve semantics. In particular, the **Place** tab should really be a `<form>`, but at least now the colours are actually labelled and indicate their state.

Prior to this PR, tabbing through the colours with VoiceOver just says “group, group, group, …”. Now “Discord Red, Brilliance Red, Hypesquad Yellow…”


packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
```diff
@@ -144,38 +147,42 @@ export default function PlacePixelTab({
...
-                  selected={color === selectedColor}
-                />
-              ))
-            : Array.from({ length: 12 }).map((_, i) => (
-                // biome-ignore lint/suspicious/noArrayIndexKey: These will never change
-                <SwatchSkeleton key={i} variant="rectangular" />
-              ))
-            }
+            <Fieldset>
+              {mainColors?.length ?
```
Rocked03: should we be using `isLoading` here?

Rocked03 (approved):
just one thought, but otherwise it looks pretty fine to me 👍

packages/frontend/src/components/action-panel/tabs/PlacePixelTab.tsx
jaskfla: There are a couple issues I wanna iron out here, actually, but yes, good call

Technically the behaviour comes out the same because `mainColors` is only ever expected to be `undefined` when `isLoading`, but using `isLoading` is much more readable. I’ll fix that in this PR tomorrow

The other thing is that this code doesn’t handle the case of an empty array, which we also never expect but it’s an easy thing to cover

Rocked03: i suppose we should just have some conscious handling of the edge case of if the database refuses to connect etc

jaskfla: Ohh, actually, here we need to prove to TypeScript that `mainColors` and `partnerColors` are arrays and not `undefined`, which is why I didn’t use `isLoading`

PR #503, opened 2026-05-04
fix: fix migrations

Rocked03:
lol i thought the migrations were fine and then i added a ci check and turns out they weren't :skull:

- also added the functions since we can do that in migrations now


jaskfla (approved):
Nice catch thanks

PR #504, opened 2026-05-04
feat: save guild records on auth

Rocked03:
- This used to be handled by the bot being in these servers, but instead this makes the web app not rely on the bot
- When a user logs in, we already get the list of their guilds, but now we save it to the database
- The web app uses this table to grab the guild names in places like listing guild frames


packages/backend/src/services/discordGuildService.ts
```diff
@@ -205,3 +206,54 @@ export function assertIsCanvasModerator(
...
+  const existingMap = new Map(existing.map((r) => [r.guild_id.toString(), r]));
+
+  // 2) compute create + update sets
+  const toCreate = entries
+    .filter(([id]) => !existingMap.has(id))
+    .map(([id, data]) => ({ guild_id: BigInt(id), name: data.name }));
+
+  const toUpdateEntries = entries.filter(([id, data]) => {
+    const ex = existingMap.get(id);
+    return !!ex && ex.name !== data.name;
```
jaskfla: Non-blocking stylistic detail: `!!` or `Boolean`? I think `Boolean` is more intentional and is worth the verbosity

```suggestion
    return Boolean(ex) && ex.name !== data.name;
```

packages/backend/src/services/discordGuildService.ts
```diff
@@ -205,3 +206,54 @@ export function assertIsCanvasModerator(
...
+  // 3) Create missing rows
+  if (toCreate.length > 0) {
+    await prisma.discord_guild_record.createMany({
+      data: toCreate,
+      skipDuplicates: true,
+    });
+  }
+
+  // 4) Update changed names in bounded parallel chunks
+  const UPDATE_CHUNK = 50;
```
jaskfla, line 247: Neat

jaskfla (approved):

packages/backend/src/services/discordGuildService.ts
Rocked03: no strong opinions, happy to go with that

Rocked03: might revert, build error needs ex to be not null for `ex.name` and `!!ex` already did the trick proving it wasn't null

PR #509, opened 2026-05-06
fix(backend): honour Discord rate limits

jaskfla:


packages/backend/src/utils/fetchWithRetries.ts
```diff
@@ -0,0 +1,77 @@
...
+} as const;
+
+export default async function fetchWithRetries(
+  input: string | URL | Request,
+  init?: RequestInit,
+  { maxAttempts, backoff, statusCodes }: RetryOptions = defaultRetryConfig,
+) {
+  if (process.env.NODE_ENV !== "production" && backoff < 1) {
+    throw new NotAcceptableError(
+      `RetryOptions backoff must be ≥1.0, but got ${backoff}`,
```
Rocked03, line 50: `≥` is nice but is it universally supported in whatever font might be printing this?

Rocked03 (approved):
looks good

turns out i also made NotAcceptableError in #514 but that's no issue, let's complete this first and i'm sure the conflicts will sort themselves out super easy anyways

packages/backend/src/utils/fetchWithRetries.ts
jaskfla: Fonts almost invariably are rendered with fallback fonts. If UTF-8 encodes it, I’m not really worried. 

Hell, [Mojo](https://www.modular.com/open-source/mojo) uses `.🔥` as its default file extension

PR #515, opened 2026-05-06
refactor(frontend): clipboard utils

Rocked03:


packages/frontend/src/util/index.ts
jaskfla, line 1: Did you run into issues? `writeText` is widely available

I haven’t seen this workaround since was in intermediate 😵‍💫

Rocked03: I don't think the legacy stuff is necessary, but copilot insisted I made my copyToClipboard a bit better

Rocked03: it was moreso `copyToClipboard(colorCode);` -> `void copyToClipboard(colorCode);` but copilot decided this was necessary too

jaskfla: I reckon just `() => void navigator.clipboard.writeText()` and ditch the util

Rocked03: 🤷

PR #519, opened 2026-05-07
feat(frontend): moderation dashboard - blocklist tab

### LINKED ISSUE #541: Moderation Dashboard - Blocklist

Rocked03:


packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+
+const BlocklistEntryTable = styled("table")`
+  border-collapse: separate;
+  border-spacing: 0 0.5rem;
+  margin-top: 1rem;
+  width: 100%;
+
+  * > th {
+    text-align: left;
+  }
```
jaskfla: Add `th { text-align: start }` to the normalize/reset layer of `core.css`

**Update:** Missed this earlier, but is `th` even a legal child of `table`??

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+
+const BlocklistBodyWrapper = styled("div")`
+  display: flex;
+  flex-direction: column;
+  gap: 0.5rem;
+`;
+
+const BlocklistEntryTable = styled("table")`
+  border-collapse: separate;
+  border-spacing: 0 0.5rem;
```
jaskfla, line 25: Really? o.O

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+  }
+`;
+
+const StyledCheckboxSetting = styled(CheckboxSetting)`
+  padding-block: 0.5rem;
+  padding-inline: 0.25rem;
+`;
+
+const StyledEntryRow = styled("tr")`
+  :hover {
```
jaskfla: Target it, not its descendants

```suggestion
  &:hover {
```

And pop it in that hover media query

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+    background-color: oklch(from var(--discord-white) l c h / 10%);
```
jaskfla: 10% is a big aggressive. Maybe 3–6%

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+  }
+`;
+
+const StyledUserRow = styled("td")`
+  display: flex;
```
jaskfla, line 68: Changing `display` of a table element breaks its semantics from a screen reader’s perspective. HTML spec for table is quite strict, and requires nodes of specific roles in the tree to be arranged quite precisely.

If you need flex layout[^foot], wrap the contents in a flexbox

[^foot]: You almost certainly do not

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+> {
+  user: BlocklistEntry;
+}
+
+function BlocklistUserEntry({
+  user,
+  "aria-busy": ariaBusy,
+  checked,
+  onChange,
+}: BlocklistUserEntryProps) {
```
jaskfla, line 106: Why just these props? Not passing global attributes breaks mental model imo

Consider using `Omit` instead of `Pick`—if one is needed at all—and delegate remaining props to root element

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+          <VisuallyHidden>User ID {userId}. Click to copy.</VisuallyHidden>
+          <Copy size={12} />
+        </UserId>
+      </StyledUserRow>
+      <td>
+        {new Date(user.dateAdded).toLocaleDateString("en-GB", {
+          day: "numeric",
+          month: "short",
+          year: "numeric",
+        })}
```
jaskfla: - Wrap this in a `<time dateTime={user.dateAdded.toISOString()}>`
- Great call not using purely numeric date format 💪
- This is user-facing, so I guess we should use US locale 😔 (…or do we just defer to user locale??)

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+  flex-direction: column;
+  gap: 0.25rem;
+  padding-block: 0.25rem;
+`;
+
+const StyledUsername = styled("span")`
+  min-width: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
jaskfla: Hmmmmmmmmmmmmmmmmmm

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+`;
+
+const StyledInput = styled(Input)`
+  flex: 1;
+  min-width: 0;
+`;
```
jaskfla, line 94: Logical props are more appropriate here: `inline-size` instead of `width`

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+  const [userIdsToBlock, setUserIdsToBlock] = useState<bigint[]>([]);
+
+  const existingBlocklistIdStrings = useMemo(
+    () => new Set(blocklist.map((entry) => entry.userId.toString())),
+    [blocklist],
+  );
+
+  function setUserSelected(
+    event: React.ChangeEvent<HTMLInputElement>,
+    userId: bigint,
```
jaskfla:
```suggestion
    userId: User["id"],
```

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+    });
+  }
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: selectedUsers is not included as we don't want the list to re-order when selection changes
+  const displayList = useMemo(() => {
+    if (!searchQuery) {
+      // No filter: selected are raised to the top
+      const selected = blocklist.filter((u) => selectedUsers.has(u.userId));
+      const unselected = blocklist.filter((u) => !selectedUsers.has(u.userId));
+      return [...selected, ...unselected];
```
jaskfla: Isn’t this just sorting?

```ts
return blocklist.sort(u => has ? -1 : 1)
```

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+    <BlocklistTabBlock {...props}>
+      <FullWidthScrollView>
+        <ActionPanelTabBody>
+          <BlocklistBodyWrapper>
+            <PanelSectionHeading>Blocklist</PanelSectionHeading>
+
+            <StyledInput
+              disabled={isLoading || blocklist.length === 0}
+              onChange={(e) => setSearchQuery(e.target.value)}
+              placeholder="Search users..."
```
jaskfla:
```suggestion
              placeholder="Search users…"
```

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+import {
+  ActionPanelTabBody,
+  FullWidthScrollView,
+  TabPanel,
+} from "@/components/action-panel/tabs/ActionPanelTabBody";
+import { UserId } from "@/components/complex-search/SearchUserEntry";
+import { Input } from "@/components/input/Input";
+import VisuallyHidden from "@/components/VisuallyHidden";
+import { useBlocklist } from "@/hooks/queries/useBlocklist";
+import CheckboxSetting from "../settings/CheckboxSetting";
```
jaskfla: I think this is an abuse of the component?

Setting semantics ≠ blocklist user entry semantics

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+            : blocklist.length === 0 ?
+              <p>The blocklist is currently empty.</p>
+            : displayList.length === 0 ?
+              <p>No users match your search.</p>
+            : <BlocklistEntryTable>
+                <thead>
+                  <tr>
+                    <th /> {/* Checkbox column */}
+                    <th>User</th>
+                    <th>Date Added</th>
```
jaskfla: Sentence case everywhere

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+              value={searchQuery}
+              style={{
+                width: "100%",
+              }}
+            />
+
+            {isLoading ?
+              <p>Loading...</p>
+            : blocklist.length === 0 ?
+              <p>The blocklist is currently empty.</p>
```
jaskfla: - No full stop unless multiple sentences
- Is `<ContentUnavailableView>` in this branch? I suspect it might not be. Ignore if it doesn’t exist yet

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+              placeholder="Search users..."
+              type="text"
+              value={searchQuery}
+              style={{
+                width: "100%",
+              }}
+            />
+
+            {isLoading ?
+              <p>Loading...</p>
```
jaskfla: Skeleton could be nice, but in the interest of time, just this will do:

```suggestion
              <p>Loading…</p>
```

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
...
+              onChange={(e) => setSearchQuery(e.target.value)}
+              placeholder="Search users..."
+              type="text"
+              value={searchQuery}
+              style={{
+                width: "100%",
+              }}
+            />
+
+            {isLoading ?
```
jaskfla, line 261: If something is loading, there should be a node with `aria-busy`

Most appropriate one here is probably `<BlocklistBodyWrapper>`. Even more appropriate would be to wrap this `{}` block

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
+import styled from "@emotion/styled";
+import { Chip, css, TextField } from "@mui/material";
```
jaskfla, line 2: - `Chip` is trivial ti implement ourselves
- `TextField` should be thinner wrapper of `<input>`, or just raw-dog it

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import axios from "axios";
+import { TriangleAlert } from "lucide-react";
+import { useState } from "react";
+import { StyledButton } from "@/components/button/DynamicButton";
```
jaskfla: Huh? `StyledButton` sounds like it’s internal to `DynamicButton`—how come it’s exported?

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+import { AutocompleteInput } from "@/components/input/Input";
+import config from "@/config/clientConfig";
+
+const BlocklistFooter = styled("div")`
```
jaskfla: `<footer>`

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+function parseUserIds(value: string): bigint[] {
+  return value
+    .split(/[\s,]+/)
+    .map((id) => id.trim())
+    .filter((id) => /^\d+$/.test(id))
+    .map((id) => BigInt(id));
+}
+
+function normalizeUserIds(values: readonly bigint[]): bigint[] {
```
jaskfla: Can inline this

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+  const inputtedUsersAlreadyBlocked = userIdsToBlock.some((id) =>
+    existingBlocklistIdStrings.has(id.toString()),
+  );
+
+  return (
+    <BlocklistAddWrapper>
+      {inputtedUsersAlreadyBlocked && (
+        <BlocklistWarning>
+          <TriangleAlert size={14} /> You have listed users that are already
+          blocked.
```
jaskfla:
```suggestion
          blocked
```

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
jaskfla, line 1: I wonder if we consider adding `es-toolkit` as a dependency 🤔

Seems like we would benefit from `uniq` and `omit` here in the frontend; and the backend could probably benefit from `toCamelCaseKeys`

I’m ambivalent. I of course avoid deps like the plague, but as far as libraries go `es-toolkit` is pretty reasonable

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+        </BlocklistWarning>
+      )}
+      <BlocklistAddBody>
+        <BlocklistAutocompleteWrapper>
+          <AutocompleteInput
+            freeSolo
+            fullWidth
+            multiple
+            options={[]}
```
jaskfla: Eh?

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+  onBlock?: (userIds: bigint[]) => void;
+  isBlocking?: boolean;
+}
+
+interface BlocklistRemoveSectionProps {
+  selectedUsers: Set<bigint>;
+  userIdsToBlock: bigint[];
+  onRemove?: (userIds: bigint[]) => void;
+  isRemoving?: boolean;
+}
```
jaskfla, line 90: User IDs can be typed as `User["id"]`

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+        </Button>
+      </BlocklistAddBody>
+    </BlocklistAddWrapper>
+  );
+}
+
+function BlocklistRemoveSection({
+  selectedUsers,
+  onRemove,
+  isRemoving,
```
jaskfla: I trust this state can be derided from `useIsMutating`?

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+}: BlocklistRemoveSectionProps) {
+  return (
+    <Button
+      disabled={selectedUsers.size === 0 || Boolean(isRemoving)}
+      onClick={() => onRemove?.([...selectedUsers])}
```
jaskfla: Useless clone of array here, right?

I’m assuming `onRemove` is expected/required to be pure

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+  const removeFromBlocklistMutation = useMutation({
+    mutationFn: async (ids: bigint[]) => {
+      const url = `${config.apiUrl}/api/v1/blocklist`;
+      await axios.delete(url, {
+        data: { userId: ids.map((id) => id.toString()) },
+        withCredentials: true,
+      });
+    },
+    onSuccess: () => queryClient.invalidateQueries({ queryKey: ["blocklist"] }),
+  });
```
jaskfla: Could be nice to extract the mutation hooks into `./queries`

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  async function handleAdd(ids: bigint[]) {
+    try {
+      await addToBlocklistMutation.mutateAsync(ids);
+      onUserIdsToBlockChange([]);
+    } catch (e) {
+      console.error(e);
+      if ((e as { response?: { status?: number } }).response?.status === 401) {
+        alert("Your session has expired. Please log in again.");
```
jaskfla: Very much out of scope, but should be bring in Sonner post-event?

packages/frontend/src/components/complex-search/SearchUserEntry.tsx
```diff
@@ -49,7 +49,7 @@ const ColorChipWrapper = styled("div")`
   overflow-x: auto;
 `;
 
-const UserId = styled("button")`
+export const UserId = styled("button")`
```
jaskfla: Now that this is exposed, it needs a more descriptive name

“UserId” doesn’t exactly imply `<button>`

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
```diff
@@ -0,0 +1,320 @@
+import styled from "@emotion/styled";
+import { Chip, css, TextField } from "@mui/material";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import axios from "axios";
+import { TriangleAlert } from "lucide-react";
+import { useState } from "react";
+import { StyledButton } from "@/components/button/DynamicButton";
+import { AutocompleteInput } from "@/components/input/Input";
```
jaskfla, line 7: This would be simpler as an `<input>` with a [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist)

Won’t make you refactor, though

packages/types/src/blocklist.d.ts
```diff
@@ -1,4 +1,8 @@
+import type { DiscordUserProfile } from "./discordUserProfile";
+
 export interface BlocklistEntry {
   userId: bigint;
```
jaskfla, line 4: This can be derived too, right?

jaskfla (dismissed):
Just imagine that everywhere I refer to the `User` type, I actually meant `DisordUserProfile` 😓

Overall architecture looks neat to me. Usual pickings about semantics ’n stuff, including some table-related gotchas that you’d be completely forgiven for not knowing about

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -1,5 +1,115 @@
...
+      <td>
+        <StyledCheckboxSetting
+          aria-busy={ariaBusy}
+          checked={checked}
+          onChange={onChange}
+          label={null}
+        />
+      </td>
+      <StyledUserRow title={username}>
+        <StyledUsername>{username}</StyledUsername>
```
jaskfla: Appreciate that you’ve thought to add `title` because of the truncation; but simpler and more accessible to just let the username reflow and have the table row grow to accommodate

packages/frontend/src/app/moderation/BlocklistTab.tsx
```diff
@@ -8,5 +118,114 @@ const BlocklistTabBlock = styled(TabPanel)`
 export default function BlocklistTab(
   props: React.ComponentPropsWithoutRef<typeof BlocklistTabBlock>,
 ) {
-  return <BlocklistTabBlock {...props}>Blocklist</BlocklistTabBlock>;
+  const { data: blocklist = [], isLoading } = useBlocklist();
+
+  const [selectedUsers, setSelectedUsers] = useState<Set<bigint>>(new Set());
```
jaskfla, line 159: This smells overly complex. What was the motivation for not using `<input type="checkbox">`s with the same `name` and letting the `<form>` submit handler give us state?

I’ve only read the code once over, but I can’t see any reason these inputs _really_ need to be controlled. It kinda looks like reimplementing browser logic

packages/frontend/src/components/input/Input.tsx
```diff
@@ -0,0 +1,55 @@
...
+  Input as MuiInput,
+  styled,
+} from "@mui/material";
+
+export const Input = styled(MuiInput)`
+  border: oklch(from var(--discord-white) l c h / 12%) 3px solid;
+  border-radius: 0.5rem;
+  padding: 0.25rem 0.75rem;
+  transition-duration: var(--transition-duration-fast);
+  transition-property: background-color, border-color, color, scale;
```
jaskfla: `scale`???

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: would like to but the type is a string rather than a bigint, since json doesn't play well with bigints over http

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
S-Ou: same comment as earlier-- our user type is string since http doesn't like bigints

it's a problem across the whole frontend, we ought to fix eventually

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
S-Ou: coulda sworn i put a comment there

S-Ou: I don't want to provide options here, it's really just more for the chips

packages/frontend/src/app/moderation/BlocklistTabFooter.tsx
S-Ou: keen

packages/frontend/src/components/input/Input.tsx
S-Ou: i think i copied that from the button which does use scale

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: arguably, the only thing that makes CheckboxSetting specific to the settings page is the styling, the rest of the functionality is rather generic

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: spacing ¯\\\\_(ツ)_/¯

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: lol

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: i've wrapped in a flex for now, but what else do you suggest to stack them?

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: hmm?

packages/frontend/src/app/moderation/BlocklistTab.tsx
S-Ou: ok i've moved away from it

stijnvdkolk (approved):
backend code approved

Rocked03:
> backend code approved

good enough for me, @jaskfla is welcome to tidy up the front-end at a later stage

PR #520, opened 2026-05-07
feat: animated canvas icon loader

Rocked03:
Replacing the old loader

https://github.com/user-attachments/assets/9e30882c-be0e-4fb2-b5fa-e738b6636373


Rocked03:
we could, in theory, have this in other loading states too like the place pixel button when it's registering the pixel placement

packages/frontend/src/components/CanvasAnimatedIcon.tsx
jaskfla, line 1: Oooooooookaaaaayyyyyy please tell Copilot this is not a reasonable implementation

I’ve seen the animation—not a single line of JavaScript is required to implement it. Try getting it to refactor it to be a CSS-only `animation` 💀

jaskfla (changes requested):

packages/frontend/src/components/CanvasAnimatedIcon.tsx
jaskfla: I do like the resultant animation, though ✨

packages/frontend/src/styles/core.css
```diff
@@ -159,6 +159,7 @@
 @layer core {
   html {
     --discord-blurple: oklch(57.74% 0.209 273.85);
+    --discord-dark-blurple: oklch(0.4859 0.1736 274.11);
```
jaskfla: This… is not in the colour system

It also looks _very close_ to a colour that could be derived from `--discord-blurple`

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1256,7 +1257,15 @@ export default function CanvasView() {
           <ActionPanel />
         </FullscreenPanelOverlay>
       )}
-      {isLoading && <CircularProgress style={{ position: "absolute" }} />}
+      {isLoading && (
+        <CanvasAnimatedIcon
+          loop
+          pattern="mixed"
+          lightColor="oklch(from var(--discord-blurple) l c h / 80%)"
+          darkColor="oklch(from var(--discord-dark-blurple) l c h / 80%)"
```
jaskfla: This API won’t do. It should take `style={{ color }}`. The light and dark colours should be internal to the component, where `lightColor` can simply be `currentColor` and `darkColor` can be derived with relative colour syntax

**Update:** I would also consider tweaking _opacity_ instead of darkening the colour

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+    return () => {
+      timeoutsRef.current.forEach(clearTimeout);
+      timeoutsRef.current = [];
+    };
+  }, [loop, runCycle, resetAll, schedule]);
+
+  return (
+    <div
+      aria-label="Loading"
+      role="status"
```
jaskfla: ARIA role should be `progress`

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    return () => {
+      timeoutsRef.current.forEach(clearTimeout);
+      timeoutsRef.current = [];
+    };
+  }, [loop, runCycle, resetAll, schedule]);
+
+  return (
+    <div
```
jaskfla: Probably render an SVG? Makes scaling more sensible, especially considering the border radius of the “pixels”

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+function shuffled(arr: number[]): number[] {
+  const a = [...arr];
+  for (let i = a.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+    [a[i], a[j]] = [a[j], a[i]];
+  }
+  return a;
+}
+
+export type AnimationPattern = "ripple" | "cascade" | "random" | "mixed";
```
jaskfla: This is cool, and it’s good to try a few variations, but let’s pick one and ditch the `pattern` configuration

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+   */
+  darkColor?: string;
+  /** Whether to loop the animation. Default: false */
+  loop?: boolean;
+  /** Pause (ms) between ripple-in finishing and ripple-out starting. Default: 600 */
+  holdPause?: number;
+  /** Pause (ms) after ripple-out finishes before next loop. Default: 400 */
+  restPause?: number;
+  className?: string;
+  style?: CSSProperties;
```
jaskfla: A lot of these can be hard-coded. No need to make `stepDelay` or `holdPause` etc configurable

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+          key={`sq-${index}`}
+          ref={(el) => {
+            squaresRef.current[index] = el;
+          }}
+          style={{
+            width: squareSize,
+            height: squareSize,
+            borderRadius: radius,
+            backgroundColor:
+              SQUARE_COLORS[index] === "light" ? lightColor : darkColor,
```
jaskfla: There are only 9 pixels—just [`:nth-child()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:nth-child) this

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+        gap,
+        padding: gap,
+        boxSizing: "border-box",
+        flexShrink: 0,
+        ...style,
+      }}
+    >
+      {ALL_INDICES.map((index) => (
+        <div
+          key={`sq-${index}`}
```
jaskfla: Moot point because I’m requesting refactor to CSS animation, but this is _no different_ from `key={index}` smh my head Copilot

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+  lightColor = "#ffffff",
+  darkColor = "#454FBF",
+  loop = false,
+  holdPause = 600,
+  restPause = 400,
+  className,
+  style,
+}: CanvasAnimatedIconProps) {
+  const squaresRef = useRef<(HTMLDivElement | null)[]>([]);
+  const timeoutsRef = useRef<ReturnType<typeof setTimeout>[]>([]);
```
jaskfla: Unhinged to run this animation on the main thread, but I give it points for at using refs to at least _suggest_ that it _might_ be thinking about UI responsiveness 😆

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  const gap = Math.round(size * gapRatio);
+  const squareSize = Math.round((size - gap * 4) / 3);
+  const radius = borderRadius ?? Math.round(squareSize * 0.18);
+  const outDuration = Math.round(duration * 1);
```
jaskfla: Excuse me???? 🥴

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,243 @@
...
+  restPause = 400,
+  className,
+  style,
+}: CanvasAnimatedIconProps) {
+  const squaresRef = useRef<(HTMLDivElement | null)[]>([]);
+  const timeoutsRef = useRef<ReturnType<typeof setTimeout>[]>([]);
+
+  const gap = Math.round(size * gapRatio);
+  const squareSize = Math.round((size - gap * 4) / 3);
+  const radius = borderRadius ?? Math.round(squareSize * 0.18);
```
jaskfla: Yeah, this is why I suggest rendering an SVG

Moot, but this could’ve been gotten around with using `border-radius: 18%`

packages/frontend/src/styles/core.css
Rocked03: This comes from the Blurple Canvas colour palette

packages/frontend/src/components/canvas/CanvasView.tsx
Rocked03: personally, i'm keen to keep the colours separate since it then gives the flexibility to have a two-toned loader

jaskfla: Reasonable enough, but call it `secondary` instead of `dark`

(I also assume you mean two-colour, because two-tone is what we already have here)

Rocked03: i totally knew the difference i'm a colour expert

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+   * - --canvas-icon-primary: light square color
+   * - --canvas-icon-secondary: dark square color
+   */
+  style?: CanvasAnimatedIconStyle;
+}
+
+export default function CanvasAnimatedIcon({ style }: CanvasAnimatedIconProps) {
+  return (
+    <StyledWrapper style={style}>
+      <StyledSvg aria-label="Loading" role="progressbar" viewBox="0 0 96 96">
```
jaskfla: Use `<title>` instead of `aria-label`

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+const StyledSquare = styled("rect")<{ $delayMs: number; $tone: Tone }>`
+  transform-box: fill-box;
+  transform-origin: center center;
+  fill: ${({ $tone }) =>
+    $tone === "primary" ? "currentColor" : (
+      "var(--canvas-icon-secondary, oklch(from var(--discord-blurple) l c h / 0.8))"
+    )};
+
+  animation: rippleCycle 3000ms infinite;
+  animation-delay: ${({ $delayMs }) => `${$delayMs}ms`};
```
jaskfla: Pass `--index` from JavasScript and make this

```suggestion
  animation-delay: calc(var(--index) * 80ms);
```

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  height: 100%;
+  width: 100%;
+`;
+
+const StyledSquare = styled("rect")<{ $delayMs: number; $tone: Tone }>`
+  transform-box: fill-box;
+  transform-origin: center center;
+  fill: ${({ $tone }) =>
+    $tone === "primary" ? "currentColor" : (
+      "var(--canvas-icon-secondary, oklch(from var(--discord-blurple) l c h / 0.8))"
```
jaskfla:
```suggestion
      "var(--canvas-icon-secondary, oklch(from currentColor l c h / 0.8))"
```

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  { index: 0, x: 8, y: 8, tone: "primary", delayMs: 0 },
+  { index: 1, x: 36, y: 8, tone: "primary", delayMs: 80 },
+  { index: 2, x: 64, y: 8, tone: "secondary", delayMs: 240 },
+  { index: 3, x: 8, y: 36, tone: "primary", delayMs: 160 },
+  { index: 4, x: 36, y: 36, tone: "secondary", delayMs: 320 },
+  { index: 5, x: 64, y: 36, tone: "primary", delayMs: 480 },
+  { index: 6, x: 8, y: 64, tone: "secondary", delayMs: 400 },
+  { index: 7, x: 36, y: 64, tone: "primary", delayMs: 560 },
+  { index: 8, x: 64, y: 64, tone: "primary", delayMs: 640 },
+] as const satisfies Square[];
```
jaskfla, line 80: - Personally I’d name it `squares` because it’s not exported, but that’s a nitpick
- Unhinged `index` property 🥴

```suggestion
interface Square {
  index: number;
  x: number;
  y: number;
  tone: Tone;
  delayMs: number;
}

const SQUARES = [
  { x: 8, y: 8, tone: "primary", delayMs: 0 },
  { x: 36, y: 8, tone: "primary", delayMs: 80 },
  { x: 64, y: 8, tone: "secondary", delayMs: 240 },
  { x: 8, y: 36, tone: "primary", delayMs: 160 },
  { x: 36, y: 36, tone: "secondary", delayMs: 320 },
  { x: 64, y: 36, tone: "primary", delayMs: 480 },
  { x: 8, y: 64, tone: "secondary", delayMs: 400 },
  { x: 36, y: 64, tone: "primary", delayMs: 560 },
  { x: 64, y: 64, tone: "primary", delayMs: 640 },
] as const;
```

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+    64% {
+      opacity: 0;
+      transform: scale(0);
+    }
+
+    100% {
+      opacity: 0;
+      transform: scale(0);
+    }
+  }
```
jaskfla, line 61: Surely this can be achieved with three keyframe: `0%`, `50%` and `100%`?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1256,7 +1257,18 @@ export default function CanvasView() {
           <ActionPanel />
         </FullscreenPanelOverlay>
       )}
-      {isLoading && <CircularProgress style={{ position: "absolute" }} />}
+      {isLoading && (
+        <CanvasAnimatedIcon
+          style={{
+            position: "absolute",
```
jaskfla: Curious that there was `position: absolute` with no `inset`

I presume removing it breaks?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1256,7 +1257,18 @@ export default function CanvasView() {
...
+            "--canvas-icon-primary":
+              "oklch(from var(--discord-blurple) l c h / 0.55)",
+            "--canvas-icon-secondary":
+              "oklch(from var(--discord-blurple) l c h / 0.35)",
```
jaskfla: This is _awfully_ similar to just setting `opacity: 80%` and letting the rest use the internal defaults

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  transform-origin: center center;
+  fill: ${({ $tone }) =>
+    $tone === "primary" ? "currentColor" : (
+      "var(--canvas-icon-secondary, oklch(from var(--discord-blurple) l c h / 0.8))"
+    )};
+
+  animation: rippleCycle 3000ms infinite;
+  animation-delay: ${({ $delayMs }) => `${$delayMs}ms`};
+
+  @keyframes rippleCycle {
```
jaskfla: Name it `--ripple-cycle`

The fact that CSS allows non `--`-namespaced user variable names is considered tech debt the CSS spec

(Agent might complain because its training data won’t have many `@keyframes` named with `--` prefix)

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  { index: 6, x: 8, y: 64, tone: "secondary", delayMs: 400 },
+  { index: 7, x: 36, y: 64, tone: "primary", delayMs: 560 },
+  { index: 8, x: 64, y: 64, tone: "primary", delayMs: 640 },
+] as const satisfies Square[];
+
+export interface CanvasAnimatedIconProps {
+  /**
+   * - --canvas-icon-primary: light square color
+   * - --canvas-icon-secondary: dark square color
+   */
```
jaskfla: Get rid of this icon and move `CanvasAnimatedIconStyle` closer to its usage, immediately above this

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  style?: CanvasAnimatedIconStyle;
+}
+
+export default function CanvasAnimatedIcon({ style }: CanvasAnimatedIconProps) {
```
jaskfla: Make `CanvasAnimatedIconProps` extend `React.ComponentPropsWithRef<T>` and forward props

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  return (
+    <StyledWrapper style={style}>
```
jaskfla: Redundant wrapper by the looks?

jaskfla (changes requested):
_Huge_ improvement 🎉

packages/frontend/src/components/CanvasAnimatedIcon.tsx
S-Ou: not really? that's then different timings and different scales/animations/whatever

packages/frontend/src/components/canvas/CanvasView.tsx
S-Ou: removing it breaks :(

packages/frontend/src/components/canvas/CanvasView.tsx
S-Ou: I'd still like to keep full control over being able to do two distinct colours

packages/frontend/src/components/CanvasAnimatedIcon.tsx
jaskfla: That’s what `animation-timing-function` is for, but I’m fine with leaving this quirk in

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Yeah, no, keep that; but _this specific instance_ can be achieved with `opacity: 80` and not overriding either of the two colours

Less code, same result

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Figured it had to be there for a reason 🤷

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+interface CanvasAnimatedIconProps extends React.ComponentPropsWithRef<
+  typeof StyledSvg
+> {
+  style?: CanvasAnimatedIconStyle;
+}
+
+export default function CanvasAnimatedIcon({
+  ...props
+}: CanvasAnimatedIconProps) {
```
jaskfla: Redundant clone of props object

packages/frontend/src/components/CanvasAnimatedIcon.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+`;
+
+interface Square {
+  x: number;
+  y: number;
+  tone: Tone;
+}
+
+const Squares = [
```
jaskfla:
```suggestion
const squares = [
```

jaskfla (approved):
Couple niggles left but pre-approving 🚀

PR #523, opened 2026-05-08
fix(backend): apply Discord refresh token

### LINKED ISSUE #493: Discord token invalidates over time -> can't fetch guilds for auth etc
We authenticate a user to have access to view and edit guild frames + check if they have perms to manage/moderate the canvas.

We get this info by making a request to the Discord API using a token created upon logging in with Discord. This token, however, expires after some indeterminate length of time (not sure how long)

Which means that once this token has expired, the user needs to log out and back in in order to regain all the above functionality

Rocked03:


packages/backend/src/middleware/discordAuth.ts
```diff
@@ -42,7 +42,11 @@ const discordStrategy = new DiscordStrategy(
 
       done(null, user, {
         discordAccessToken: accessToken,
+        discordRefreshToken: refreshToken,
         discordGuildFlags: userGuildFlags,
+        // Unfortunately, we don't get the expires_in value from passport-discord, so we set a temporary 1h expiry
+        // for the access token before getting refreshed after (and getting the real expiry time from the refresh token response)
+        discordTokenExpiresAt: Date.now() + 60 * 60 * 1000,
```
jaskfla, line 49: Looks like you use `3600000` everywhere else?

I personally prefer the single const, formatted as `3_600_000`

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  const response = await fetchWithRetries(
+    "https://discord.com/api/v10/oauth2/token",
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+      },
+      body: body.toString(),
```
jaskfla, line 50: Why is `body` coming from `URLSearchParams`? Weird Discord API design?

[This example URL they provide](https://docs.discord.com/developers/topics/oauth2#:~:text=https://discord.com/oauth2/authorize?response_type=code&client_id=157730590492196864&scope=identify%20guilds.join&state=15773059ghq9183habn&redirect_uri=https%3A%2F%2Fnicememe.website&prompt=consent&integration_type=0) seems to suggest it takes search params; but also the Python example code provides it as `data` so I’m not sure 🤷

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Returns a valid access token from the session, refreshing if expired or missing.
+ */
+export async function getValidAccessToken(req: Request): Promise<string> {
+  const session = req.session as DiscordRequestSession;
+
+  // If access token exists and not expired (or no expiry present), return it
+  const token = session.discordAccessToken as string | undefined;
+  const expiresAt = session.discordTokenExpiresAt as number | null | undefined;
```
jaskfla, line 97: Shouldn’t these types be inferred from `DiscordRequestSession`?

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (token && expiresAt && Date.now() < expiresAt - 30000) {
```
jaskfla, line 98:
```suggestion
  if (token && expiresAt && Date.now() < expiresAt - 30_000) {
```

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  // Otherwise try to refresh using refresh token
+  return await refreshDiscordAccessToken(req);
+}
+
+export default {
+  refreshDiscordAccessToken,
+  getValidAccessToken,
+};
```
jaskfla, line 109: Isn’t this redundant? Both are already exported; and this doesn’t change how consumers import them

I don’t know the details intimately, but I suspect this might make this module less tree-shakeable

jaskfla (approved):
Raising an eyebrow at the Discord API, but trusting you to have been through the docs in more detail and to have tested

Just optional details. Pre-approving ✅

Rocked03:
Going to redo with a different package

PR #524, opened 2026-05-09
feat: replace passport-discord + refresh token

### LINKED ISSUE #493: Discord token invalidates over time -> can't fetch guilds for auth etc

Rocked03:
- replaced `passport-discord` with `discord-strategy`
- added support for the refresh token


packages/backend/src/middleware/canvasAuth.ts
```diff
@@ -5,18 +5,25 @@ import {
...
 } from "@/services/discordGuildService";
+import { withDiscordAccessToken } from "@/services/discordTokenService";
 
 interface AuthenticatedRequest extends Request {
   user: DiscordUserProfile;
   session: Request["session"] & {
     discordAccessToken: string;
+    discordRefreshToken?: string;
+    discordTokenExpiresAt?: number;
+    discordTokenLifetimeMs?: number;
```
jaskfla, line 16: Appreciate the `ms` suffix 🙏

packages/backend/src/middleware/canvasAuth.ts
```diff
@@ -5,18 +5,25 @@ import {
...
   };
 }
 
 export function assertLoggedIn(
   req: Request,
 ): asserts req is AuthenticatedRequest {
-  if (!req.user || !req.session.discordAccessToken) {
+  if (
+    !req.user ||
+    (!req.session.discordAccessToken && !req.session.discordRefreshToken)
```
jaskfla: Completely optional:

```suggestion
    !(req.session.discordAccessToken || req.session.discordRefreshToken)
```

packages/backend/src/middleware/canvasAuth.ts
```diff
@@ -42,8 +49,10 @@ export async function requireCanvasModerator(
   try {
     assertLoggedIn(req);
 
-    const accessToken = req.session.discordAccessToken;
-    const userIsCanvasModerator = await isCanvasModerator(accessToken);
+    const userIsCanvasModerator = await withDiscordAccessToken(
+      req.session,
+      (accessToken) => isCanvasModerator(accessToken),
```
jaskfla: Redundant wrapper by the looks? I don’t think this has changed the signature at all

```suggestion
      isCanvasModerator,
```

Ditto line 79

packages/backend/src/middleware/discordAuth.ts
```diff
@@ -15,13 +17,32 @@ import { getProfilePictureUrlFromHash } from "@/services/discordProfileService";
 
 const discordStrategy = new DiscordStrategy(
   {
-    passReqToCallback: true,
     clientID: config.discord.clientId,
     clientSecret: config.discord.clientSecret,
+    authorizationURL: "https://discord.com/api/oauth2/authorize",
     callbackURL: "/api/v1/discord/callback",
-    scope: ["identify", "guilds", "guilds.members.read"],
+    tokenURL: "https://discord.com/api/oauth2/token",
```
jaskfla, line 24: Do these not need `/v10/`? 🤨

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -97,15 +98,19 @@ frameRouter.put<FrameIdParam>(
...
-        frameId,
-        bodyQueryResult.data.name,
-        x0,
-        y0,
-        x1,
-        y1,
+      const frame = await withDiscordAccessToken(
+        req.session,
+        async (accessToken) =>
+          await editFrame(
```
jaskfla, line 104: Out of scope, but this should probably take a single context object instead of 7 params

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,122 @@
...
+  session.discordAccessToken = refreshedToken.accessToken;
+
+  if (refreshedToken.refreshToken) {
+    session.discordRefreshToken = refreshedToken.refreshToken;
+  }
+
+  if (typeof session.discordTokenLifetimeMs === "number") {
+    session.discordTokenExpiresAt = Date.now() + session.discordTokenLifetimeMs;
+  } else {
+    delete session.discordTokenExpiresAt;
```
jaskfla: Is it important that this property be _absent_, or would it be okay to set it as `undefined`?

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,122 @@
...
+  accessToken: string;
+  refreshToken?: string;
+}
+
+function shouldRefreshDiscordToken(session: DiscordTokenSession): boolean {
+  if (!session.discordAccessToken) {
+    return true;
+  }
+
+  if (typeof session.discordTokenExpiresAt !== "number") {
```
jaskfla: It’s possible this is overly defensive—you make the judgement:

```suggestion
  if (typeof session.discordTokenExpiresAt !== "number" || !Number.isFinite(session.discordTokenExpiresAt) {
```

(Mostly to handle `NaN`, but `isFinite` _also_ handles `Number.POSITIVE_INFINITY`)

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,122 @@
...
+    return false;
```
jaskfla, line 29: This could be a one-liner:

```suggestion
  return typeof session.discordTokenExpiresAt !== "number"; // and potentially isNaN or isFinite check
```

jaskfla: Something to watch out for that’s come up in a couple reviews:

```ts
if (predicate) return true;
else return false;
// is just
return predicate; // or Boolean(predicate)
```

packages/backend/src/services/discordTokenService.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        },
+      );
+
+      const action = vi
+        .fn<
+          Parameters<typeof withDiscordAccessToken>[1],
+          ReturnType<typeof withDiscordAccessToken>
+        >()
+        .mockRejectedValueOnce(new UnauthorizedError("stale token"))
+        .mockResolvedValueOnce("success");
```
jaskfla: Couldn’t find these strings in `discordTokenService.ts`—does it come from Discord API?

```suggestion
        .mockRejectedValueOnce(new UnauthorizedError("stale token"))
        .mockResolvedValueOnce("success");
```

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,122 @@
...
+export async function withDiscordAccessToken<T>(
+  session: DiscordTokenSession,
+  action: (accessToken: string) => Promise<T>,
+): Promise<T> {
+  const accessToken = await getDiscordAccessToken(session);
+
+  try {
+    return await action(accessToken);
+  } catch (error) {
+    if (error instanceof UnauthorizedError && session.discordRefreshToken) {
```
jaskfla, line 118: Nice handling here

packages/backend/src/services/discordTokenService.ts
```diff
@@ -0,0 +1,122 @@
...
+      const refreshedAccessToken = await refreshDiscordAccessToken(session);
+      return await action(refreshedAccessToken);
+    }
+
+    throw error;
+  }
+}
```
jaskfla, line 125: I’m a little bit conflicted about this `withDiscordAccessToken` pattern. Something about it feels more verbose than it needs to be, but I couldn’t come up with a better alternative

Would be curious to see @pumbas600 or @PolarWolf314’s opinion if either of you have the time

jaskfla (commented):
No showstoppers, just withholding approval for second opinion

Won’t let it sit too long. If it ends up blocked for longer than we’d like I’m actually happy to stamp it

packages/backend/src/middleware/discordAuth.ts
volcanofr: We don't :)
Discord itself don't includes it within their given urls.
image

packages/backend/src/services/discordTokenService.ts
volcanofr: We could remove variables?

```suggestion
export async function withDiscordAccessToken<T>(
  session: DiscordTokenSession,
  action: (accessToken: string) => Promise<T>,
): Promise<T> {
    try {
        return await action(await getDiscordAccessToken(session));
    } catch (error) {
        if (error instanceof UnauthorizedError && session.discordRefreshToken)
            return await action(await refreshDiscordAccessToken(session));

        throw error;
    }
}
```
I personally tend to remove as much variables as possible to reduce RAM usage - but I don't know if that's a thing we aim for in that repo.

Rocked03: I think the RAM difference here is rather negligible, and we'd be trading off readability for it

packages/backend/src/routes/api/v1/frame.ts
S-Ou: will keep in mind when i get back to making some frame updates ty

packages/backend/src/services/discordTokenService.ts
S-Ou: well, not quite? there's still the line below

jaskfla (approved):
Stamping

PR #531, opened 2026-05-13
feat: silenced the unavailable database logs

Rocked03:
they were doing my head in

now it's just a single line telling the database is unavailable


packages/backend/src/errors/ApiError.ts
```diff
@@ -21,9 +22,37 @@ export default class ApiError extends Error {
...
+      return true;
+    }
+
+    if (!(error instanceof Error)) {
+      return false;
+    }
+
+    return /can't reach database server|database server is not reachable|ECONNREFUSED/i.test(
+      error.message,
+    );
```
stijnvdkolk: Wouldn't it make more sense to switch these around?
```suggestion
    if (error instanceof Error) {
      return /can't reach database server|database server is not reachable|ECONNREFUSED/i.test(
        error.message,
      );
    }

    return false;
```

stijnvdkolk (approved):

PR #533, opened 2026-05-14
fix(backend): link user and discord_user_profile together and fix profile upsert

### LINKED ISSUE #529: Discord profile insertion error
When logging-in for the first time, I receive:

Image

The main information is "Foreign key constraint violated on the constraint: `discord_user_profile_user_id_fkey`" at [backend discordProfileService.ts on createOrUpdateDiscordProfile()](https://github.com/project-blurple/Canvas-Web/blob/main/packages/backend/src/services/discordProfileService.ts#L21-L31).

stijnvdkolk:
Foreign key already existed in the database so I did not have to add that.


Rocked03 (approved):
thanks for fixing this 

lol that we didn't actually populate the user table at all

PR #534, opened 2026-05-14
feat(frontend): load tab data only when active

Rocked03:
- Stuff like frames and history etc should only load when the tab is active
- Not a huge issue right now, but it's a pattern we need to apply to the moderation dashboard blacklist etc later so it's good to put it place here now

### DIFF
--- a/packages/frontend/src/hooks/queries/useFrame.tsx
+++ b/packages/frontend/src/hooks/queries/useFrame.tsx
@@ -45,7 +45,13 @@ export function useFrameById({ frameId }: UseFrameByIdParams) {
   });
 }
 
-export function useUserFrames({ canvasId, userId }: UseUserFramesParams) {
+export function useUserFrames(
+  { canvasId, userId }: UseUserFramesParams,
+  options?: Omit<
+    UseQueryOptions<FrameRequest.UserFramesResBody>,
+    "queryKey" | "queryFn"
+  >,
+) {
   const getFrames = async (): Promise<FrameRequest.UserFramesResBody> => {
     if (!userId) return {} as FrameRequest.UserFramesResBody;
 
@@ -56,9 +62,10 @@ export function useUserFrames({ canvasId, userId }: UseUserFramesParams) {
   };
 
   return useQuery<FrameRequest.UserFramesResBody>({
+    ...options,
     queryKey: ["frame", "user", canvasId, userId],
     queryFn: getFrames,
-    enabled: Boolean(userId),
+    enabled: Boolean(userId) && (options?.enabled ?? true),
     refetchOnMount: false,
     refetchOnWindowFocus: false,
     placeholderData: {} as FrameRequest.UserFramesResBody,
--- a/packages/frontend/src/hooks/queries/useFrame.tsx
+++ b/packages/frontend/src/hooks/queries/useFrame.tsx
@@ -52,6 +52,8 @@ export function useUserFrames(
     "queryKey" | "queryFn"
   >,
 ) {
+  const { enabled = true } = options ?? {};
+
   const getFrames = async (): Promise<FrameRequest.UserFramesResBody> => {
     if (!userId) return {} as FrameRequest.UserFramesResBody;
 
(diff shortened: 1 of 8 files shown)


packages/frontend/src/hooks/queries/useFrame.tsx
```diff
@@ -56,9 +62,10 @@ export function useUserFrames({ canvasId, userId }: UseUserFramesParams) {
   };
 
   return useQuery<FrameRequest.UserFramesResBody>({
+    ...options,
     queryKey: ["frame", "user", canvasId, userId],
     queryFn: getFrames,
-    enabled: Boolean(userId),
+    enabled: Boolean(userId) && (options?.enabled ?? true),
```
jaskfla: This is correct, but slightly verbose. Perhaps:

```ts
const { enabled = true } = options;

return useQuery({ enabled: enabled && Boolean(userId) });
```

jaskfla (approved):
Just one pattern suggestion

I think we could seriously consider mounting one tab at a time, but this achieves the right result with much less code, so is the right decision for now—and maybe for a long while to come

volcanofr:
Is there an other way than using history request to know which pixel the user clicked, client-side (for userscript)?

jaskfla:
> Is there an other way than using history request to know which pixel the user clicked, client-side (for userscript)?

Can you elaborate? Unfortunately right now the currently-selected pixel is only stored in a React context, which I’d really like to refactor out into a search param; but I don’t think I’ll be able to get that done before this year’s event

Not sure if that answers your question. Do you mean which pixels a given user interacted with throughout an event? That would be possible, but we don’t track when users simply click on a pixel without actually painting it

volcanofr:
I've done a [userscript](https://blurple.volcaneau.fr/userscript.js) to have an overlay above the canvas.
When a user selected a pixel, I looked for history request params to know what pixel they clicked on and suggest them the color of the overlay.

I may have 2 way to still be able "easily" to detect position:
- css of the reticle (or custom tag attribute, but Rocked said that nothing outside of the web-usage case will be added)
- the button to place a pixel itself, if it already includes the data at all times

(everyone would be able to use my far-from-perfect overlay system)

jaskfla:
Ah, I see what you’re saying. TL;DR is that I don’t think we’ll have the capacity to “officially” cater for this, at least in time for this year’s event

We’re both working on this after-hours alongside full-time work and have already scoped out quirks that need to be ironed out for folks to have an acceptable (hopefully) experience with the web app

That said, feature-addition-wise, we’re certainly aware of the usefulness an overlay could and would provide. It’s just too substantial a feature for us to implement (or potentially even review) and have merged within the next week 🥲

Hopefully that’s not too much of a disappointment or shock; but I’m certainly glad to hear your perspective and see you work around our current constraints 😅

jaskfla:
I _suppose_ we could consider emitting a custom event which scripts or browser extensions could listen to, which _might_ be more realistic; but I’d also have to think through potential privacy implications that come with exposing that data

volcanofr:
The website is already amazing, and better every day since a month. You guys are making an **incredible** work.
I help where I can, I make things outside of the website when it is faster for everyone (won't take your time reviewing it, won't break all of my plans).

I will 100% continue contributing to it where I found myself useful.

If I really-really want to add a feature and feels like it will take too long for this year's, I'll just make a PR which clearly states to not be used within the website but by scripts.

jaskfla:
Ah, don’t get me wrong; we won’t (and _can’t_) stop you or anyone else from opening PRs. We’d very much like the contributor base for this repo to grow and for it to be more collectively owned than it is today. If you have the ideas and the inclination, please do!

If you’re volunteering your time and efforts to contribute, it’s only reasonable that we make the time to review and accept proposals that align with our vision; I just can’t promise any timelines

Despite some awkward quirks in the codebase, and a couple things we’d do differently today, we still try to keep a high standard of code quality—especially as more code becomes agent-written, which will pick up on whatever patterns/precedents we set. That likely means a slower pace of work, but this is a comprise we’re accepting (at least for the time being)

PR #545, opened 2026-05-16
fix(frontend): simplify `<NumberField>`

jaskfla:
Death to MUI ✊[^foot]

Screenshot 2026-05-16T154031

Could do with more custom styling, but let’s get a strong semantics/a11y foundation before we worry about perfecting cosmetics

[^foot]: Not actually and not everywhere all the time. Just on a case-by-case basis.


Rocked03 (approved):
ah, thanks

PR #547, opened 2026-05-16
feat(frontend): style colour autocomplete field

jaskfla:


packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
```diff
@@ -13,45 +14,25 @@ const SelectedColorChips = styled("div")`
...
-      color: color-mix(
-        in oklab,
-        contrast-color(var(--color-select-chip-color)) 94%,
-        var(--color-select-chip-color)
-      );
-    }
-  }
+const ColorPreview = styled("div")`
+  background-color: currentColor;
+  border: 1px solid oklch(from white l c h / 15%);
```
Rocked03:
```suggestion
  border: var(--card-border);
```

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
```diff
@@ -13,45 +14,25 @@ const SelectedColorChips = styled("div")`
...
+  height: 1em;
+  width: 1em;
+  border-radius: calc(infinity * 1px);
```
Rocked03: what's up with the infinity here? 

also ☝️alphasort

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
```diff
@@ -139,33 +120,38 @@ export default function ComplexSearchColorSelect({
...
-            </li>
-          );
-        }}
-        isOptionEqualToValue={(option, value) =>
-          option.id === (value as Palette[number]).id
-        }
+        renderOption={(props, option) => (
+          <ListItem {...props} key={props.key ?? option.id}>
+            <ColorPreview style={{ color: rgbaToCssColor(option.rgba) }} />
+            <Code>{option.code}</Code> {option.name}
```
Rocked03, line 142: should the `{option.name}` be wrapped in a `<span>` or something?

Rocked03 (approved):
just a couple inconsequential comments

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
jaskfla: - It’s a clearer, more intentional equivalent to `999px.` (But since this is a square, it actually renders the same as `50%` or `100%`.)
- Whoops, thanks—good catch

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
jaskfla: If we wanted to style it or needed to add a custom ARIA role, yes; but in this case it’s perfectly fine rendering as a `#text` node as a direct descendant of the `<li>` (which MUI styles as a flex box 🙄)

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
jaskfla: Whoops, yep! I had originally tried a variant of `currentColor` but it didn’t work out, so this is a remnant of that. Good eye

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
jaskfla: e281b924860405f8aa2bd097337367e5fe39f705

packages/frontend/src/components/complex-search/ComplexSearchColorSelect.tsx
jaskfla: e281b924860405f8aa2bd097337367e5fe39f705

PR #553, opened 2026-05-17
refactor(backend): enhance error handling and validation in routes

stijnvdkolk:
The error validation was done per route, same for the body, query and params validation. This PR streamlines this in a type-safe way.


jaskfla:
Whoa, neat! I’ve only had a chance to have a quick glance and like what I see

Will make the time to give it a proper review in the next day or two—thanks! 🙏

stijnvdkolk:
> Whoa, neat! I’ve only had a chance to have a quick glance and like what I see
> 
> Will make the time to give it a proper review in the next day or two—thanks! 🙏

Nice! Tried to make it as readable and maintainable as possible, but I'm sure there are things to improve haha

packages/backend/src/middleware/errorHandler.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      message: "Invalid request data",
+      errors: [{ code: "custom", path: ["name"], message: "required" }],
+    });
+  });
+
+  it("falls back to a generic 500 for unknown errors", () => {
+    const consoleErrorSpy = vi
+      .spyOn(console, "error")
+      .mockImplementation(() => {});
+    const error = new Error("boom");
```
jaskfla, line 52: 💥

packages/backend/src/models/blocklist.models.ts
```diff
@@ -1,25 +1,13 @@
...
-  const ids = Array.isArray(userId) ? userId : [userId];
-  return ids.map(BigInt);
-}
+export const BlocklistBodyModel = z
+  .object({
+    userId: z.union([snowflakeString, z.array(snowflakeString)]),
+  })
+  .transform(({ userId }) =>
+    (Array.isArray(userId) ? userId : [userId]).map(BigInt),
+  );
```
jaskfla, line 13: This seems so obvious now 😅

packages/backend/src/routes/api/v1/palette.ts
```diff
@@ -17,126 +17,80 @@ import {
...
-      const colorId = await parseColorId(req.params);
-
-      await deleteColor(colorId);
-
-      res.status(200).json({ message: "Color deleted" });
-    } catch (error) {
-      return ApiError.sendError(res, error);
-    }
+    await deleteColor(req.params.colorId);
+    res.status(200).json({ message: "Color deleted" });
```
jaskfla, line 66: Huh, curious that we don’t just 204 here (cc @Rocked03)

Doesn’t really fit this PR, though, we can clean this up separately

packages/backend/src/routes/api/v1/blocklist.test.ts
```diff
@@ -41,6 +42,7 @@ const createApp = ({ authenticated = false, moderator = false } = {}) => {
 
   app.use(setTestRequestState);
   app.use("/api/v1/blocklist", blocklistRouter);
+  app.use(errorHandler);
```
jaskfla, line 45: I forget if order matters here. Does the `blocklistRouter` need to be mounted later than `errorHandler`?

jaskfla (approved):
I’m admittedly not the strongest backend dev, but I’ve had a pretty thorough comb through and basically have no meaningful comments. No qualms regarding readability or maintainability

I’d been aware of the clumsy duplication[^foot], but just never put the time into coming up with a solution

[^foot]: Also don’t like the deep nesting that comes with `try`–`catch` blocks

Stoked with this code change. Neat stuff. Thanks so much 🥳

packages/backend/src/routes/api/v1/palette.ts
Rocked03: did i hear something? nah it must have been the wind

Rocked03 (approved):
no particularly thoughts from me, but it's very nicely done

looking forward to the ✨merge conflicts 😆

packages/backend/src/routes/api/v1/palette.ts
stijnvdkolk: Will make a quick separate PR for this later :D

packages/backend/src/routes/api/v1/blocklist.test.ts
stijnvdkolk: order matters indeed, but error handler needs to be later in the stack as it will be next after the blocklist router instead of before the blocklist router in the stack which means errors are not being caught.

PR #555, opened 2026-05-17
feat(backend): move allColorsGlobal to canvas table

### LINKED ISSUE #490: Move `ALL_COLORS_GLOBAL` from backend `.env` to database value
Either add into the `event` or `info` table or maybe even the `canvas` table?

Either way, it'd be good to be able to edit this while the app is live rather than setting it in the env

stijnvdkolk:


jaskfla (approved):
Seems reasonable!

Rocked03 (approved):
very neatly done, thank you 🙏

PR #557, opened 2026-05-17
feat(frontend): admin dashboard - canvas dashboard

### LINKED ISSUE #538: Admin Dashboard - Canvas Editor

Rocked03:
image


packages/frontend/src/components/NumberField.tsx
jaskfla, line 1: Good catch on min/max

Label should _probably_ not be optional? I’ll pull branch and look more closely

packages/frontend/src/components/canvas/CanvasPreviewCard.tsx
```diff
@@ -109,18 +135,54 @@ export function CanvasPreviewCard({
...
       </EventCanvasMeta>
       {canvasStats && (
         <EventCanvasStats>
-          <EventCanvasStat>
+          <EventCanvasStat
+            title={`${canvasStats.totalUsersInvolved.toLocaleString()} total ${canvasStats.totalUsersInvolved === 1 ? "user" : "users"} involved`}
+          >
             <Users size={16} />
-            <span>{canvasStats.totalUsersInvolved.toLocaleString()}</span>
+            <span>{`${canvasStats.totalUsersInvolved.toLocaleString()} ${canvasStats.totalUsersInvolved === 1 ? "user" : "users"}`}</span>
```
jaskfla: - `title` attribute is redundant because of the node content
- `<span>` wrapper also redundant

packages/frontend/src/hooks/queries/useCanvasInfo.tsx
```diff
@@ -20,3 +20,48 @@ export function useCanvasInfo(canvasId?: CanvasInfo["id"]) {
...
+
+  return useMutation({
+    mutationFn: (
+      data: Partial<Pick<CanvasInfo, "name" | "isLocked" | "cooldownLength">>,
+    ) => {
+      const requestUrl = `${config.apiUrl}/api/v1/canvas/${encodeURIComponent(canvasId)}`;
+
+      return axios.put(requestUrl, data, {
+        withCredentials: true,
+      });
```
jaskfla: async await

packages/frontend/src/components/CanvasIcon.tsx
jaskfla, line 1: This looks duplicated?

The move is probably to have `<CanvasIcon>` default to being static, and if given a `loading` (or `role="progressbar"`?) then enable the animation

Rocked03: Kinda largely is, but this has the properties of an icon compared to the loader we have going

Rocked03:
Todo (probably a stacked branch): loop in the canvas updates into the socket so users don't have to refresh to see changes

packages/frontend/src/app/layout.tsx
```diff
@@ -82,6 +82,7 @@ const defaultCanvasInfo = {
   eventId: 1,
   webPlacingEnabled: false,
   allColorsGlobal: false,
+  cooldownLength: 0,
```
jaskfla: Nitpick, but “duration” is probably clearer than “length”

packages/frontend/src/components/CanvasIcon.tsx
jaskfla: I reckon they could be unioned 🤔

Rocked03: I'll see what I can do

packages/frontend/src/app/layout.tsx
Rocked03: Tbh blame younger me for doing that in the SQL

I'll probably update that name everywhere aside from the schema itself

packages/frontend/src/components/NumberField.tsx
Rocked03: I'm creating the label myself in the other column of the table, so there's no need for me to have a second label on the input

packages/frontend/src/components/canvas/CanvasPreviewCard.tsx
Rocked03: mm yeah the title was in there before i added a bit more detail into the visual text

packages/backend/src/services/canvasService.ts
```diff
@@ -99,26 +110,40 @@ export function unlockedCanvasToPng(unlockedCanvas: UnlockedCanvas): PNG {
...
-      id: true,
-      name: true,
-      event_id: true,
-      locked: true,
-    },
-    where: {
-      event_id: eventId,
-    },
-  });
+  const whereSQL =
```
jaskfla: `whereClause` (or at least `whereSql`)

packages/backend/src/services/canvasService.ts
```diff
@@ -303,9 +330,15 @@ function saveCanvasToFileSystem(canvas: canvas, pixels: PixelColor[]): string {
...
-    await fs.promises.rm(cachedCanvas.canvasPath);
-    console.debug(`Cleared canvas ${canvasId} from file system`);
+  try {
+    if (cachedCanvas?.isLocked) {
+      await fs.promises.rm(cachedCanvas.canvasPath);
+      console.debug(`Cleared canvas ${canvasId} from file system`);
+    }
+  } catch {
+    console.warn(
+      `Failed to clear canvas ${canvasId} from file system. It may have already been removed.`,
```
jaskfla, line 340: Nice warning 👌

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
...
+  gap: 0.75rem;
+  overflow-x: auto;
+  overflow-y: hidden;
+  padding-bottom: 0.5rem;
+  width: 100%;
+
+  & > button {
+    flex: 0 0 10rem;
+    width: 10rem;
+  }
```
jaskfla, line 45: Don’t action any of these in this PR, but:

- Slightly awkward styling here. Here you’re asking the parent (`CanvasList`) to manage the dimensions of the children, which is what CSS Grid is good at. But trusting it works as expected, so leave it for now.
- I assume this is targeting `<StyledButton>`. Could arguably localise the styles there.

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
...
+  align-items: center;
+  display: flex;
+  font-size: 1.5rem;
+  font-weight: 600;
+  gap: 0.5rem;
+`;
+
+const Table = styled("table")`
+  border-collapse: separate;
+  border-spacing: 2rem 0.5rem;
```
jaskfla: Curious why you seem to prefer `separate`? 🤨

I almost always reach for cell padding before ~~wreaking havoc on~~ separating them

packages/frontend/src/app/admin/canvas/page.tsx
jaskfla, line 45: Would prefer logical properties here

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
   width: 100%;
 `;
 
+const CanvasList = styled("div")`
+  -webkit-overflow-scrolling: touch;
+  display: flex;
+  flex-wrap: nowrap;
+  gap: 0.75rem;
+  overflow-x: auto;
+  overflow-y: hidden;
```
jaskfla: Why would there be y-axis overflow? Does the element not grow to accommodate?

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
   width: 100%;
 `;
 
+const CanvasList = styled("div")`
```
jaskfla, line 33: This is actually a really great candidate for grid with children using subgrid; but let’s defer this

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
...
+`;
+
+const ErrorText = styled("div")`
+  color: #e04848;
+  font-size: 0.85rem;
+  margin-top: 4px;
+`;
+
+const SaveStatusText = styled("p")`
+  font-size: 0.9rem;
```
jaskfla: This font size isn’t used… anywhere? How does `0.875rem` sound to you?

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
...
+
+const createDefaults = {
+  allColorsGlobal: false,
+  cooldownDuration: 0,
+  height: 1,
+  isLocked: true,
+  name: "",
+  width: 1,
+  startCoordinates: [1, 1] as [number, number],
+};
```
jaskfla: `as const Satisfies Something`

Also makes `as [number, number]` redundant. (It was also wider than necessary.)

packages/frontend/src/app/admin/canvas/page.tsx
```diff
@@ -17,10 +30,515 @@ const CanvasInfoWrapper = styled("div")`
...
+  onSavingChange: (isSaving: boolean) => void;
+  onFormValuesChange: (values: CanvasSettingsFormProps["formValues"]) => void;
+  onSaved: (canvasId: CanvasInfo["id"]) => Promise<void>;
+}
+
+function CanvasSettingsForm({
+  activeCanvas,
+  formValues,
+  isDirty,
+  mode,
```
jaskfla, line 172: This flag feels like a code smell:

Why can’t this be derived from the canvas? Presence of an ID should imply we’re editing an existing one. Absence implies new

jaskfla (commented):
Submitting partial review

packages/backend/src/services/canvasService.ts
S-Ou: ? but it's already `whereSql`

packages/frontend/src/app/admin/canvas/page.tsx
S-Ou: nope, CanvasPreviewCard is the button

packages/frontend/src/app/admin/canvas/page.tsx
S-Ou: fine by me

packages/frontend/src/app/admin/canvas/page.tsx
S-Ou: The canvas is directly derived from the active canvas, if we derive it here, all we'd be doing is managing the mode check a bit higher in the stack

packages/backend/src/services/canvasService.ts
stijnvdkolk: I think jasper meant the casing of the sql being fully uppercase?

Rocked03: oh 🤦

stijnvdkolk (approved):
backend code ✅

PR #558, opened 2026-05-17
feat(backend): clear canvas cache

Rocked03:
In any case where the cached canvas image is incorrect/corrupt/etc, this forces a refresh

_Shouldn't_ need to be used, although when you create a canvas for the first time, it takes a little bit to generate all the pixel rows and the cached image might load before all the pixel rows are present

This PR is branched off the admin canvas dashboard since that's where i've implemented a button to trigger it


packages/frontend/src/hooks/queries/useCanvasInfo.tsx
```diff
@@ -65,3 +65,22 @@ export function useCreateCanvas() {
...
+export function useClearCanvasCache(canvasId: CanvasInfo["id"]) {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: () => {
+      const requestUrl = `${config.apiUrl}/api/v1/canvas/${encodeURIComponent(canvasId)}/cache`;
+
+      return axios.delete(requestUrl, {
+        withCredentials: true,
+      });
```
jaskfla: async–await here for that stack trace

packages/frontend/src/hooks/queries/useCanvasInfo.tsx
```diff
@@ -65,3 +65,22 @@ export function useCreateCanvas() {
...
+    },
+    onSuccess: async () => {
+      await queryClient.invalidateQueries({
+        queryKey: ["canvasInfo", canvasId],
```
jaskfla: I don’t think this corresponds to the canvas cache? From memory we don’t use TanStack query to fetch the cached canvas; I think it’s just a `useEffect`

Rocked03: We don't, it's all automated-- but we need some was to trigger a cache invalidation since it otherwise has no expiry

Rocked03: Oh wait, I misunderstood the question, will need to check

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -105,6 +106,21 @@ canvasRouter.put(
   },
 );
 
+canvasRouter.delete<CanvasIdParam>(
+  "/:canvasId/cache",
+  requireCanvasAdmin,
+  async (req, res) => {
+    try {
```
stijnvdkolk: try catch not needed anymore, error middleware takes care of it!

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -105,6 +106,21 @@ canvasRouter.put(
...
+      const canvasId = await parseCanvasId(req.params);
```
stijnvdkolk: Same for this, can be moved to validate function!

packages/backend/src/services/canvasService.ts
```diff
@@ -242,6 +242,18 @@ export async function getCanvasPng(canvasId: number): Promise<CachedCanvas> {
...
 
+/**
+ * Clears a canvas from the in-memory cache. If the canvas is locked, the cached image is also
+ * removed from the file system.
+ *
+ * @param canvasId The ID of the canvas to clear from cache
+ */
+export async function clearCachedCanvas(canvasId: number): Promise<void> {
+  await clearCanvasFromFileSystem(canvasId);
+  delete CANVAS_CACHE[canvasId];
```
stijnvdkolk: Could chance this to a map and then we can just do `CANVAS_CACHE.delete(canvasId)`, in my way nicer than this delete statement.

stijnvdkolk (approved):
Small comments, would like those to be resolved before merging, also looks like some workflow are failing

PR #562, opened 2026-05-17
feat: add support for limited partner colors and force refresh of guilds

### LINKED ISSUE #559: Restrict partner colors by server membership
Turns out we never actually implemented this, we just left the `allColorsGlobal` switch set to `true` and called it a day

https://github.com/project-blurple/Canvas-Web/blob/67b19c72685b0a7a94b9ce0dc3e67c2dc594043e/packages/backend/src/services/pixelService.ts#L74-L78

We already get their server membership via auth, we just need to check against it. It's not high stakes so we shouldn't really need to validate the auth in the backend _every_ time (ratelimits 😒), but it's probably worth instead keeping medium-length cache records that re-validates periodically

stijnvdkolk:
If invite link is present:
image

If no invite link is present:
image

Clicking on refresh shows this inline (temporary makes it scrollable):
image

Added force refresh button on me page:
image

### DIFF
--- a/packages/backend/src/routes/api/v1/discord.ts
+++ b/packages/backend/src/routes/api/v1/discord.ts
@@ -1,12 +1,16 @@
-import type { DiscordUserProfile, GuildData } from "@blurple-canvas-web/types";
+import type { DiscordUserProfile } from "@blurple-canvas-web/types";
 import { Router } from "express";
+import type { SessionData } from "express-session";
 import passport from "passport";
 
 import config from "@/config";
 import { UnauthorizedError } from "@/errors";
+import ApiError from "@/errors/ApiError";
+import { guildRefreshLimiter } from "@/middleware/ratelimit";
 import {
-  getCurrentUserGuildFlags,
+  getCachedUserGuildFlags,
   getGuildPermissionsForUser,
+  refreshCachedUserGuildFlags,
   syncDiscordGuildRecords,
 } from "@/services/discordGuildService";
 import { saveDiscordProfile } from "@/services/discordProfileService";
@@ -40,17 +44,38 @@ discordRouter.get("/guilds/permissions-map", async (req, res) => {
     throw new UnauthorizedError("User is not authenticated");
   }
 
-  const guildFlags =
-    req.session.discordGuildFlags ??
-    (await withDiscordAccessToken(req.session, (accessToken) =>
-      getCurrentUserGuildFlags(accessToken),
-    ));
+  const guildFlags = await withDiscordAccessToken(req.session, (accessToken) =>
+    getCachedUserGuildFlags(req.session, accessToken),
+  );
 
   req.session.discordGuildFlags = guildFlags;
 
   res.status(200).json({
     guilds: guildFlags,
   });
+});
+
+discordRouter.post("/guilds/refresh", guildRefreshLimiter, async (req, res) => {
+  try {
+    const profile = req.user as DiscordUserProfile;
+
+    if (!profile?.id) {
+      throw new UnauthorizedError("User is not authenticated");
+    }
+
+    const guildFlags = await withDiscordAccessToken(
+      req.session,
+      (accessToken) => refreshCachedUserGuildFlags(req.session, accessToken),
+    );
+
+    res.status(200).json({
+      guilds: guildFlags,
+    });
+
+    await syncDiscordGuildRecords(guildFlags);
+  } catch (error) {
+    ApiError.sendError(res, error);
+  }
 });
 
 /**
@@ -73,15 +98,7 @@ discordRouter.get(
   }),
   async (req, res) => {
     const discordProfile = req.user as DiscordUserProfile;
-    const authInfo = req.authInfo as
-      | {
-          discordAccessToken?: string;
-          discordRefreshToken?: string;
-          discordTokenExpiresAt?: number;
-          discordTokenLifetimeMs?: number;
-          discordGuildFlags?: Record<string, GuildData>;
-        }
-      | undefined;
+    const authInfo = req.authInfo as Partial<SessionData> | undefined;
 
     if (authInfo?.discordAccessToken) {
       req.session.discordAccessToken = authInfo.discordAccessToken;
@@ -89,6 +106,8 @@ discordRouter.get(
       req.session.discordTokenExpiresAt = authInfo.discordTokenExpiresAt;
       req.session.discordTokenLifetimeMs = authInfo.discordTokenLifetimeMs;
       req.session.discordGuildFlags = authInfo.discordGuildFlags;
+      req.session.discordGuildFlagsFetchedAt =
+        authInfo.discordGuildFlagsFetchedAt ?? Date.now();
     }
 
     res.cookie("profile", JSON.stringify(discordProfile), {
--- a/packages/backend/src/services/discordGuildService.test.ts
+++ b/packages/backend/src/services/discordGuildService.test.ts
@@ -1,0 +1,142 @@
+import type { SessionData } from "express-session";
+import fetchWithRetries from "@/utils/fetchWithRetries";
+import { getCachedUserGuildFlags } from "./discordGuildService";
+
+vi.mock("@/utils/fetchWithRetries", () => ({
+  default: vi.fn(),
+}));
+
+const mockFetch = vi.mocked(fetchWithRetries);
+
+function mockGuildsResponse(
+  guilds: Array<{
+    id: string;
+    name: string;
+    permissions?: string;
+    approximate_member_count?: number;
+  }>,
+) {
+  mockFetch.mockResolvedValueOnce({
+    status: 200,
+    ok: true,
+    headers: {
+      get: (key: string) =>
+        key === "content-type" ? "application/json; charset=utf-8" : null,
+      entries: () => [][Symbol.iterator](),
+    } as unknown as Headers,
+    json: async () => guilds,
+  } as unknown as Response);
+}
+
+function makeSession(overrides: Partial<SessionData> = {}): SessionData {
+  return { cookie: {} as SessionData["cookie"], ...overrides };
+}
+
+const sampleGuilds = [
+  {
+    id: "1",
+    name: "Guild 1",
+    permissions: "0",
+    approximate_member_count: 10,
+  },
+];
+
+describe("getCachedUserGuildFlags", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-01-01T00:00:00Z"));
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("fetches fresh flags when the session has no cache", async () => {
+    mockGuildsResponse(sampleGuilds);
+
+    const session = makeSession();
+    const result = await getCachedUserGuildFlags(session, "token");
+
+    expect(result).toMatchObject({ "1": { name: "Guild 1" } });
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    expect(session.discordGuildFlags).toEqual(result);
+    expect(session.discordGuildFlagsFetchedAt).toBe(Date.now());
+  });
+
+  it("returns cached flags within the TTL window without hitting Discord", async () => {
+    const cachedFlags = {
+      "1": {
+        name: "Cached Guild",
+        memberCount: 10,
+        administrator: false,
+        manageGuild: false,
+      },
+    };
+    const session = makeSession({
+      discordGuildFlags: cachedFlags,
+      discordGuildFlagsFetchedAt: Date.now(),
+    });
+
+    vi.advanceTimersByTime(14 * 60 * 1000);
+
+    const result = await getCachedUserGuildFlags(session, "token");
+
+    expect(result).toEqual(cachedFlags);
+    expect(mockFetch).not.toHaveBeenCalled();
+  });
+
+  it("refetches when the cached flags are older than the TTL", async () => {
+    const cachedFlags = {
+      "1": {
+        name: "Stale Guild",
+        memberCount: 10,
+        administrator: false,
+        manageGuild: false,
+      },
+    };
+    const session = makeSession({
+      discordGuildFlags: cachedFlags,
+      discordGuildFlagsFetchedAt: Date.now(),
+    });
+
+    vi.advanceTimersByTime(15 * 60 * 1000 + 1);
+
+    mockGuildsResponse([
+      {
+        id: "2",
+        name: "Refreshed Guild",
+        permissions: "0",
+        approximate_member_count: 5,
+      },
+    ]);
+
+    const result = await getCachedUserGuildFlags(session, "token");
+
+    expect(result).toMatchObject({ "2": { name: "Refreshed Guild" } });
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    expect(session.discordGuildFlags).toEqual(result);
+    expect(session.discordGuildFlagsFetchedAt).toBe(Date.now());
+  });
+
+  it("refetches when discordGuildFlagsFetchedAt is missing", async () => {
+    mockGuildsResponse(sampleGuilds);
+
+    const session = makeSession({
+      discordGuildFlags: {
+        old: {
+          name: "Old",
+          memberCount: null,
+          administrator: false,
+          manageGuild: false,
+        },
+      },
+    });
+
+    const result = await getCachedUserGuildFlags(session, "token");
+
+    expect(result).toMatchObject({ "1": { name: "Guild 1" } });
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    expect(session.discordGuildFlagsFetchedAt).toBe(Date.now());
+  });
+});
--- a/packages/backend/src/services/discordGuildService.test.ts
+++ b/packages/backend/src/services/discordGuildService.test.ts
@@ -16,16 +16,13 @@ function mockGuildsResponse(
     approximate_member_count?: number;
   }>,
 ) {
-  mockFetch.mockResolvedValueOnce({
-    status: 200,
-    ok: true,
-    headers: {
-      get: (key: string) =>
-        key === "content-type" ? "application/json; charset=utf-8" : null,
-      entries: () => [][Symbol.iterator](),
-    } as unknown as Headers,
-    json: async () => guilds,
-  } as unknown as Response);
+  mockFetch.mockResolvedValueOnce(
+    new Response(JSON.stringify(guilds), {
+      headers: {
+        "Content-Type": "application/json; charset=utf-8",
+      },
+    }),
+  );
 }
 
 function makeSession(overrides: Partial<SessionData> = {}): SessionData {
--- a/packages/backend/src/services/discordGuildService.ts
+++ b/packages/backend/src/services/discordGuildService.ts
@@ -1,4 +1,5 @@
 import type { GuildData } from "@blurple-canvas-web/types";
+import type { SessionData } from "express-session";
 import { prisma } from "@/client";
 import config from "@/config";
 import { ApiError } from "@/errors";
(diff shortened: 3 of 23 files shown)


packages/backend/src/routes/api/v1/discord.ts
```diff
@@ -83,22 +106,16 @@ discordRouter.get(
...
+    const authInfo = req.authInfo as Partial<SessionData> | undefined;
 
     if (authInfo?.discordAccessToken) {
       req.session.discordAccessToken = authInfo.discordAccessToken;
       req.session.discordRefreshToken = authInfo.discordRefreshToken;
       req.session.discordTokenExpiresAt = authInfo.discordTokenExpiresAt;
       req.session.discordTokenLifetimeMs = authInfo.discordTokenLifetimeMs;
       req.session.discordGuildFlags = authInfo.discordGuildFlags;
+      req.session.discordGuildFlagsFetchedAt =
+        authInfo.discordGuildFlagsFetchedAt ?? Date.now();
```
jaskfla, line 110: A smidge conflicted about using ms from epoch. Would normally prefer an ISO 8601 string, but considering `discordTokenExpiresAt: number`, this is the better choice

packages/backend/src/services/discordGuildService.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  mockFetch.mockResolvedValueOnce({
+    status: 200,
+    ok: true,
+    headers: {
+      get: (key: string) =>
+        key === "content-type" ? "application/json; charset=utf-8" : null,
+      entries: () => [][Symbol.iterator](),
+    } as unknown as Headers,
+    json: async () => guilds,
+  } as unknown as Response);
```
jaskfla: Will be worth sanity checking this, because I’ve just blindly written it into GitHub comment, but I think this should do the trick

```suggestion
  mockFetch.mockResolvedValueOnce(new Response(guilds, {
    headers: {
      'Content-Type': "application/json; charset=utf-8",
    },
  }))
```

(Also may fail Prettier check)

packages/backend/src/services/discordGuildService.ts
```diff
@@ -8,6 +9,8 @@ import TooManyRequestsError from "@/errors/TooManyRequestsError";
 import UnauthorizedError from "@/errors/UnauthorizedError";
 import fetchWithRetries from "@/utils/fetchWithRetries";
 
+const GUILD_FLAGS_CACHE_TTL_MS = 15 * 60 * 1000;
```
jaskfla: Optional/stylistic, but I tend to like the stricter inferred type from a `const`:

```suggestion
const GUILD_FLAGS_CACHE_TTL_MS = 900_000; // 15 min
```

packages/backend/src/services/discordGuildService.ts
```diff
@@ -219,6 +222,34 @@ export async function getCurrentUserGuildFlags(
...
+  const isFresh =
+    cached !== undefined &&
+    typeof fetchedAt === "number" &&
+    Date.now() - fetchedAt < GUILD_FLAGS_CACHE_TTL_MS;
+
+  if (isFresh) {
+    return cached;
+  }
+
+  return refreshCachedUserGuildFlags(session, accessToken);
```
jaskfla: Technically optional, but [to make stack traces easier to debug](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/errorhandling/returningpromises.md):

```suggestion
  return await refreshCachedUserGuildFlags(session, accessToken);
```

packages/backend/src/services/pixelService.ts
```diff
@@ -55,16 +55,23 @@ export async function validatePixel(
...
  *
  * @param colorId - The ID of the color
  * @param canvasId - The ID of the canvas the color is being used on
+ * @param userGuildIds - Discord guild IDs the user is a member of (as decimal strings)
  * @returns The corresponding color object
  */
 export async function validateColor(
   colorId: number,
   canvasId: number,
+  userGuildIds: ReadonlySet<string>,
```
jaskfla, line 74: I had no idea `ReadonlySet` existed!

packages/backend/src/test/mockAuth.ts
```diff
@@ -9,6 +13,30 @@ export const mockAuth = (req: Request, _res: Response, next: NextFunction) => {
       username: "test",
       profilePictureUrl: "test",
     };
+
+    const guildIdsHeader = req.header("x-TestGuildIds") ?? "";
```
jaskfla, line 17: Not a change request (because I realise that line 9 also has an `X-` header), but taking this opportunity to share that the `X-` prefix for proprietary HTTP headers is no longer advised:

- [MDN on HTTP Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers)
- [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648)

packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -0,0 +1,158 @@
...
+  border: 0;
+  color: inherit;
+  cursor: pointer;
+  font: inherit;
+  padding: 0;
+  text-decoration: underline;
+
+  &:disabled {
+    cursor: progress;
+  }
```
jaskfla: I think we can improve semantics here. I realise that in practice the button is—appropriately—only disabled when `isPending`; but I think clearer to handle the semantics separately by additionally applying `aria-busy={isPending}` and changing this selector:

```suggestion
  &:[aria-busy='true']{
    cursor: progress;
  }
```

packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -0,0 +1,158 @@
...
+`;
+
+const SUCCESS_TEXT_RESET_MS = 5_000;
+
+function getErrorText(error: unknown): string | null {
+  if (!error) return null;
+  if (error instanceof AxiosError && error.response?.status === 429) {
+    return "Slow down — try again in a minute.";
+  }
+  return "Couldn't reach Discord. Try again shortly.";
```
jaskfla: One of my hobby horses: using [educated quotes](https://practicaltypography.com/straight-and-curly-quotes.html) instead of typographic approximations

```suggestion
  return "Couldn’t reach Discord. Try again shortly.";
```

If you find that non-obvious, update the return type to `React.ReactNode` and:

```suggestion
  return <>Couldn&rsquo;t reach Discord. Try again shortly.</>;
```

packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -0,0 +1,158 @@
...
+  showSuccess: boolean;
+  errorText: string | null;
+  statusText: string | null;
+};
+
+export function useRecheckMemberships(): RecheckMembershipsController {
+  const { user } = useAuthContext();
+  const mutation = useRefreshGuildMemberships(user?.id);
+  const { isPending, isSuccess, error, reset } = mutation;
+  const [showSuccess, setShowSuccess] = useState(false);
```
jaskfla:
```suggestion
  const [didJustSucceed, setDidJustSucceed] = useState(false);
```

packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -0,0 +1,158 @@
...
+      reset();
+    }, SUCCESS_TEXT_RESET_MS);
+    return () => clearTimeout(timer);
+  }, [isSuccess, reset]);
+
+  const errorText = getErrorText(error);
+  const statusText =
+    isPending ?
+      "Checking with Discord…"
+    : (errorText ?? (showSuccess ? "Server list updated." : null));
```
jaskfla: We’re pretty consistent about using full stops only when there are multiple sentences

```suggestion
    : (errorText ?? (showSuccess ? "Server list updated" : null));
```

Same on line 106

packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -0,0 +1,158 @@
...
+    <InlineTrigger
+      type="button"
+      onClick={() => mutate()}
+      disabled={isPending}
+      aria-label="Refresh your Discord server list"
+    >
+      {children}
+    </InlineTrigger>
+  );
+}
```
jaskfla, line 132: I think this reads slightly awkwardly in the accessibility tree. Would prefer to have only content/`children` and omit the `aria-label`

I would suggest making the link text ‘Refresh here’ instead of just ‘here’ so this link (technically button, I guess) stands independently of the text around it

packages/frontend/src/components/action-panel/tabs/SelectedColorInfoCard.tsx
```diff
@@ -50,32 +59,41 @@ export default function ColorInfoCard({
...
+  const recheck = useRecheckMemberships();
 
   if (!color) return <Wrapper>No color selected</Wrapper>;
 
   const { name: colorName, code: colorCode } = color;
 
   const guildName = color.guildName ?? "a partnered server";
+  const guildNameNode =
+    invite ?
+      <a href={invite} target="_blank" rel="noreferrer">
```
jaskfla: While we’re here:

```suggestion
      <a href={invite} target="_blank" rel="external noreferrer">
```

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
```diff
@@ -7,11 +7,11 @@ const StyledSwatch = styled(StaticSwatch, { shouldForwardProp: () => true })`
   border-style: solid;
   border-width: 3px;
   transition: var(--transition-duration-fast) ease;
-  transition-property: border-color, outline-width, padding, scale;
+  transition-property: border-color, outline-width, opacity, padding, scale;
```
jaskfla: Keen eye, but intentionally omitted `opacity` here to make the GUI more responsive 🔬⚡

Otherwise when you mouse around it can look like the hover effect is lagging behind your cursor

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
```diff
@@ -26,13 +26,20 @@ const StyledSwatch = styled(StaticSwatch, { shouldForwardProp: () => true })`
...
+  &:active:not([aria-disabled="true"]) {
     scale: 97%;
   }
 
   &:disabled {
     cursor: not-allowed;
   }
+
+  /* aria-disabled keeps the swatch clickable (so the user can select it to see
+   * why it can't be placed yet) while making it visually distinct. */
```
jaskfla: Oh, nice! Love this new extra bit of communication to the user. In this case, perhaps it’s more appropriate to never apply `disabled`? Can then delete a couple styles here 🧼

The fact that `&::not(disabled)[aria-disabled='true']` is possible feels oxymoronic

jaskfla (changes requested):
Like it overall! Couple nitpicks details from UX/UI perspective, but arguably commensurate with your attentiveness to detail

Great addition, and appreciate the screenshots in the PR description 🙏

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
```diff
@@ -26,13 +26,20 @@ const StyledSwatch = styled(StaticSwatch, { shouldForwardProp: () => true })`
...
+  &[aria-disabled="true"] {
+    cursor: not-allowed;
+    opacity: 45%;
```
Rocked03: I get the reason why this is like this, but I'm feeling a bit on the fence about dimming the opacity - since Canvas is so visual, muting the colors feels like it'd then be more difficult to present to the user their potential full palette... is there some other way we can express this?

Rocked03: random idea spitball
- different coloured outline
- lock icon
- smaller inner swatch size

jaskfla: Lock icon seems reasonable. The others might be harder to pull off design-wise

stijnvdkolk: I was thinking about the lock icon too, I will play around a bit with it after work

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
stijnvdkolk: Removed it

packages/backend/src/test/mockAuth.ts
stijnvdkolk: Made separate issue

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
stijnvdkolk: image

Rocked03: how does that look on a small scale?

Rocked03: image

Rocked03: hm, i'm just worried that the icons are a little small... how would it look to blow them up a bit bigger? perhaps remove the backdrop on the icons and have them centred in the swatch?

stijnvdkolk: Normal view:
image
With dev hints on:
image

packages/backend/src/services/discordGuildService.ts
jaskfla, line 1: Ah, I presume this is why @Rocked03 was asking me about my progress on #574

packages/frontend/src/components/action-panel/tabs/SelectedColorInfoCard.tsx
```diff
@@ -50,32 +59,44 @@ export default function ColorInfoCard({
   ...props
 }: ColorInfoCardProps) {
   const { canvas } = useCanvasContext();
+  const recheck = useRecheckMemberships();
```
jaskfla: Optional, but I reckon a touch clearer that this is a `UseMutationResult` (augmented) and not a function:

```suggestion
  const recheckMutation = useRecheckMemberships();
```

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
volcanofr: I like the fact that it is bigger, but no background at all feels weak to me...
Suggestions of things to try:
- add a semi-transparent background to the lock icon
- darken effect on the color, even darker on hover
- add a chain that keeps visually the lock in place from side to side of the color selector
  image

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -261,14 +284,15 @@ function NamedPalette({ colors, disabled, name }: NamedPaletteProps) {
...
-      <Fieldset disabled={disabled}>
+      <Fieldset>
         {isLoading ?
           Array.from({ length: 12 }).map((_, i) => (
             // biome-ignore lint/suspicious/noArrayIndexKey: These will never change
             <SwatchSkeleton key={i} variant="rectangular" />
           ))
         : colors.map((color) => (
             <InteractiveSwatch
+              aria-disabled={isColorDisabled?.(color) || undefined}
```
jaskfla, line 295: I’ll flag here that I think this is inaccessible to screen reader users; but this isn’t a regression so let’s merge with it anyway

Should be a pretty small clean-up that I’m happy to take care of, but no urgency needed here

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
stijnvdkolk: Rocked doesn't want to change the swatch color (see first comment in this thread), so I rather not change anything in terms of background/color effects. I'm not so sure if having chains around the lock will fit the design style of the site, but that's up to @Rocked03 and @jaskfla to decide

volcanofr: Okay, the issue with having an effect on the color is not being able to see the real color to potentially plan things.
But I believe a chunk of a corner having the true color would be enough? 
image
Instead of darken on hover, reveal the true color completely (but keep the lock)

Rocked03: I like how it feels with this latest screenshot, I do think that adding chains would really moreso fit a different design style more than having intentional accessibility improvements.

I do get the whole "weak" feeling though, although I'm not sure if it needs to be much "stronger"... If anything I'd say it should blend a bit more with the swatch but that's pulling the needle in the opposite direction

Perhaps we can have a kind of weightiness effect on hover? via scaling and/or drop shadow perhaps

@jaskfla is the go-to person for the design style of this webapp though, so I'll defer to him for any final call

packages/frontend/src/components/swatch/StaticSwatch.tsx
```diff
@@ -26,10 +41,16 @@ export function StaticSwatch({
 
   return (
     <SwatchBase
-      style={{ ...style, backgroundColor: `rgb(${rgb} / ${alphaFloat})` }}
+      style={{
+        ...style,
+        backgroundColor: `rgb(${rgb} / ${alphaFloat})`,
+        // This ensures the swatch text color is legible on both light and dark swatches
+        color: monochromeContrastColor(rgba),
+      }}
```
jaskfla: Ah, awesome, this was the only concern I had with the icon—thanks for this

Implementation-wise, I’d prefer to delegate this to the rendering engine rather than calculating the colour on the main thread by swapping out this `styled` prop for a `styled(SwatchBase)` with these styles[^foot]:

```ts
color: contrast-color(var(--swatch-color));

@supports not (color: contrast-color(black)) {
  color: var(--swatch-color);
  filter: invert(1) grayscale(1) brightness(1.3) contrast(9000);
  mix-blend-mode: luminosity;
}
```

(Neat helper functions, though!)

[^foot]: We’ve duplicated this a couple times now. Might be worth us coming up with something like a ``` polyfilledContrastColor(color) => css`` ``` function

jaskfla (approved):
Couple optional details, one request with the fancy luminance logic, but pre-approving ✅

Thanks Stijn 🙏

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
stijnvdkolk: > The `aria-disabled` attribute, when set to `true`, indicates that the element upon which it is set and all of its focusable descendants are meant to be in the disabled state. This declaration will inform people using assistive technologies, such as screen readers, that such elements are not meant to be editable or otherwise operable.
~ https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-disabled

According to spec it should be picked up by screen readers.

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
jaskfla: I actually had to stop myself from this sort of bike-shedding in the review I just submitted 😅

I agree that we probably can come up with better or different solutions, but I think this is already in a good enough place to release. Let‘s accept this detail as-is for now, and we can quibble about small design details in their own threads/PRs

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
jaskfla: ~~I stand corrected!~~

Oh, sorry, my understanding was correct. This declares the element as non-selectable, which means screen reader users won’t get to click on the colour and see _why_ it’s locked

i.e. It’s equivalent to a sighted user interacting with a `:disabled` element

But, again, this issue applies to the current design too, so happy to merge with this

stijnvdkolk: Oh true, it's late haha, I'll leave it to you to fix in a different PR xd

packages/backend/src/services/discordGuildService.ts
Rocked03: Partly? But also I was getting ratelimited while testing my own stuff which is why I started thinking about it lol

PR #565, opened 2026-05-18
feat: canvas update socket

Rocked03:
Since we can now update the canvas properties via endpoints, this communicates updates to the front-end via the existing socket

This means users don't need to refresh their page to see the canvas lock/unlock etc


packages/backend/src/app/SocketHandler.ts
```diff
@@ -88,4 +88,9 @@ export class SocketHandler {
   ) {
     this.io.emit(`place pixel ${canvasId}`, payload);
   }
+
+  public broadcastCanvasUpdate(canvasId: number, payload: CanvasInfo) {
```
jaskfla:
```suggestion
  public broadcastCanvasUpdate(canvasId: Canvas["id"], payload: CanvasInfo) {
```

packages/frontend/src/contexts/CanvasContext.tsx
```diff
@@ -40,11 +47,43 @@ export const CanvasProvider = ({
...
+      socket.disconnect();
+    };
+  }, [mainCanvasInfo.id]);
+
+  useEffect(() => {
+    const onCanvasUpdate = (canvas: CanvasInfo) => {
+      void Promise.all([
+        queryClient.invalidateQueries({ queryKey: ["canvas"] }),
+        queryClient.invalidateQueries({ queryKey: ["canvasInfo"] }),
+      ]);
```
jaskfla: If not awaiting, then you can just fire and forget

```suggestion
      void queryClient.invalidateQueries({ queryKey: ["canvas"] }),
      void queryClient.invalidateQueries({ queryKey: ["canvasInfo"] }),
```

jaskfla (commented):
Just a 30-second glance. Can review tomorrow if no one beats me to it

packages/backend/src/app/SocketHandler.ts
```diff
@@ -83,9 +83,17 @@ export class SocketHandler {
...
     payload: PlacePixelSocket.Payload,
   ) {
     this.io.emit(`place pixel ${canvasId}`, payload);
   }
+
+  public broadcastCanvasUpdate(
+    canvasId: CanvasInfo["id"],
+    payload: CanvasInfo,
+  ) {
+    this.io.emit("canvas update", payload);
```
stijnvdkolk: Would it make sense to put this in a shared enum/type/something in the types package so it is shared and will be changed in both places if ever changed?

Rocked03: good point, i'll do that

volcanofr:
I will 100% include this on my side to stop fetching infos every 10 minutes 👀 (hum)

Rocked03:
Why are you even fetching that often, this data changes at an incredibly low frequency

volcanofr:
To catch the start of the canvas when it does start, same with stop and potential pause.
10 minutes for a request that takes 100ms at most is quite large 👀

(did my program before the announcement of the exact date)

packages/types/src/socket.ts
```diff
@@ -0,0 +1,9 @@
+export const SocketEvents = {
+  canvasUpdate: "canvas update",
+  canvasUpdateForCanvas(canvasId: number) {
```
stijnvdkolk: I feel like saying `canvasUpdateForSpecificCanvas` makes more sense

stijnvdkolk (approved):
backend code approved, I think frontend looks good too but would be nice if someone else can also take a look at the frontend code 😅

packages/types/src/socket.ts
jaskfla: Or just call it `canvasUpdate` and make the `canvasId` argument optional, returning `"canvas update"` if no ID is provided?

jaskfla:
> backend code approved, I think frontend looks good too but would be nice if someone else can also take a look at the frontend code 😅

I’m happy with the frontend changes. Presuming @Rocked03 has done some manual testing, once https://github.com/project-blurple/Canvas-Web/pull/565#discussion_r3267818512 is resolved I’m happy to see this merged

packages/types/src/socket.ts
stijnvdkolk: Also an option yea

PR #570, opened 2026-05-19
feat: history pagination

### LINKED ISSUE #434: Paginate history
Currently history is limited to 100 entries, we don't expect any pixel's history to exceed that but it'd be nice to accommodate for it

https://github.com/project-blurple/Canvas-Web/blob/main/packages/backend/src/services/pixelService.ts

volcanofr:
Adds pagination to pixel's history

<div style="display: flex;">
image
image
image
image
</div>


stijnvdkolk:
Merge conflicts due to my big revamp of validation ✨

volcanofr:
I'm trying to resolving it rn x')

volcanofr:
Feels like it was not the exact git commands to make, but it works

stijnvdkolk:
> Feels like it was not the exact git commands to make, but it works

As long as the conflicts are fixed 💪

volcanofr:
I'm curious about the `executionDurationMs` of the `Point` `{ x: 727, y: 727 }` in the 2025 canvas, first page and last page.
To see if the `OFFSET` is the right way to go for us

packages/frontend/src/components/action-panel/tabs/PixelHistoryListItem.tsx
```diff
@@ -36,13 +36,15 @@ const ColorName = styled("p")`
...
+  id,
 }: {
   record?: PixelHistoryRecord;
+  id?: string;
 }) {
   const { color, userProfile } = record ?? {};
 
   return (
-    <Wrapper>
+    <Wrapper id={id}>
```
jaskfla: Ooh, yeah, it is a bit non-intuitive that even global attributes don’t get forwarded

Worth making `interface PixelHistoryListItem extends React.ComponentPropsWithRef<typeof Wrapper> {}` here and delegating all non-`record props

packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
```diff
@@ -104,25 +154,65 @@ export default function PixelInfoTab({
...
+                justifyContent: "center",
+              },
+            }}
+            shape="rounded"
+            size="small"
+            showFirstButton
+            showLastButton
+            siblingCount={0}
+          />
+        </ActionPanelTabBody>
```
jaskfla: Now that this is kinda duplicated, perhaps worth exporting our own `<Pagination>` from the `components/` module?

packages/frontend/src/hooks/queries/usePixelHistory.tsx
```diff
@@ -5,51 +5,74 @@ import type {
...
+export interface PixelHistoryQuery {
+  point: Point;
+  page?: number;
+  size?: number;
+}
+
 export function usePixelHistory(
   canvasId: CanvasInfo["id"],
-  coordinates: Point | null,
+  query: PixelHistoryQuery | null,
```
jaskfla: Maybe `params` is a bit clearer than `query`, only because of the overlap between “query params” and `UseQueryResult`

packages/frontend/src/hooks/queries/usePixelHistory.tsx
```diff
@@ -5,51 +5,74 @@ import type {
...
     ...options,
-    queryKey: ["pixelHistory", canvasId, coordinates],
+    queryKey: ["pixelHistory", canvasId, query],
     queryFn: fetchHistory,
-    enabled: Boolean(coordinates) && (options?.enabled ?? true),
+    placeholderData: keepPreviousData,
+    enabled: Boolean(query) && (options?.enabled ?? true),
     refetchOnMount: false,
     refetchOnWindowFocus: false,
+    staleTime: 30_000, // 30 seconds
```
jaskfla, line 65: Ooh, yes, thanks for adding this

packages/frontend/src/hooks/queries/usePixelHistory.tsx
```diff
@@ -5,51 +5,74 @@ import type {
...
+        },
         signal,
       },
     );
+
     return response.data;
   };
 
-  return useQuery({
+  const queryResult = useQuery({
```
jaskfla: I suspect you did this for debugging? Fine to inline the return value imo

jaskfla (commented):
Couple smaller details, but looks solid overall 🙏

jaskfla:
Replying to screenshots in PR description (thanks for attaching those):

I think harmless to just render the `<Pagination>` below the list, not pinned to the bottom of the action panel. Bit more forgiving on shorter viewports

packages/frontend/src/hooks/queries/usePixelHistory.tsx
volcanofr: I believe I've quite copied this function from both `fetchComplexHistory` of this file and `useLeaderboard`

volcanofr:
> Replying to screenshots in PR description (thanks for attaching those):
> 
> I think harmless to just render the `<Pagination>` below the list, not pinned to the bottom of the action panel. Bit more forgiving on shorter viewports

Gonna add it to `<FullWidthScrollView>` 👍

packages/frontend/src/hooks/queries/usePixelHistory.tsx
volcanofr: Wait, so do I change what `useComplexPixelHistory` is using too, and the name of the type?

packages/frontend/src/components/action-panel/tabs/PixelHistoryListItem.tsx
volcanofr: I thought I've understood but thought about multiple ways of understanding that after a huge night of sleep
I just don't have enough knowledge about React and stuff to know what you are talking about :')

If you want to add attributes, cannot I add this?
```tsx
...
export default function PixelHistoryListItem({
  record,
  id,
  props,
}: {
  record?: PixelHistoryRecord;
  id?: string;
  props?: React.ComponentPropsWithoutRef<typeof Wrapper>;
}) {
  const { color, userProfile } = record ?? {};

  return (
    <Wrapper id={id} {...props}>
...
```

volcanofr:
<div style="display: flex;">
image
image
image
image
</div>

packages/backend/src/services/historyService.ts
```diff
@@ -152,6 +161,8 @@ async function getPixelHistoryRowsWithCount({
       : Prisma.sql`${Prisma.join(whereFragments, " AND ")}`;
   }
 
+  const take = Math.min(Math.max(size, 1), 100); // Arbitrary maximum
```
stijnvdkolk, line 164: Would it be possible to put this in a constant on top of this file?

stijnvdkolk (approved):
backend code approved with one small comment

packages/backend/src/services/historyService.ts
volcanofr: If I add a constant, will add it as a .env config, and same with frontend, and same with leaderboard
Will do in another PR if I'm not required to do it there :)

stijnvdkolk: Fine by me, as long as we do not forget haha

Rocked03:
just bumping @jaskfla

packages/frontend/src/app/leaderboard/Leaderboard.tsx
```diff
@@ -1,15 +1,8 @@
...
-import Pagination from "@mui/material/Pagination";
-import PaginationItem from "@mui/material/PaginationItem";
-import {
-  ChevronFirst,
-  ChevronLast,
-  ChevronLeft,
-  ChevronRight,
-} from "lucide-react";
 import { useId, useState } from "react";
+import StyledPagination from "@/components/Pagination";
```
jaskfla: Could probably just call it `Pagination`

```suggestion
import Pagination from "@/components/Pagination";
```

packages/frontend/src/components/Pagination.tsx
```diff
@@ -0,0 +1,40 @@
+import { Pagination, PaginationItem, styled } from "@mui/material";
+import {
+  ChevronFirst,
+  ChevronLast,
+  ChevronLeft,
+  ChevronRight,
+} from "lucide-react";
+
+const StyledPaginationItem = styled(PaginationItem)`
+  font-variant-numeric: tabular-nums;
```
jaskfla: I think this was oversight on my part:

```suggestion
  font-variant-numeric: lining-nums tabular-nums;
```

packages/frontend/src/components/Pagination.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <Pagination
+      color="primary"
+      renderItem={(item) => (
+        <StyledPaginationItem slots={customIconSlots} {...item} />
+      )}
+      sx={{
+        "& .MuiPagination-ul": {
+          justifyContent: "center",
+        },
+      }}
```
jaskfla, line 37: How come this is needed? I’d’ve thought the default would be pretty sensible 🤔

packages/frontend/src/components/Pagination.tsx
```diff
@@ -0,0 +1,40 @@
...
+const StyledPaginationItem = styled(PaginationItem)`
+  font-variant-numeric: tabular-nums;
+`;
+
+const customIconSlots = {
+  first: ChevronFirst,
+  previous: ChevronLeft,
+  next: ChevronRight,
+  last: ChevronLast,
+};
```
jaskfla:
```suggestion
} as const;
```

packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
```diff
@@ -104,21 +140,53 @@ export default function PixelInfoTab({
       })) ??
     "";
 
+  const currentId = useId();
+  const pastId = useId();
+  const listId = truePage > 1 ? pastId : `${currentId} ${pastId}`;
```
jaskfla, line 143: What are these used for? 🤨

packages/frontend/src/hooks/queries/usePixelHistory.tsx
jaskfla: I mean that you we can just do `return useQuery()` rather than `const query; return query`

packages/frontend/src/components/action-panel/tabs/PixelHistoryListItem.tsx
jaskfla: Ah, sorry; I presumed a bit too much familiarity. Here’s what I meant:

```ts
interface PixelHistoryListItemProps extends React.ComponentPropsWithRef<
  typeof Wrapper
> {
  record?: PixelHistoryRecord;
}

export default function PixelHistoryListItem({
  record,
  ...props
}: PixelHistoryListItemProps) {
  return <Wrapper {...props} />
}
```

Your solution does technically work, but it creates a slightly non-obvious contract where if you provide…

```jsx
<PixelHistoryListItem id="foo" props={{ id: 'bar' }} />
```

…then we end up with `<div id="bar" />`. It also feels slightly odd to have a `props` property in the `props` argument to a function component, especially because this component isn’t slotted

e.g. a form field which might have a `<div>` containing a `<label>` and `<input>`, we might then set up the props like:

```ts
interface FormField extends DetailedHTMLProps<
  HTMLAttributes<HTMLDivElement>,
  HTMLDivElement
> {
  slotProps: {
    input: DetailedHTMLProps<
      LabelHTMLAttributes<LabelInputElement>,
      HTMLLabelElement
    >;
    label: DetailedHTMLProps<
      InputHTMLAttributes<HTMLInputElement>,
      HTMLInputElement
    >;
  };
}
```

jaskfla (commented):
Looking solid!

- One question about `MuiPagination-ul` styling; and
- I’ve clarified what I meant about prop forwarding last time

packages/frontend/src/app/leaderboard/Leaderboard.tsx
volcanofr: Would feel duplicate from the Mui pagination component?

packages/frontend/src/components/Pagination.tsx
volcanofr: It centers the box where the buttons are, by defaults aligned to the left.
Does not changes leaderboard at all (have checked), only useful for history as of now

packages/frontend/src/components/action-panel/tabs/PixelInfoTab.tsx
volcanofr: I tried to copy what you've done on the leaderboad

packages/frontend/src/app/leaderboard/Leaderboard.tsx
volcanofr: @jaskfla, I'm not sure about using the exact same name no - but I'm open to suggestions as Styled can seems not right

Rocked03: We can just use `Pagination`, there's no issue there since we're not importing MUI's `Pagination` in this file as well

packages/frontend/src/components/Pagination.tsx
Rocked03: If we only need it for history, we should only apply it in `PixelHistoryListItem.tsx‎`

volcanofr: Isn't for items, but to align those buttons in the middle of the available width
image

Since I don't see any case with the current UI it would have to be left-aligned, let's keep as it is now - unless you have an idea of where it could be used otherwise?

Rocked03 (approved):
approving to get this merged

@jaskfla is welcome to tidy anything up at a later stage

PR #571, opened 2026-05-19
fix(backend): missing all_colors_global in seedings

volcanofr:
With the latest #555


volcanofr:
Without it I have "Argument `all_colors_global` is missing." on seeding.
With it I have "src/services/pixelService.ts:77:17 - error TS2353: Object literal may only specify known properties, and 'all_colors_global' does not exist in type 'canvasSelect<DefaultArgs>'." on building.

Only on ubuntu, pulled from commit 88613a10e62533e7ea85e612fc9c419432a823b2.

volcanofr:
Re-regenerating Prisma types fixed it. `pnpm dlx prisma generate`

PR #572, opened 2026-05-19
fix(frontend): leaderboard css

### LINKED ISSUE #530: Leaderboard css
The profile pictures `` are not centered in their `<object>`, despite having the same width/height.
Centering the image should fix the visual, is that the way to go?

Image

Also, multiple digits positions number are rendered on mutliple lines, which slows the reading spead.

volcanofr:
Fixes:
- Rank placement having a line break
- Avatar not centered

Does not break the page `/me`, that also uses Avatar.

image

### DIFF
--- a/packages/frontend/src/app/leaderboard/LeaderboardRow.tsx
+++ b/packages/frontend/src/app/leaderboard/LeaderboardRow.tsx
@@ -26,7 +26,7 @@ const TableRow = styled("li")`
 
 const Rank = styled("div")`
   color: oklch(from currentColor l c h / 45%);
-  display: grid;
+  display: block;
   font-variant-numeric: tabular-nums;
   font-weight: 900;
   grid-row: 1 / -1;
--- a/packages/frontend/src/app/leaderboard/LeaderboardRow.tsx
+++ b/packages/frontend/src/app/leaderboard/LeaderboardRow.tsx
@@ -14,7 +14,6 @@ const TableRow = styled("li")`
   display: grid;
   grid-column: 1 / -1;
   grid-template-columns: subgrid;
-  grid-template-rows: auto auto;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/app/leaderboard/LeaderboardRow.tsx
jaskfla, line 1: Oh, whoa, this is a nice catch. There’s a _lot_ of dead styles here that aren’t doing anything meaningful. Mind if I push a commit to this branch?

~~**Update:** Looks like there was a pretty horrific merge conflict resolution at some point 😓~~ Actually not sure what happened

volcanofr: Nope, feel free! You're the pro :)

jaskfla: Neat, I’ve just gotten rid of a bunch of styles from an older attempt where wanted to keep the username and pixel count as direct children of the `<li>`; but incorrectly used `grid-template-columns` instead of `grid-template-rows`

Side note: I actually removed your `display: block` because that’s the default display mode for `<div>`, but the fix should be equally valid

I’ll revisit/review the avatar style fixes this time tomorrow. Thanks for raising this PR 🙏

jaskfla (approved):
I took the liberty of also moving your avatar tweak from the `style` prop into the `styled` component (and removing a couple more redundant styles)

It actually turned out to be the `border` on the `<object>` element causing the fallback `` to be pushed to the bottom right by its `border-width`, so I have (rather infuriatingly) wrapped the entire avatar in a semantically useless `<div>` and popped the border on that

This PR was helpful for debugging, though—thanks again

PR #573, opened 2026-05-19
feat(frontend): admin dashboard - notices dashboard

### LINKED ISSUE #539: Admin Dashboard - Notice Editor

Rocked03:
Create / modify / delete notices

image


packages/backend/src/models/notice.models.ts
```diff
@@ -4,25 +4,37 @@ export const NoticeIdParamModel = z.object({
...
+    ctx.addIssue({
+      code: "custom",
+      path: ["endAt"],
+      message: "endAt must be after startAt",
+    });
+  }
+}
+
+export const CreateNoticeBodyModel =
+  NoticeBodyFieldsModel.superRefine(validateNoticeWindow);
```
stijnvdkolk: if we do the superRefine inline, then we do not have to type the types there since they will be inferred.

packages/backend/src/models/notice.models.ts
```diff
@@ -4,25 +4,37 @@ export const NoticeIdParamModel = z.object({
...
+
+export type CreateNoticeBody = z.infer<typeof CreateNoticeBodyModel>;
 
-export type CreateNoticeBody = z.infer<typeof ModifyNoticeBodyModel>;
+export const ModifyNoticeBodyModel =
+  NoticeBodyFieldsModel.partial().superRefine(validateNoticeWindow);
```
stijnvdkolk: since the update is a put request, we shouldnt make everything partial here, only the fields that can be set to null in the database

packages/backend/src/services/noticeService.ts
```diff
@@ -113,8 +145,17 @@ export async function updateNotice({
...
+  const existingTimestamps = await getNoticeTimestamps(noticeId);
+
+  const updatedStartAt =
+    startAt !== undefined ? startAt : existingTimestamps.startAt;
+  const updatedEndAt = endAt !== undefined ? endAt : existingTimestamps.endAt;
+
+  const normalizedWindow = normalizeNoticeWindow({
+    startAt: updatedStartAt,
+    endAt: updatedEndAt,
+  });
```
stijnvdkolk: See other comment about the model, since this is called in a put, we should expect all data to be overwritten, not that we re-use the old dates

stijnvdkolk (changes requested):

packages/backend/src/services/noticeService.ts
```diff
@@ -113,8 +142,17 @@ export async function updateNotice({
...
```
stijnvdkolk: We should revert this code change since now you can not clear the startAt and endAt, which you should be able to do with a put since you can initially send it in with it being optional, would be weird if you try to clear the time in the ui and then it doesn't let you I think

stijnvdkolk (changes requested):
1 last change

packages/backend/src/services/noticeService.ts
Rocked03: tbh it's now redundant, startAt and endAt would never be `undefined`, only `null`, but in doing so, they're now redundant

stijnvdkolk (approved):

PR #574, opened 2026-05-19
feat(backend): cache Discord API requests

jaskfla:
> [!NOTE]
> Had started work on this, but then momentum trailed off.
> 
> **This PR is missing unit tests and has not been comprehensively manually tested.** Cursory smoke testing in local dev environment hasn’t raised any flags, but I’m not comfortable deploying without more thorough testing.

You’ll see I’ve implemented `cache.invalidate()`, but don’t actually call it anywhere; but only because I haven’t reviewed the endpoints to see which ones warrant imperative invalidation.

## 🥞 Stack

1. #574 👈
2. #575


jaskfla:
@stijnvdkolk I understand from @Rocked03 you’re working on something related to guild memberships. We figure you might have some thoughts?

stijnvdkolk:
> @stijnvdkolk I understand from @Rocked03 you’re working on something related to guild memberships. We figure you might have some thoughts?

Yeah I did some stuff with guild membership caching in #562 so we don't always fetch the new guild memberships. Will look through this PR sometime tomorrow 🙌

Rocked03:
> > @stijnvdkolk I understand from @Rocked03 you’re working on something related to guild memberships. We figure you might have some thoughts?
> 
> Yeah I did some stuff with guild membership caching in #562 so we don't always fetch the new guild memberships. Will look through this PR sometime tomorrow 🙌

bumping :)

stijnvdkolk:
> > > @stijnvdkolk I understand from @Rocked03 you’re working on something related to guild memberships. We figure you might have some thoughts?
> > 
> > 
> > Yeah I did some stuff with guild membership caching in #562 so we don't always fetch the new guild memberships. Will look through this PR sometime tomorrow 🙌
> 
> bumping :)

Hope to be able to get to it today, got some other things that came in between yesterday!

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!this.cache.has(cacheKey)) {
+      logWithTag(
+        "Cache",
+        styleText(["bold", "yellow"], "MISS"),
+        "for",
+        styleText(["italic"], cacheKey),
+      );
+      logWithTag({ url, init, retryOptions });
+      const response = await fetchWithRetries(url, init, retryOptions);
+      return response;
```
stijnvdkolk: Cache never gets set after completion, also we should only set cache on a successful response

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,63 @@
...
+      styleText(["italic"], cacheKey),
+    );
+    const body = JSON.stringify(this.cache.get(cacheKey));
+    const options = {
+      headers: {
+        "content-type": "application/json",
+      },
+    };
+
+    return new Response(body, options);
```
stijnvdkolk: We are missing status here, might be nice to set it

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,63 @@
...
+    retryOptions?: RetryOptions,
+  ): Promise<Response> {
+    const url = new URL(input, this.baseUrl);
+
+    const method = init?.method ?? "GET";
+    if (method !== "GET") {
+      return await fetchWithRetries(url, init, retryOptions);
+    }
+
+    const cacheKey = `${url.pathname}${url.search}`;
```
stijnvdkolk: It's global, not per user.

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!this.cache.has(cacheKey)) {
+      logWithTag(
+        "Cache",
+        styleText(["bold", "yellow"], "MISS"),
+        "for",
+        styleText(["italic"], cacheKey),
+      );
+      logWithTag({ url, init, retryOptions });
```
stijnvdkolk: We log auth header here too

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,63 @@
+import { styleText } from "node:util";
+import Cache from "@/app/Cache";
+import fetchWithRetries, { type RetryOptions } from "@/utils/fetchWithRetries";
+
+function logWithTag(...args: Parameters<typeof console.debug>) {
+  console.debug(styleText(["dim"], "[DiscordApiClient]"), ...args);
+}
+
+class DiscordApiClient {
+  declare baseUrl: string;
```
stijnvdkolk, line 35: Are you sure that we need declare here?

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,63 @@
+import { styleText } from "node:util";
+import Cache from "@/app/Cache";
+import fetchWithRetries, { type RetryOptions } from "@/utils/fetchWithRetries";
+
+function logWithTag(...args: Parameters<typeof console.debug>) {
+  console.debug(styleText(["dim"], "[DiscordApiClient]"), ...args);
+}
```
stijnvdkolk, line 7: I feel like we are using a variants of this in multiple places, can we pull this out and make a full logger setup? Might be worth it to do this in a separate PR.

packages/backend/src/app/Cache.ts
```diff
@@ -0,0 +1,54 @@
+const oneMinuteInMs = 60_000;
+
+export default class Cache {
+  declare private cache: Map<string, any>;
```
stijnvdkolk: We could give a generic to Cache so we can type this instead of any.

packages/backend/src/services/discordGuildService.ts
```diff
@@ -61,7 +62,7 @@ async function discordRequest<T>({
   endpoint,
   authorization,
 }: DiscordRequestOptions): Promise<T> {
-  const response = await fetchWithRetries(
+  const response = await discordApi.fetch(
     `${DISCORD_API_BASE_URL}${endpoint}`,
```
stijnvdkolk: We still provide the full URL here even though we already set a base url in the class

packages/backend/src/app/Cache.ts
```diff
@@ -0,0 +1,54 @@
...
+  invalidate(scope: string): number {
+    let deleted = 0;
+    for (const key of this.cache.keys()) {
+      if (key.startsWith(scope)) {
+        void this.delete(key);
+        deleted++;
+      }
+    }
+    return deleted;
+  }
```
stijnvdkolk: Mutating a `Map` while iterating `keys()` is safe in JavaScript (the spec guarantees deleted keys won't appear again), but it's non-obvious. A collect-then-delete pattern is more readable:

```ts
const toDelete = [...this.cache.keys()].filter(key => key.startsWith(scope));
for (const key of toDelete) {
  this.delete(key);
}
return toDelete.length;
```

Also: `void this.delete(key)` discards the return value — that's fine, but `void` is unnecessary here (it's not a Promise).

packages/backend/src/app/Cache.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  get<T = unknown>(key: string): T | undefined {
+    return this.cache.get(key) as T | undefined;
+  }
+
+  has(key: string): boolean {
+    return this.cache.has(key);
+  }
+
+  private delete(key: string): boolean {
```
stijnvdkolk, line 43: Can we push this private method to the bottom, just so its not in between all public ones? Also might be nice to make the reset explicitly public?

stijnvdkolk (changes requested):

Rocked03:
Any chance we can give this a nudge please?

jaskfla:
> Any chance we can give this a nudge please?

Oh, are you wanting to release this mid-event? It’s a bit of a high-risk change

packages/backend/src/discord-api/DiscordApiClient.ts
jaskfla: > Cache never gets set after completion

Software is my profession 😎

stijnvdkolk: hahaha, reading from an empty cache is the best, right? 🤭

packages/backend/src/discord-api/DiscordApiClient.ts
jaskfla: Ended up refactoring to cache entire `Response` instance. Saves from having to manually clone status, headers[^foot], etc

[^foot]: Undecided if this is a good or bad thing. Probably a mix, but I’d like to avoid having to hand-pick what headers we want and don’t want to cache.

packages/backend/src/discord-api/DiscordApiClient.ts
jaskfla: Apparently so 🤷

```ts
Property 'baseUrl' does not exist on type 'DiscordApiClient'.ts(2339)
```

packages/backend/src/services/discordGuildService.ts
jaskfla: Great catch. I was also building the resultant URL incorrectly in `DiscordApiClient#fetch`, because the default base URL in there had `/api/v10` baked into it, which the `URL` constructor throws away when given as its second parameter:

```js
new URL('/foo', 'https://example.com/api/v10') → 'https://example.com/foo'
```

packages/backend/src/discord-api/DiscordApiClient.ts
stijnvdkolk: I think this is a fair choice, indeed not fully sure if saving all the headers is a good or bad thing but probably will be fine

packages/backend/src/app/Cache.ts
jaskfla: Fair! 

- Accepted with slight tweak (7192a3e559697fbec8768952c59196750766295b): I filter first, _then_ instantiate array
- I like that this gets rid of the `deleted` count
- I personally quite like `void` as an indicator that I’m intentionally throwing away a return value. (Promise or not, it doesn’t matter in this context.) Kinda like the convention of assigning to `_` in Python. That said, don’t feel strongly so removed anyway.

packages/backend/src/app/Cache.ts
jaskfla: - Pushed down in 7cfe385a3d5a3c589130bcbaf7f6c4fa66058c04
- Declared `public` in 608a475eb5d9459428550241f5eaa3b9813d9b43

packages/backend/src/app/Cache.ts
jaskfla: d7856b64468d6ffb9c04b16cd70aaf8e17422192

packages/backend/src/discord-api/DiscordApiClient.ts
jaskfla: 529d9361bca1b8a9b896a24c253a8ffc08d9d695

packages/backend/src/discord-api/DiscordApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+}
+
+function sanitizeRequestInit(requestInit: RequestInit): RequestInit {
+  const headersInit = requestInit.headers;
+  if (headersInit === undefined) return requestInit;
+  const headers = new Headers(headersInit);
+  headers.delete("Authorization");
+  return { ...requestInit, headers };
+}
```
jaskfla, line 26: Would’ve been nice to keep the whatever shape `HeadersInit` was supplied in, but:

- `Headers` interface matches case-insensitively, whereas `[string, string][]` and `{ [key: string]: string }` requires more manual work; and
- normalising for a production log feels fine to me. In local dev we can just `console.log` the entire thing with no redactions

packages/backend/src/discord-api/DiscordApiClient.ts
jaskfla: Def on board with setting up proper logging. Just realised we don’t even have Morgan (or equivalent) at the moment? 🥴

When you say pull this out, do you mean just remove all logging-related additions from this PR?

stijnvdkolk: No, just keep it like this for now, and pick it up separately, there is a separate issue for it, but we could look into a short time fix to streamline it a bit what we already have if you want by just copying these functions in a shared file that we can use in the other places too, before we switch to a better logging solution

packages/backend/src/discord-api/DiscordApiClient.ts
stijnvdkolk: Saw you made it based on token, is that smart? There is a part of the token that is the user id, cant we just only take that part and leave the rest out? Might be a bit more secure

jaskfla: > is that smart

Not really. If you snoop around the commit history you’ll see that I described it as rudimentary user scoping or something like that 😅

Realistically I’m not _that_ worried because it’s in memory in the backend. My bigger issue with this is just that it’s rather inelegant. I didn’t realise the user ID was baked into the JWT; I’ll look into that. It was my first instinct to use user ID, but I thought I had to go through 3–4 layers of dependency injection

My other gripe is that if you provide the same search params in a different order, it’s a cache miss. Originally hoped keys could just be deterministically derived from the request, but might end up putting burden on the dev to come up with sensible keys like TanStack `QueryKey`s. Not sure yet

All of which is to say: still working on it! 👷

PR #578, opened 2026-05-20
fix(frontend): locked colour accessibility

jaskfla:
Updates locked colours not to be declared as `aria-disabled` to screen readers.

Still has the same colour contrast issues for swatches that are close to 50% grey. I have some ideas, but left out of scope for this PR.

localhost_3000_

### DIFF
--- a/packages/frontend/src/components/RecheckMemberships.tsx
+++ b/packages/frontend/src/components/RecheckMemberships.tsx
@@ -3,7 +3,7 @@
 import { CircularProgress, styled } from "@mui/material";
 import { AxiosError } from "axios";
 import { type ReactNode, useEffect, useState } from "react";
-import { Button } from "@/components/button";
+import { Button, PrimitiveButton } from "@/components/button";
 import { useAuthContext } from "@/contexts";
 import { useRefreshGuildMemberships } from "@/hooks";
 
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -21,14 +21,9 @@ const StatusText = styled("p")`
...
-const InlineTrigger = styled("button")`
-  background: none;
-  border: 0;
-  color: inherit;
+const InlineTrigger = styled(PrimitiveButton)`
   cursor: pointer;
-  font: inherit;
-  padding: 0;
-  text-decoration: underline;
+  text-decoration-line: underline;
```
jaskfla, line 26: `text-decoration` shorthand overrides all of

- `text-decoration-color`,
- `text-decoration-line`,
- `text-decoration-style`, and
- `text-decoration-thickness`

but we set `text-decoration-thickness: from-font` globally, so switching back to longhand here to override only the relevant property

Note the difference between ‘Marvel Discord’ (`from-font`) and ‘Refresh here’ (`auto`) here:

Screenshot 2026-05-20T152410

Rocked03 (approved):

packages/frontend/src/components/RecheckMemberships.tsx
```diff
@@ -46,7 +41,7 @@ const SUCCESS_TEXT_RESET_MS = 5_000;
 function getErrorText(error: unknown): string | null {
   if (!error) return null;
   if (error instanceof AxiosError && error.response?.status === 429) {
-    return "Slow down — try again in a minute.";
+    return "Slow down — try again in a minute";
   }
   return "Couldn’t reach Discord. Try again shortly.";
```
stijnvdkolk, line 46: maybe also wanna remove the . here?

jaskfla: Full stops when there are multiple sentences!

PR #583, opened 2026-05-20
refactor: new frame ownership structure

### LINKED ISSUE #426: Refactor `frame` schema model to split up `owner_id`
Currently frame ownership is handled quite jankily with an `owner_id` and `is_guild_owned` but it means that it's ambiguous whether `owner_id` refers to a user or a guild.

stijnvdkolk:


packages/backend/src/models/frame.models.ts
```diff
@@ -1,6 +1,12 @@
 import z from "zod";
+import { DiscordSnowflakeSchema } from "@/utils/discordRouteUtils";
 import { CanvasIdParamModel } from "./canvas.models";
 
+export enum FrameOwnerType {
+  User = "user",
+  Guild = "guild",
+}
```
volcanofr: This feels like a close duplicate of `@types/frame.ts`, shoudn't we use the `types` package for that enum?

packages/backend/src/models/frame.models.ts
```diff
@@ -40,17 +46,25 @@ export const FrameDataParamModel = z
...
-});
+export const FrameOwnerParamModel = z.discriminatedUnion("type", [
+  z.object({
+    type: z.literal(FrameOwnerType.User),
+    userId: DiscordSnowflakeSchema,
+  }),
+  z.object({
+    type: z.literal(FrameOwnerType.Guild),
+    guildId: DiscordSnowflakeSchema,
+  }),
```
volcanofr: The `type` is already saying if that's a user or server - I suggest either using `type` & `id` or `userId`/`guildId`

volcanofr (commented):
Nice little change to the DB

packages/backend/src/models/frame.models.ts
stijnvdkolk: Fixed!

packages/backend/src/models/frame.models.ts
stijnvdkolk: Fixed!

packages/backend/prisma/migrations/20260520180000_split_frame_owner_id/migration.sql
Rocked03, line 1: is it possible to have some sort of psql constraint that necessitates that they can't both be not-null?

stijnvdkolk: Done :)

stijnvdkolk: But for "system" frames they can both be null right?

Rocked03: true... although so far the only system frames that exist are ones that i've hardcoded into the webapp (rather that being stored in the db)

Rocked03: but i suppose in theory yeah if they were in the db both would be null

stijnvdkolk: I'll remove the check again then XD

volcanofr: backend don't support non-guild and non-user at the same time anyway isn't it? Or did that changed?
If we do want to support system frames, why no creating a is_system or system_id DB column?

stijnvdkolk: Got a another idea, will push in a sec

stijnvdkolk: The backend doesn't allow you to create a system frame, but the database supports it (both user and guild id will be empty)

stijnvdkolk: Discussed with Rocked, we will keep it like this with constraint for now.

Rocked03 (approved):
looks good :) i'm glad the migration is actually more straightforward than i anticipated

PR #588, opened 2026-05-21
feat(frontend): restyle frame edit panel

### LINKED ISSUE #428: Style frame UI
- [Evaluate whether we need to use MUI](https://github.com/project-blurple/Canvas-Web/pull/419#discussion_r3107419604)
- style the frame edit menu itself
- style the sign in prompt
- style the discard changes dialog
- style the confirm delete dialog

Rocked03:


packages/frontend/src/components/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+      if (!dialog.open) {
+        dialog.showModal();
+      }
+      return;
+    }
+
+    if (dialog.open) {
+      dialog.close();
+    }
+  }, [open]);
```
jaskfla: This should use an event listener, not an Effect

packages/frontend/src/components/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+  background-color: var(--discord-legacy-not-quite-black);
+  border: var(--card-border);
+  margin: auto;
+  max-width: calc(100vw - 2rem);
+  padding: 1rem;
+  border-radius: 8px;
+
+  &::backdrop {
+    backdrop-filter: blur(4px);
+    background: oklch(0% 0% / 0.75);
```
jaskfla: 1. Prefer `background-color` longhand; we’re only setting a colour here
2. Invalid `oklch` syntax: missing a channel, and only L and A channels can be represented as a percentage. C must be a float and H uses `deg`

packages/frontend/src/components/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  &[open] {
+    display: flex;
+    flex-direction: column;
```
jaskfla: Default to normal flow layout. Make a `styled` version in the consuming module if we want to override

packages/frontend/src/components/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (clickedOutside) {
+      onRequestClose();
+    }
+  }
+
+  return (
+    <StyledDialog
+      ref={dialogRef}
+      onCancel={handleCancel}
+      onPointerDown={handlePointerDown}
```
jaskfla, line 87: What does this do? Bit lost from reading the code. Smells like it’s trying to do a normal thing in a very strange way

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -18,27 +9,54 @@ import { useCanvasContext } from "@/contexts";
...
   }
 `;
 
 const StyledButton = styled(Button)`
-  color: white;
+  background-color: var(--discord-legacy-dark-but-not-black);
+  color: var(--discord-white);
 
-  &:hover,
-  &:focus-visible {
```
jaskfla, line 33: Lost the focus style?

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -18,27 +9,54 @@ import { useCanvasContext } from "@/contexts";
...
 `;
 
-const RedButton = styled(StyledButton)`
-  &:hover,
-  &:focus-visible {
-    background-color: rgb(255, 0, 0);
+const RedStyledButton = styled(StyledButton)`
+  @media (hover: hover) and (pointer: fine) {
+    :hover {
+      background-color: rgb(255, 0, 0);
```
jaskfla: Is there no `--discord-red`?

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -18,27 +9,54 @@ import { useCanvasContext } from "@/contexts";
...
+    }
+  }
+`;
+
+const DialogButtons = styled("div")`
```
jaskfla: `ButtonRow` or `ButtonGroup`

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -18,27 +9,54 @@ import { useCanvasContext } from "@/contexts";
...
+  display: flex;
+  gap: 0.5rem;
+  justify-content: flex-end;
+  margin-top: 1rem;
```
jaskfla:
```suggestion
  margin-block-start: 1rem;
```

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -18,27 +9,54 @@ import { useCanvasContext } from "@/contexts";
...
+    :hover {
+      background-color: rgb(255, 0, 0);
+    }
+  }
+`;
+
+const DialogButtons = styled("div")`
+  display: flex;
+  gap: 0.5rem;
+  justify-content: flex-end;
```
jaskfla, line 32: If flex end, consider where the primary button lies. Is it rightmost?

If so, `flex-direction: row-reverse` so tab order reaches the primary action first

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -128,45 +146,37 @@ export default function ComplexSearchEraseHistory({
...
-                }}
-              />
-            }
-            label={`Add ${usersLength.toLocaleString()} user${usersLength !== 1 ? "s" : ""} to blocklist`}
+        <h2 id="erase-history-dialog-title">Erase history?</h2>
+        <p id="erase-history-dialog-description">
+          Delete {entriesCount.toLocaleString()} history&nbsp;
+          {entriesCount !== 1 ? "entries" : "entry"}? This cannot be undone.
+        </p>
+        <StyledLabel aria-disabled={entriesCount === 0}>
```
jaskfla: Why would a label be disabled? It’s not interactive

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -128,45 +146,37 @@ export default function ComplexSearchEraseHistory({
...
+          <input
+            type="checkbox"
+            ref={blockWhileEraseRef}
+            defaultChecked={false}
             disabled={entriesCount === 0}
+            aria-disabled={entriesCount === 0}
```
jaskfla: First rule of ARIA is to avoid it as much as possible! This doesn’t add anything in addition to `disabled`.

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -18,27 +9,54 @@ import { useCanvasContext } from "@/contexts";
...
+      background-color: var(--discord-blurple);
+    }
   }
 `;
 
-const RedButton = styled(StyledButton)`
-  &:hover,
-  &:focus-visible {
-    background-color: rgb(255, 0, 0);
+const RedStyledButton = styled(StyledButton)`
```
jaskfla: Name this semantically: `EraseButton`, `DestructiveButton`

“Red” is how it happens to be styled, not what it is

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -128,45 +146,37 @@ export default function ComplexSearchEraseHistory({
...
         Erase {entriesCount.toLocaleString()} history{" "}
         {entriesCount !== 1 ? "entries" : "entry"}
-      </RedButton>
+      </RedStyledButton>
       <StyledDialog
         open={isEraseConfirmOpen}
-        onClose={handleCancelErase}
+        onRequestClose={handleCancelErase}
         aria-labelledby="erase-history-dialog-title"
         aria-describedby="erase-history-dialog-description"
```
jaskfla: These two aren’t doing anything, because you’ve already selected perfectly appropriate roles in the `<dialog>`’s content

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
```diff
@@ -128,45 +146,37 @@ export default function ComplexSearchEraseHistory({
...
+        <h2 id="erase-history-dialog-title">Erase history?</h2>
+        <p id="erase-history-dialog-description">
+          Delete {entriesCount.toLocaleString()} history&nbsp;
+          {entriesCount !== 1 ? "entries" : "entry"}? This cannot be undone.
+        </p>
+        <StyledLabel aria-disabled={entriesCount === 0}>
+          <input
+            type="checkbox"
+            ref={blockWhileEraseRef}
+            defaultChecked={false}
```
jaskfla: Redundant, right? 🧐

packages/frontend/src/components/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
+import { styled } from "@mui/material";
+import type React from "react";
+import { useEffect, useRef } from "react";
+
+const StyledDialog = styled("dialog")`
+  background-color: var(--discord-legacy-not-quite-black);
+  border: var(--card-border);
+  margin: auto;
+  max-width: calc(100vw - 2rem);
```
jaskfla: Use logical props, and never use `vh` and `vw` alone

Prefer the `d`, `l` and `s` versions, with the plain one as fallback

```suggestion
  max-inline-size: calc(100dvi - 2rem);
```

packages/frontend/src/components/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+  padding: 1rem;
+  border-radius: 8px;
```
jaskfla: Is this `var(--card-border-radius)`? Might be, from memory

jaskfla (changes requested):
Nice direction! Handful of requests and flags, but always exciting to get to use the new baseline widely available APIs 😎

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
Rocked03: it's not a particularly aggressive red

packages/frontend/src/components/Dialog.tsx
Rocked03: native html dialogs don't come with baked-in "exit when clicking outside the dialog" behaviour

`handlePointerDown` handles that, although the bounds to calculate `clickedOutside` was because it was only getting ignored when a child component of the dialog (text etc) was blocking events but not the dialog itself

jaskfla: You mean [`popover="auto"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover)? 😉

(Though I suppose you are _technically_ correct that `<dialog>` doesn’t do it ☝️🤓)

packages/frontend/src/components/Dialog.tsx
S-Ou: no, that's 1.5rem = 24px

S-Ou: we can use that though, it looks fine

packages/frontend/src/components/complex-search/ComplexSearchEraseHistory.tsx
S-Ou: the buttons are [Cancel] and [Erase], and Cancel should be primary since Erase is the destructive one

packages/frontend/src/components/BoundsSelect/BoundsSelect.tsx
```diff
@@ -122,7 +123,9 @@ export default function ComplexSearchBoundsSelect({
           value={displayBounds?.top ?? startY}
         />
       </CoordinateInputWrapper>
-      <Scan size={24} />
+      {
+        showFrameButton && <Scan size={24} /> // This will eventually be a button that can used to select bounds via an existing frame
```
jaskfla, line 127: Remove comment

packages/frontend/src/components/button/BasicButtons.tsx
```diff
@@ -0,0 +1,30 @@
+import { styled } from "@mui/material";
+import { Button } from "./Button";
+
+export const BasicButton = styled(Button)`
+  background-color: var(--discord-legacy-dark-but-not-black);
+  color: var(--discord-white);
+
+  &:hover,
```
jaskfla, line 8: This one’s not guarded by the media query

Unfortunately it does mean duping the `border-color`

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -57,6 +45,58 @@ const EditContainer = styled("div")`
...
+  font-size: 0.875rem;
+`;
+
+const OwnerTypeOptions = styled("div")`
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0.5rem;
+`;
+
+const OwnerTypeOption = styled("label")`
```
jaskfla, line 61: Option implies `<option>` to me 🤔

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -57,6 +45,58 @@ const EditContainer = styled("div")`
...
+  align-items: center;
+  background-color: var(--discord-legacy-not-quite-black);
+  border-radius: 8px;
+  border: var(--card-border);
+  cursor: pointer;
```
jaskfla, line 66: Huh?

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -120,13 +182,16 @@ function splitGuildsByFramePresence(
   );
 }
 
+export type Mode = "create" | "edit";
+
 export default function FrameEditPanel({
   setActivePanel,
-  isCreateMode,
+  mode,
```
jaskfla, line 181: Boolean flag can probably be derived from presence of frame ID

jaskfla (approved):
Couple refactors I’d like to see/do, but discussed offline that they don’t belong in scope of this PR

PR #593, opened 2026-05-21
fix(frontend): frame info panel

Rocked03:
- titles don't overflow
- buttons hide in middle snap state on mobile
- also the slideable drawer is a little more snappy

image

https://github.com/user-attachments/assets/ed14a374-dabf-474a-944b-ff48a602fc93


packages/frontend/src/app/leaderboard/Leaderboard.tsx
```diff
@@ -73,11 +73,9 @@ export default function Leaderboard() {
           ))
         : isLeaderboardEmpty ?
           <NoContentsMessage>No leaderboard found</NoContentsMessage>
-        : entries?.map(
-            (
-              entry: any, // ToDo: re-apply types
```
jaskfla: Whoa, not sure how this got through. Thanks for cleaning up

packages/frontend/src/components/action-panel/tabs/FrameThumbCard.tsx
```diff
@@ -46,7 +46,6 @@ const MobileThumbnailCanvas = styled(PreviewCanvas)`
 const FrameTitle = styled("h3")`
   margin: 0;
   padding: 0.25rem 0.5rem;
-  white-space: nowrap;
```
jaskfla, line 49: Love it

packages/frontend/src/components/action-panel/tabs/FrameThumbCard.tsx
```diff
@@ -46,7 +46,6 @@ const MobileThumbnailCanvas = styled(PreviewCanvas)`
 const FrameTitle = styled("h3")`
   margin: 0;
   padding: 0.25rem 0.5rem;
```
jaskfla:
```suggestion
  padding-block: 0.25rem;
  padding-inline: 0.5rem;
```

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -16,10 +17,38 @@ import {
...
+  grid-template-rows: 1fr;
+  opacity: 1;
+  overflow: hidden;
+  transform: translateY(0);
+  transition-duration: 280ms;
+  transition-property: grid-template-rows, transform;
+  transition-timing-function: ease;
+
+  &[aria-hidden="true"] {
+    grid-template-rows: 0fr;
```
jaskfla, line 35: Huh, what does this do?

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -16,10 +17,38 @@ import {
...
 
+const FrameInfoPanelBodyShell = styled("div")`
+  display: grid;
+  grid-template-rows: 1fr;
+  opacity: 1;
+  overflow: hidden;
+  transform: translateY(0);
+  transition-duration: 280ms;
+  transition-property: grid-template-rows, transform;
+  transition-timing-function: ease;
```
jaskfla: This almost certainly needs a more aggressive easing function. At the very least, one of the quadratic functions defined in `core.css`, but this one might even benefit from cubic or quartic

I can never remember which direction transitions go between, say, `&` and `&[hidden]`, but the critical thing is that when animating _in_ we use an ease-out function; and when animating _out_ we use an ease-in. Basically:

1. Try `--ease-in-quad` and `--ease-out-quad`. One of them will feel natural and the other will feel broken.
2. Then tweak the suffix between `-quad`, `-cubic`, …, `-circ`. As you go to higher order functions, the transition will _appear_ faster and you will likely need to lengthen the transition duration

(`ease` is also the default so no need to explicitly declare it)

packages/frontend/src/components/header/Header.tsx
```diff
@@ -61,9 +61,7 @@ const Logo = styled(Image)(
   `,
 );
 
-export default function Header({
-  isCanvasPage: isCanvasPage = false,
-}: HeaderProps) {
+export default function Header({ isCanvasPage = false }: HeaderProps) {
   return (
     <Wrapper>
```
jaskfla, line 65: Forward `<header>` props here

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
```diff
@@ -71,6 +79,16 @@ interface SlideableDrawerProps {
   children: React.ReactNode;
 }
 
+type SlideableDrawerContextValue = {
```
jaskfla: `interface`

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
```diff
@@ -71,6 +79,16 @@ interface SlideableDrawerProps {
...
+  isMiddleSnap: boolean;
```
jaskfla: What does this mean?

packages/frontend/src/components/frames/FrameInfoPanel.tsx
Rocked03: it's the way i've managed to make it so the lower buttons completely disappear, and not just goes invisible but still take up space in the vertical stack

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
Rocked03: the slideable drawer snaps to three states: fully collapsed (bottom-most), full expanded (top-most), and half-expanded (this one)

packages/frontend/src/components/frames/FrameInfoPanel.tsx
jaskfla: So do I understand correctly that the criteria you wanna hit are:

- hides the buttons completely (visually, from keyboard access, from the a11y tree, et al); but
- can appear/disappear without layout shift?

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
jaskfla: Smells like a `boolean` is inappropriate? There are three states 🤔

More often than not, a boolean value can and should be derived from something else

packages/frontend/src/components/frames/FrameInfoPanel.tsx
Rocked03: Have a play with it on prod on mobile-- if you select a frame it takes up most of the visual space in the mid-snap. The idea is I want to completely hide it until you've pulled the draw up fully

packages/frontend/src/components/slideable-drawer/SlideableDrawer.tsx
Rocked03: suppose so? This is only really configured to be used specifically by the frame info panel right now, I suppose we can generalise it a fair bit more

jaskfla: I’ll play around in mobile and get a better sense of the code first. Don’t mint it that much; was mostly just flagging because the name was slightly opaque

Rocked03: Have actually changed it up, I'm not using a context here anymore and instead piggybacking off the same calculation that the PlacePixelTab uses

jaskfla (changes requested):
This can and probably should be done with a [container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries)

I’ll have a look tomorrow and potentially take over the PR depending on how that goes. If that falls flat, this code change seems reasonable

jaskfla (approved):
🛫

PR #596, opened 2026-05-21
fix(frontend): show login info if user is not logged in instead of refresh here check

### LINKED ISSUE #586: Partner Color, Refresh -> Log In
With that issue, we also cannot join the servers to be able to place those colors.

Testing commit version: bfe191230213c6e94564542b9e716febdb4663d1

| Deskstop (deprecated) | Mobile (deprecated) |
| --- | --- |
| Image | Image |

stijnvdkolk:


packages/frontend/src/components/action-panel/tabs/SelectedColorInfoCard.tsx
```diff
@@ -89,12 +90,16 @@ export default function ColorInfoCard({
...
+          <Subtitle>
             Exclusive to {guildNameNode}. Already joined?{" "}
             <RecheckMembershipsLink controller={recheckMutation}>
               Refresh here
             </RecheckMembershipsLink>
-            .
+          </Subtitle>
+        : <Subtitle>
+            Exclusive to {guildNameNode}. <a href="/signin">Log in</a> to verify
+            your membership
```
Rocked03:
```suggestion
            Exclusive to {guildNameNode}. <a href="/signin">Log in</a> to verify
            your membership.
```

Rocked03: Although perhaps something more like this? (this kinda hopes/assumes the user is already in the server)
```suggestion
            Exclusive to {guildNameNode}. <a href="/signin">Log in</a> to use.
```

Rocked03 (approved):
just a comment on wording but otherwise looks good

packages/frontend/src/components/action-panel/tabs/SelectedColorInfoCard.tsx
stijnvdkolk: I'm fine with both wordings, you can choose which one you think fits best 🙌

Rocked03: lets go for the rephrasing, keeps it shorter

PR #597, opened 2026-05-22
fix(backend): make frame id check case insensitive

### LINKED ISSUE #591: Frame links don't work
something broke somewhere :sob:

https://canvas.projectblurple.com/canvas/2025?f=E1FC60

not initialising frame view on load

Rocked03:
fixes #591


packages/backend/src/services/frameService.test.ts
stijnvdkolk, line 1: Why are we mocking prisma, that's not needed right? Also I think we can make it work without all those throw error's and with just expect calls

packages/backend/src/services/frameService.ts
```diff
@@ -34,9 +34,12 @@ const frameSelect = {
 } as const satisfies FrameSelect;
 
 async function findFrameForType(frameId: string) {
-  return prisma.frame.findUnique({
+  return prisma.frame.findFirst({
     where: {
-      id: frameId,
+      id: {
+        equals: frameId,
+        mode: "insensitive",
```
stijnvdkolk: Might be nice to use the prisma type here instead of raw string (`Prisma.QueryMode.insensitive`).

Also might be nice to add a gin index to this so it is performant.

packages/backend/src/services/frameService.ts
```diff
@@ -187,9 +190,12 @@ function asGuildFrame(frame: Frame): asserts frame is GuildOwnedFrame {
 }
 
 export async function getFrameById(frameId: string): Promise<Frame> {
-  const frame = await prisma.frame.findUnique({
+  const frame = await prisma.frame.findFirst({
```
stijnvdkolk, line 193: Same here

stijnvdkolk (changes requested):

stijnvdkolk (approved):

PR #599, opened 2026-05-22
feat(frontend): canvas overlay refactor

Rocked03:


packages/frontend/src/contexts/ImageOverlayContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+  width: number;
+}
+
+interface ImageOverlayContextType {
+  imageOverlay: ImageOverlayState | null;
+  showOverlay: boolean;
+  topLeftCoordinates: Point;
+  setImageOverlay: (overlay: ImageOverlayState | null) => void;
+  setShowOverlay: (show: boolean) => void;
+  setTopLeftCoordinates: (point: Point) => void;
```
jaskfla, line 19:
```suggestion
  setOrigin: (point: Point) => void;
```

packages/frontend/src/components/canvas/CanvasOverlays.tsx
```diff
@@ -0,0 +1,78 @@
...
+  } = useSelectedBoundsContext();
+
+  return (
+    <div
+      aria-hidden
+      style={{
+        inset: 0,
+        pointerEvents: "none",
+        position: "absolute",
+        zIndex: 1,
```
jaskfla, line 46: Is z-index necessary? Can the node be rendered later in the tree instead?

packages/frontend/src/app/canvas/[canvasId]/admin/paste/page.tsx
```diff
@@ -425,6 +430,37 @@ function AdminPasteTab() {
...
 
+  useEffect(() => {
+    if (!uploadedImage) {
+      setImageOverlay(null);
+      setShowOverlay(false);
+      return;
+    }
+
+    setImageOverlay({
+      alt: "Preview of uploaded image",
```
Comment (jaskfla) [inline packages/frontend/src/app/canvas/[canvasId]/admin/paste/page.tsx:441]: Can probably remove this; it’s not adding any meaning for a screen reader user

packages/frontend/src/contexts/ImageOverlayContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+interface ImageOverlayState {
+  alt: string;
+  file: File | null;
+  height: number;
+  width: number;
+}
+
+interface ImageOverlayContextType {
+  imageOverlay: ImageOverlayState | null;
+  showOverlay: boolean;
```
jaskfla, line 15: Is `showOverlay` is doing anything that can’t be derived from `imageOverlay`? Looks like `showOverlay` is always equivalent to `imageOverlay === null` by my read

packages/frontend/src/app/canvas/[canvasId]/admin/paste/page.tsx
```diff
@@ -390,6 +390,11 @@ function AdminPasteTab() {
   const [uploadedImage, setUploadedImage] = useState<UploadedImage | null>(
     null,
   );
+  const {
+    setImageOverlay,
```
Comment (jaskfla) [inline packages/frontend/src/app/canvas/[canvasId]/admin/paste/page.tsx:394]: This control feels inverted. Currently: `setImageOverlay` is being called by a child of `ImageOverlayContext`. This causes a render loop that settles, which is one technical consideration; but it’s secondary to the responsibility maybe being backward

There’s only one consumer, so I’m not quite sure (admittedly from just reading the code) that it _should_ be a context.

Tell me if this is impractical, but `imageOverlay` should be defined in the same component that renders `ImageOverlayContext` (big finger quotes around “context”), whose `key` should be tied to `imageOverlay`—and that sounds suspiciously like regular state management, not a context 🤔

jaskfla (commented):
Big question is: is context the right tool?

PR #603, opened 2026-05-23
fix(frontend): canvas picker always redirects

### LINKED ISSUE #442: When switching canvas on leaderboard page, it redirects back home

Rocked03:
Closes #442


jaskfla (approved):
Voicing a strong opposition to this approach—the boolean flag is quite a code smell—but will allow this band-aid fix as-is for now

To me this indicates that something about the redirection logic is fundamentally at odds with how the app navigation works

PR #606, opened 2026-05-23
feat(frontend): existing cooldown loads on first load

### LINKED ISSUE #577: See current cooldown on page load in frontend when logged in
Backend call was implemented in #556

Rocked03:


packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -155,9 +161,37 @@ export default function PlacePixelTab({
...
+  useEffect(
+    function initializeCooldownOnLoad() {
+      if (!active || isInitialCooldownLoaded || !initialCooldown) return;
+
+      const cooldown = initialCooldown.cooldownEndTime;
+      setCooldownSeconds(cooldown ? Math.ceil(cooldown / 1000) : 0);
+      setIsInitialCooldownLoaded(true);
+    },
+    [active, initialCooldown, isInitialCooldownLoaded],
+  );
```
jaskfla, line 180: I think these Effects are just reimplementing `isSuccess`? And then to avoid it refetching after initial request you can set `staleTime` to either `Infinity` or `'static'`

Quite confident there’s an Effectless solution to this. Also saves the extra `useState` added

packages/frontend/src/hooks/queries/useCanvasCooldown.tsx
```diff
@@ -0,0 +1,25 @@
+import type { Cooldown } from "@blurple-canvas-web/types";
+import { useQuery } from "@tanstack/react-query";
+import axios from "axios";
+import config from "@/config/clientConfig";
+
+export function useCanvasCooldown(canvasId: number, enabled: boolean) {
```
jaskfla: We have a pretty strong convention of having the second arg be `UseQueryOptions<T>`; probably worth sticking to that, but happy to see this change deferred

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
Rocked03: messed with it a bit and i couldn't fully get rid of the effect

S-Ou (approved):
discussed off-platform, approving for now and can revisit later

PR #607, opened 2026-05-23
fix: blank pixel overlay

### LINKED ISSUE #425: Placing a blank pixel blends with the pixel below
The blank pixel has unique behaviour, where it is blurple with translucency. However, this means that when the blank pixel is placed on the canvas, instead of visually completely covering the pixel below it, you can still see the previous colour through the translucency of the blank pixel.

This bug only occurs in your current session, as restarting the page refetches a fresh view of the canvas with all the up-to-date colours.

Rocked03:
Placing blank pixels now no longer stacks on top of the colour below!!!!!!

closes #425

### DIFF
--- a/packages/backend/src/services/pixelService.ts
+++ b/packages/backend/src/services/pixelService.ts
@@ -1,5 +1,4 @@
 import type {
-  Palette,
   PaletteColor,
   PixelColor,
   Point,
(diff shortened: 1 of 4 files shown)


jaskfla (approved):
I actually think the simple solution would be to make new pixels opaque, rather than masking what’s below and putting a translucent pixel over the top

i.e. color-mix the new pixel with Not Quite Black, or even simulate the behaviour by literally placing a Not Quite Black pixel and the new-coloured pixel in the same co-ordinate

That should let us get away with reusing existing code

…but I’m also fine enough with this

PR #608, opened 2026-05-23
feat(backend): export canvas and frame as png

### LINKED ISSUE #398: Export whole canvas / frames as images
Image

Rocked03:
- canvas now caches as a png in 1x, 2x, and 4x sizes
- image generation now handled by `sharp` instead of `pngjs`
- dynamic frame routes that crop out a section of the canvas


packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -65,6 +68,17 @@ canvasRouter.get(
   },
 );
 
+canvasRouter.get(
+  "/:canvasId/:size.png",
+  validate({ params: CanvasExportParamModel }),
+  async (req, res) => {
+    const size = req.params.size as CanvasExportSize;
```
stijnvdkolk: The `as` can be removed if you use my zod model change

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -143,23 +157,30 @@ function sendCachedCanvas(
   res: Response,
   canvasId: number,
   cachedCanvas: CachedCanvas,
+  size: CanvasExportSize = 1,
```
stijnvdkolk: Can use `DEFAULT_CANVAS_EXPORT_SIZE` here

packages/types/src/models/canvas.models.ts
```diff
@@ -33,3 +35,13 @@ export const CanvasPasteBodyModel = z.object({
...
 });
+
+export const CanvasExportParamModel = z.object({
+  canvasId: CanvasIdParamModel.shape.canvasId,
+  size: z.coerce
+    .number()
+    .int()
+    .refine((size) => CANVAS_EXPORT_SIZES.includes(size as CanvasExportSize), {
+      message: `size must be one of: ${CANVAS_EXPORT_SIZES.join(", ")}`,
+    }),
```
stijnvdkolk:
```suggestion
  size: z.literal(CANVAS_EXPORT_SIZES),
```

packages/types/src/models/frame.models.ts
```diff
@@ -78,3 +80,13 @@ export const FrameGuildIdsQueryModel = z.object({
...
 });
+
+export const ExportFrameParamModel = z.object({
+  frameId: FrameIdParamModel.shape.frameId,
+  size: z.coerce
+    .number()
+    .int()
+    .refine((size) => CANVAS_EXPORT_SIZES.includes(size as CanvasExportSize), {
+      message: `size must be one of: ${CANVAS_EXPORT_SIZES.join(", ")}`,
+    }),
```
stijnvdkolk:
```suggestion
  size: z.literal(CANVAS_EXPORT_SIZES),
```

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -35,6 +38,36 @@ frameRouter.get(
   },
 );
 
+frameRouter.get(
+  "/:frameId/:size.png",
+  validate({ params: ExportFrameParamModel }),
+  async (req, res) => {
+    const frame = await getFrameById(req.params.frameId);
+    const size = req.params.size as CanvasExportSize;
```
stijnvdkolk: The `as` can be removed if you use my zod model change

packages/backend/src/services/frameService.ts
```diff
@@ -455,3 +464,124 @@ export async function assertMaxOwnerFramesNotExceeded({
...
+        .extract({
+          left: cropX,
+          top: cropY,
+          width: cropWidth,
+          height: cropHeight,
+        })
+        .png();
+}
+
+function pixelsToRgbaBuffer(
```
stijnvdkolk: Also exists in `canvasService.ts`, should be moved to a shared util

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -35,6 +38,36 @@ frameRouter.get(
...
+        frameId: req.params.frameId,
+        size,
+      });
+      stream.on("error", (err) => {
+        console.error("Error streaming frame PNG:", err);
+        if (!res.headersSent) res.sendStatus(500);
+      });
+      stream.pipe(res);
+    } catch (err) {
+      console.error("Failed to export frame stream:", err);
```
stijnvdkolk: This swallows errors from the exportFrameAsStream function, we should be able to just remove the try catch

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -35,6 +38,36 @@ frameRouter.get(
   },
 );
 
+frameRouter.get(
+  "/:frameId/:size.png",
+  validate({ params: ExportFrameParamModel }),
+  async (req, res) => {
+    const frame = await getFrameById(req.params.frameId);
```
stijnvdkolk: We request frame twice, here and in the exportFrameAsStream function, this value is only used for the frame id content disposition header, we can just use the params here

packages/backend/src/services/frameService.ts
```diff
@@ -455,3 +464,124 @@ export async function assertMaxOwnerFramesNotExceeded({
...
+          height: unlocked.height * size,
+          kernel: sharp.kernel.nearest,
+        })
+        .extract({
+          left: cropX,
+          top: cropY,
+          width: cropWidth,
+          height: cropHeight,
+        })
+        .png();
```
stijnvdkolk: When size > 1, this resizes the whole canvas to (up to) 4x before extracting the (typically much smaller) frame region. Since the kernel is `nearest`, you'll get an identical result by extracting first and resizing the crop:

```suggestion
  return source
    .extract({ left: x0, top: y0, width, height })
    .resize({ width: width * size, height: height * size, kernel: sharp.kernel.nearest })
    .png();
```

packages/backend/src/services/canvasService.ts
```diff
@@ -227,8 +296,27 @@ export async function getCurrentCanvas(): Promise<[number, CachedCanvas]> {
...
+    return cachedCanvas;
+  }
+
+  const canvasPath = getLockedCanvasPath(cachedCanvas.canvasPaths, size);
+
+  if (!canvasPath) {
+    throw new Error(
+      `There is no cached canvas file for canvas ${canvasId} at ${size}x`,
+    );
+  }
```
stijnvdkolk: A canvas cache is already created in the method above on line 303 right? This includes any missing versions so the updated code for this function could look like this:
```
export async function getCanvasPng(canvasId: number): Promise<CachedCanvas> {
  return getOrFetchCacheCanvas(canvasId);
}
```

packages/backend/src/services/canvasService.ts
```diff
@@ -344,54 +466,109 @@ async function clearCanvasFromFileSystem(canvasId: number): Promise<void> {
...
-  // response.
-  return unlockedCanvas;
+export function getLockedCanvasPath(
+  canvasPaths: Partial<Record<CanvasExportSize, string>>,
+  size: CanvasExportSize,
+): string | undefined {
+  return (
+    canvasPaths[size] ?? canvasPaths[1] ?? canvasPaths[2] ?? canvasPaths[4]
+  );
 }
```
stijnvdkolk, line 502:
```suggestion
export function getLockedCanvasPath(
  canvasPaths: Partial<Record<CanvasExportSize, string>>,
  size: CanvasExportSize,
): string | undefined {
  return canvasPaths[size];
}
```

packages/backend/src/services/canvasService.ts
```diff
@@ -44,25 +50,49 @@ export type CachedCanvas = LockedCanvas | UnlockedCanvas;
...
+  const lockedCanvasPaths = new Map<
+    number,
+    Partial<Record<CanvasExportSize, string>>
+  >();
+
   for (const filename of fs.readdirSync(config.paths.canvases)) {
-    const match = filename.match(/^blurple-canvas__(\d+)__locked.png$/);
+    const match = filename.match(
+      /^blurple-canvas__(\d+)__locked(?:__(\d)x)?\.png$/,
+    );
```
stijnvdkolk, line 69:
```suggestion
  const match = filename.match(
    /^blurple-canvas__(\d+)__locked(?:__(\d+)x)?\.png$/,
  );
```

So we can also handle multi digit sizes (if ever wanted)

packages/types/src/canvasExport.ts
```diff
@@ -0,0 +1,7 @@
+export type CanvasExportSize = 1 | 2 | 4;
+
+export const CANVAS_EXPORT_SIZES: readonly CanvasExportSize[] = [1, 2, 4];
+
+export const DEFAULT_CANVAS_EXPORT_SIZE: CanvasExportSize = 1;
+
+export default CanvasExportSize;
```
stijnvdkolk: Is this default export needed?

stijnvdkolk (changes requested):

packages/backend/src/services/canvasService.ts
jaskfla: I’d suggest using `@2x` suffix convention over `__2x`

packages/backend/src/routes/api/v1/canvas.ts
jaskfla: Call it either **scale factor** or **density**; **size** really suggests something like `800w` and not `2x`

packages/frontend/src/hooks/useCanvasImage.ts
```diff
@@ -9,7 +9,7 @@ export function useCanvasImage(canvasId: number): HTMLImageElement | null {
     const image = new Image();
     image.decoding = "async";
     image.crossOrigin = "anonymous";
-    image.src = `${config.apiUrl}/api/v1/canvas/${canvasId}`;
+    image.src = `${config.apiUrl}/api/v1/canvas/${canvasId}/1.png`;
```
jaskfla: I missed where the corresponding backend code is, but:

```suggestion
    image.src = `${config.apiUrl}/api/v1/canvas/${canvasId}@1x.png`;
```

packages/types/src/canvasExport.ts
Rocked03: only really for the existing canvas png endpoint which i've left untouched, but we could fully just hardcode that

S-Ou: oh, i saw the wrong line for this comment

stijnvdkolk (approved):
Looks good!

PR #609, opened 2026-05-23
feat: move models to types package to have shared types between frontend and backend

stijnvdkolk:
I noticed that we were putting things down in multiple places, now we have them shared between backend and frontend so they should never deviate again!


Rocked03 (approved):
a lotta stuff but it all seems alright to me

packages/frontend/src/app/admin/notice/EditableNotice.tsx
Rocked03, line 1: what's up with the toISOStrings?

stijnvdkolk: The mutation expects them to be send as iso strings. Before they were automatically changed to iso strings before sending, with the current changes we have to set them here already, maybe we can do some updating in the mutations to make sure it happens behind the scenes again

Rocked03: ah I see

I don't really mind either way

PR #634, opened 2026-05-24
feat(backend): move canvas stats and event stats summary to views

### LINKED ISSUE #601: Use views for history stats
#538 creates a bunch of history stats, we should migrate these to be calculated via views instead

stijnvdkolk:
If I missed any, let me know! Then I'll see if I can convert those too.


Rocked03 (approved):
thanks, views are way tidier this way

PR #635, opened 2026-05-24
feat: opengraph embed

Rocked03:
image

this should hopefully translate well into Discord

### DIFF
--- a/packages/backend/src/routes/api/v1/canvas.ts
+++ b/packages/backend/src/routes/api/v1/canvas.ts
@@ -1,5 +1,6 @@
 import {
   CanvasExportParamModel,
+  CanvasExportQueryModel,
   type CanvasExportScale,
   CanvasIdParamModel,
   CanvasPasteBodyModel,
@@ -9,8 +10,8 @@ import {
   type DiscordUserProfile,
   EditCanvasBodyModel,
 } from "@blurple-canvas-web/types";
-
 import { type Response, Router } from "express";
+import type z from "zod";
 import { UnauthorizedError } from "@/errors";
 import { requireCanvasAdmin } from "@/middleware/canvasAuth";
 import { typedRouter } from "@/middleware/typedRouter";
@@ -31,6 +32,7 @@ import {
   pasteCanvasData,
   unlockedCanvasToPngStream,
 } from "@/services/canvasService";
+import { exportCanvasBoundsAsStream } from "@/services/exportService";
 import { getUserCanvasCooldown } from "@/services/pixelService";
 import { pixelRouter } from "./pixel";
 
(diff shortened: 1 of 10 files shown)


packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -187,12 +195,13 @@ canvasRouter.delete(
...
  * Handles sending a cached canvas as a response.
  */
-function sendCachedCanvas(
+async function sendCachedCanvas(
   res: Response,
   canvasId: number,
   cachedCanvas: CachedCanvas,
   scale: CanvasExportScale = DEFAULT_CANVAS_EXPORT_SCALE,
-): void {
+  bounds?: z.infer<typeof OptionalFrameBoundsModel>,
```
stijnvdkolk: Might be cleaner to do the zod infer in the types package

packages/frontend/src/app/canvas/[canvasId]/page.tsx
```diff
@@ -1,6 +1,127 @@
...
+
+    return toMetadata({
+      title: `Blurple Canvas | ${canvasInfo.name} (${x}, ${y})`,
+      description: `(${x}, ${y}) in ${canvasInfo.name}`,
+      imageUrl,
+    });
+  }
+
+  const scale = calculateScale(canvasInfo.width * canvasInfo.height);
+  const imageUrl = `${config.apiUrl}/api/v1/canvas/${encodeURIComponent(canvasInfo.id)}@${scale}x.png`;
```
Comment (stijnvdkolk) [inline packages/frontend/src/app/canvas/[canvasId]/page.tsx]: I've seen the `${config.apiUrl}/api/v1/canvas/${encodeURIComponent(canvasInfo.id)}` written a few times now, might make sense to store parts in a shared const so we do not repeat ourselves, same for the frame url.

packages/frontend/src/app/canvas/[canvasId]/page.tsx
```diff
@@ -1,6 +1,127 @@
...
+}
+
+export async function generateMetadata({
+  params,
+  searchParams,
+}: {
+  params: { canvasId?: string };
+  searchParams: Promise<NextSearchParams>;
+}): Promise<Metadata> {
+  const canvasId = params?.canvasId ? Number(params.canvasId) : undefined;
```
Comment (stijnvdkolk) [inline packages/frontend/src/app/canvas/[canvasId]/page.tsx]: What do we do if the canvas here is not a number and the Number parse fails?

packages/types/src/models/frame.models.ts
```diff
@@ -14,6 +14,14 @@ const FrameBoundsModel = z.object({
   y1: z.coerce.number().int().positive(),
 });
 
+export const OptionalFrameBoundsModel = z.union([
+  z
+    .object({})
+    .strict()
+    .transform(() => undefined),
```
stijnvdkolk: This feels very weird, didn't we want to use the .optional() from zod here?

packages/frontend/src/app/canvas/[canvasId]/page.tsx
Comment (Rocked03) [reply packages/frontend/src/app/canvas/[canvasId]/page.tsx]: 🤔

Comment (stijnvdkolk) [reply packages/frontend/src/app/canvas/[canvasId]/page.tsx]: https://canvas.projectblurple.com/canvas/abc as example

packages/types/src/models/frame.models.ts
Rocked03: i tried but it didn't work out-- it just complained that all four values were NaN

packages/types/src/models/frame.models.ts
```diff
@@ -14,13 +14,27 @@ const FrameBoundsModel = z.object({
...
-    .object({})
-    .strict()
-    .transform(() => undefined),
-  FrameBoundsModel,
-]);
+export const OptionalFrameBoundsModel = FrameBoundsModel.partial()
+  .superRefine((data, ctx) => {
+    const vals = [data.x0, data.y0, data.x1, data.y1];
+    const definedCount = vals.filter((v) => v !== undefined).length;
+    if (definedCount > 0 && definedCount < 4) {
```
Rocked03, line 21: this ain't gonna cause issues if you do something like `x1=5&x1=6` will it?

stijnvdkolk: That returns an error since its changed into an array [5,6], and the coerce.number().int() turns that into NaN, but that will not cause any issues with the definedCount

stijnvdkolk (approved):

PR #636, opened 2026-05-24
fix(backend): update CORS configuration to allow localhost for development

### LINKED ISSUE #243: CORS preventing localhost requests
CORS is currently preventing requests from **localhost** on our deployed version. It would be good if we could get this to work as it would allow people to test the frontend by changing the API URL, without needing to start/setup the backend.

Note that the CORS configuration for the WebSockets is separate, but I believe it's even more restrictive so it'll be good to make that publically available too.

stijnvdkolk:


packages/backend/src/app/index.ts
```diff
@@ -19,9 +19,9 @@ export function createApp(): App {
   const app = express();
 
   const corsOptions = {
-    origin: config.frontendUrl,
+    origin: [config.frontendUrl, "http://localhost:3000"],
```
volcanofr, line 22: RegExp should be working according to [config options](https://www.npmjs.com/package/cors#user-content-configuration-options)[^foot]
```suggestion
    origin: [config.frontendUrl, /^https?:\/\/localhost:\d+/],
```

[^foot]: Update by @jaskfla: Also applies to line 40

packages/backend/src/app/index.ts
```diff
@@ -37,7 +37,7 @@ export function createApp(): App {
   const server = createServer(app);
   const io = new Server(server, {
     cors: {
-      origin: config.frontendUrl,
+      origin: [config.frontendUrl, "http://localhost:3000"],
```
volcanofr, line 40: By checking the [docs](https://socket.io/docs/v4/server-options/#cors), same change should work
```suggestion
      origin: [config.frontendUrl, /^https?:\/\/localhost:\d+/],
```

packages/backend/src/app/index.ts
jaskfla: I reckon just set the env var rather than accepting any port. I’d keep the original

jaskfla (approved):
`config.frontendUrl` should arguably be set; but happy to have a sensible default

Might be overly defensive, but could add the localhost URL only if NODE_ENV isn’t production

Rocked03:
> add the localhost URL only if NODE_ENV isn’t production

I think the point of this PR is so that you could locally host the front-end but hook into the hosted back-end

volcanofr:
Can we get a public reason why this was abandoned?

jaskfla:
> Can we get a public reason why this was abandoned?

Yeah, not unreasonable. Really, it’s just that this PR came up as a devX convenience, and it wasn’t worth the security trade-off

PR #637, opened 2026-05-24
fix(backend): remove double socket handler update broadcast

stijnvdkolk:


Rocked03 (approved):
lol how'd i manage that

PR #638, opened 2026-05-24
feat(frontend): frame download button

### LINKED ISSUE #398: Export whole canvas / frames as images

Rocked03:


packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -44,6 +45,17 @@ const FrameInfoPanelBodyShell = styled("div")`
...
 
+const ButtonWrapper = styled("div")`
+  background: transparent;
+  display: flex;
+  gap: 0.5rem;
+  padding: 0;
+
+  > * {
+    flex: 1;
+  }
```
jaskfla, line 54: Looks like you actually want a grid

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -44,6 +45,17 @@ const FrameInfoPanelBodyShell = styled("div")`
   }
 `;
 
+const ButtonWrapper = styled("div")`
+  background: transparent;
```
jaskfla, line 47: This is default

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -85,6 +97,13 @@ export default function FrameInfoPanel({
   );
 }
 
+function calculateScale(pixelCount: number): number {
+  // TODO: this is duped from another PR, should dedupe once merged
```
jaskfla: You can also define them in the same part of the same file and merge resolution should take care of that

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -108,6 +127,38 @@ function FrameInfoPanelBody({
...
 
   if (selectedFrame) {
+    const downloadLink = (() => {
+      const scale = calculateScale(
+        (selectedFrame.x1 - selectedFrame.x0 + 1) *
+          (selectedFrame.y1 - selectedFrame.y0 + 1),
+      );
+
+      if (selectedFrame.owner.type === FrameOwnerType.System) {
+        const baseUrl = `${config.apiUrl}/api/v1/canvas/${canvas.id}@${scale}.png`;
```
jaskfla: URI-encode this

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -108,6 +127,38 @@ function FrameInfoPanelBody({
...
+
+        const isWholeCanvas =
+          selectedFrame.x0 === 0 &&
+          selectedFrame.y0 === 0 &&
+          selectedFrame.x1 === canvas.width - 1 &&
+          selectedFrame.y1 === canvas.height - 1;
```
jaskfla: Edge case: What happens if coordinates go _beyond_ canvas bounds?

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -121,17 +172,28 @@ function FrameInfoPanelBody({
...
+                }}
+              >
+                Copy frame link
+              </TooltipDynamicButton>
+            )}
+            <a href={downloadLink} target="_blank" rel="noopener noreferrer">
+              <DynamicButton
+                color={color}
+                disabled={!frameUrl}
+                style={{ width: "100%" }}
```
jaskfla: Prefer logical props

jaskfla (commented):
Seems neat 📥

packages/frontend/src/components/frames/FrameInfoPanel.tsx
Rocked03: I tried that but there can be a case where you don't have a frame link to share (system frames) and grid still made the single button only take half the space

packages/frontend/src/components/frames/FrameInfoPanel.tsx
Rocked03: Handled by the backend, but it shouldn't happen because selected bounds can't extend past the canvas bounds

packages/frontend/src/components/frames/FrameInfoPanel.tsx
Rocked03: i think i was using it in a completely new file

jaskfla: Git can handle that!

packages/frontend/src/components/frames/FrameInfoPanel.tsx
jaskfla: Sure, let’s leave as-is. Marking as resolved

packages/frontend/src/components/frames/FrameInfoPanel.tsx
Rocked03: 🤷

packages/frontend/src/components/frames/FrameInfoPanel.tsx
jaskfla: Any chance you misinterpreted my question? The second half of your answer addresses it, but `isWholeCanvas` will incorrectly evaluate to false if `selectedFrame` for _some_ reason extends beyond bounds

```suggestion
        const isWholeCanvas =
          selectedFrame.x0 <= 0 &&
          selectedFrame.y0 <= 0 &&
          selectedFrame.x1 >= canvas.width - 1 &&
          selectedFrame.y1 >= canvas.height - 1;
```

Minor because in general it’s nice not to be overly defensive, but noting that
- this code runs in browser, where user could theoretically alter it; and
- this is such a trivial change.

Rocked03: Yeah, I'm fine with that

What I meant was that if you passed in those larger bounds through to the backend, it'd just normalise them to be the canvas bounds when generating the crop anyways (so would just not crop it)

packages/frontend/src/components/frames/FrameInfoPanel.tsx
S-Ou: nope

the ActionPanelTabBody parent gives all children
```css
  > * {
    background-color: var(--discord-legacy-dark-but-not-black);
```

packages/frontend/src/components/frames/FrameInfoPanel.tsx
```diff
@@ -121,17 +165,28 @@ function FrameInfoPanelBody({
...
-              onAction={() => {
-                navigator.clipboard.writeText(frameUrl);
-              }}
-            >
-              Copy frame link
-            </TooltipDynamicButton>
-          )}
+          <ButtonWrapper>
+            {selectedFrame.owner.type !== FrameOwnerType.System && (
+              <TooltipDynamicButton
```
Rocked03: toasts pr removes tooltips in favour of toasts

Rocked03:
Going to fold this into #776

Rocked03:
Replaced by #793

PR #641, opened 2026-05-24
feat(canvas): implement arrow key navigation and focus handling

### LINKED ISSUE #623: Being able to move your selected pixel around with arrow keys
It would be cool if you could use your arrow keys to move the pixel you have selected

stijnvdkolk:
- Added keyboard navigation for the canvas using arrow keys, allowing users to move the selection.
- Introduced focus styles for the canvas wrapper to enhance accessibility.
- Updated state management to handle offset and zoom references during navigation.

Pressing SHIFT + <arrow key> allows you to travel faster in a direction (10 pixels at a time)


Rocked03:
Just had a play and I like it! Not sure though if I like it fully snapping when the reticle goes off the edge though... 

This is how I kinda imagined it would work:
- if the reticle is currently on the screen
  - and is within the accepted threshold and leaves it
    - pan the viewport only enough to get the reticle visually back to where it was
  - and is not within the threshold already
    - pan the viewport the minimum distance to get it back within the threshold
- if the reticle is not currently on the screen
  - pan to recentre the reticle in the screen
  
might be a bit overkill, dunno

stijnvdkolk:
> Just had a play and I like it! Not sure though if I like it fully snapping when the reticle goes off the edge though...
> 
> This is how I kinda imagined it would work:
> 
> * if the reticle is currently on the screen
>   
>   * and is within the accepted threshold and leaves it
>     
>     * pan the viewport only enough to get the reticle visually back to where it was
>   * and is not within the threshold already
>     
>     * pan the viewport the minimum distance to get it back within the threshold
> * if the reticle is not currently on the screen
>   
>   * pan to recentre the reticle in the screen
> 
> might be a bit overkill, dunno


https://github.com/user-attachments/assets/2ee06a69-357e-478e-aab3-ae26e7ee6707

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -87,6 +88,10 @@ const CanvasWrapper = styled("div")`
   & {
     user-select: none;
   }
+
+  &:focus {
```
jaskfla:
```suggestion
  &:focus-visible {
```

(In almost all cases were we habitually reach for `:focus`, [`:focus-visible` is probably the more appropriate selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:focus-visible#focus_vs_focus-visible))

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -433,6 +438,15 @@ const RETICLE_SIZE = RETICLE_ORIGINAL_SIZE * 10;
 const RETICLE_SCALE = 1 / (RETICLE_ORIGINAL_SCALE * 10);
 const PREVIEW_PIXEL_SIZE = 0.8 * RETICLE_ORIGINAL_SCALE * 10;
 
+const ARROW_KEY_DELTAS: Record<string, Point> = {
+  ArrowLeft: { x: -1, y: 0 },
+  ArrowRight: { x: 1, y: 0 },
+  ArrowUp: { x: 0, y: -1 },
+  ArrowDown: { x: 0, y: 1 },
+};
```
jaskfla:
```suggestion
const ARROW_KEY_DELTAS = {
  ArrowLeft: { x: -1, y: 0 },
  ArrowRight: { x: 1, y: 0 },
  ArrowUp: { x: 0, y: -1 },
  ArrowDown: { x: 0, y: 1 },
} as const satisfies Record<string, Point>;
```

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -433,6 +438,15 @@ const RETICLE_SIZE = RETICLE_ORIGINAL_SIZE * 10;
...
+const ARROW_KEY_SHIFT_STEP_SIZE = 10;
```
jaskfla: Hmmm, “shift” is an implementation detail. Perhaps “jump” or “skip”

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -433,6 +438,15 @@ const RETICLE_SIZE = RETICLE_ORIGINAL_SIZE * 10;
...
+const ARROW_KEY_AUTO_PAN_PADDING = 80;
```
jaskfla: I’m fine with this, but worth considering that [scroll margin](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scroll-margin) already exists as a concept and is named so

It is also a property of the pixel, not the arrow key

packages/frontend/src/components/canvas/autoPan.ts
jaskfla, line 1: Appreciate the context object as arguments here, versus a dozen arguments 🙏

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -87,6 +88,10 @@ const CanvasWrapper = styled("div")`
   & {
     user-select: none;
   }
+
+  &:focus {
+    outline: none;
```
jaskfla: Excuse me if I’m reading the code incorrectly (haven’t pulled code yet sorry), but I think we _should_ show when focus is in a place where arrow key navigation is available

Something like

```css
&:is(:focus-visible, :focus-within) {
  outline: var(--focus-outline);
}
```

jaskfla (commented):
Broadly happy with it! I haven’t had a chance to run the code locally yet to get a _feel_ for the keyboard navigation, so will do a follow-up review from an interaction design perspective[^foot]

[^foot]: …or an approval if I have no more to say!

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1225,6 +1241,62 @@ export default function CanvasView({
...
+        newCoords,
+        offset,
+        container: { width: containerWidth, height: containerHeight },
+        canvas: { width: canvas.width, height: canvas.height },
+        zoom,
+        padding,
+      });
+
+      if (nextOffset.x !== offset.x || nextOffset.y !== offset.y) {
+        setOffset(clampOffset(nextOffset, zoom));
```
Rocked03: wondering if we can hook this into `handlePan()`? because currently it's visually very snappy, whereas the click and drag to pan has some smooth velocity to it that I think `handlePan` calculates

volcanofr:
While moving with arrow keys or something that frequently switch pixel, it should not insta-request pixel's history while in the Look tab
Otherwise looks very nice!

Rocked03:
> While moving with arrow keys or something that frequently switch pixel, it should not insta-request pixel's history while in the Look tab Otherwise looks very nice!

No, we don't need to do that-- the pixel history should only load when you're on the Look tab, so if you're not interested in making history requests, then you don't go onto that tab.

If we're concerned about making a lot of requests, we should make sure we can cancel outdated history requests mid-flight

packages/frontend/src/components/canvas/CanvasView.tsx
stijnvdkolk: Made some changes! Feel free to try again :)

packages/frontend/src/components/canvas/autoPan.ts
```diff
@@ -0,0 +1,121 @@
...
+  zoom: number;
+}): number {
+  return containerSize / 2 + offset + (coord + 0.5 - canvasSize / 2) * zoom;
+}
+
+/**
+ * Calculates the new viewport offset for one axis when the reticle moves from
+ * `oldCoord` to `newCoord`, following these rules:
+ *
+ * 1. If the reticle was off-screen on this axis, fully recenter on it.
```
Rocked03: Looks good! Although we might need to either apply some lenience to this or have the snap back animation a little stiffer because I'm finding if I leave the arrow key pressed, it sometimes manages to escape far enough to trigger the recentre

packages/frontend/src/components/canvas/autoPan.ts
```diff
@@ -0,0 +1,121 @@
...
+ * 2. If the reticle was inside the padded threshold and stays inside, do not pan.
+ * 3. If the reticle was inside the threshold and leaves it, pan by the movement delta
```
Rocked03, line 54: Got an odd bug. Can only repro in Chromium-- it's partly a known existing bug.

When you zoom, the canvas blurs until the zoom settles. We gotta fix this eventually but it's not necessarily in the scope of this PR, but it's interacting weirdly with the autopan

- If the last mouse interaction was selecting a pixel, there's no issues
- If the last mouse interaction was a zoom, then every autopan inherits the same blur bug

https://github.com/user-attachments/assets/c640f12a-a900-4d01-a3b4-4a736292fab3

stijnvdkolk: Made a fix for part of it I think

packages/frontend/src/components/canvas/autoPan.ts
stijnvdkolk: Made some changes!

Rocked03 (approved):
brain isn't fully mathsing today but the whole thing feels good and the code looks good too so I'm happy with it 👍

packages/frontend/src/components/canvas/useCanvasMomentum.ts
jaskfla, line 1: This is a neat! I would request that we guard it with a [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion) check, though

If `window.matchMedia("(prefers-reduced-motion: reduce)")`, then skip the animation and just snap to the new spot

jaskfla (commented):
Just one a11y concern. Will let you decide whether to fold that into this PR or do it separately

Issue #600, opened 2026-05-23
Auto-apply migrations when deploying


Rocked03:
Currently we manually run `pnpm prisma:migrate` on the prod database before deploying prod, this should be baked into the actual deployment

PR #643, opened 2026-05-24
feat: add soft-lock support

### LINKED ISSUE #418: Canvas soft-lock
To combat griefing, the canvas could "soft-lock", where placing is only limited to people with history already on the canvas. Useful for the last few hours to allow for unobstructed cleanup

Would require another column in the `canvas` database for this flag / modifying the existing `locked` column from bool

stijnvdkolk:


packages/frontend/src/components/action-panel/tabs/place/PlacePixelButton.tsx
Rocked03: can we dedupe the `<DynamicButton {...props} disabled>`

Rocked03 (commented):
Looks good! I'm wondering if there's any better name we could use for this instead of "soft-lock" though (I don't mind it but it was certainly just the first thing that came to mind)

(cc @jaskfla @pumbas600 if you have ideas?)

packages/frontend/src/components/canvas/CanvasPreviewCard.tsx
```diff
@@ -139,7 +142,9 @@ export function CanvasPreviewCard({
         <EventCanvasNameWrapper>
           <EventCanvasName>{canvas.name}</EventCanvasName>
           {currentEventId === canvas.eventId && <Radio size={16} />}
-          {canvas.isLocked && <Lock size={16} />}
+          {[CanvasPlaceState.NoOne, CanvasPlaceState.NoNewUsers].includes(
+            canvas.placeState,
```
Rocked03: separate icons?

packages/types/src/index.ts
Rocked03, line 1: why are we `.js` files now?

Rocked03 (approved):
Pre-approving because I think ultimately it's good

but it's also a generally more fundamental change and it'd be good if somebody else (@jaskfla?) could have a quick peek too. 

It'd be good to roll this out before the end of the event so we can use it, but we've got to be careful about when and how we do the migration.... #600?

in the meanwhile we shouldn't rush to merge this one

packages/frontend/src/components/canvas/CanvasPreviewCard.tsx
stijnvdkolk: Or we just only show the lock when in no_new_users and you are a new user? And no lock for users that already have history?

packages/types/src/index.ts
stijnvdkolk: Something went wrong with the seed script since that now uses the types package too... Did some research online and apparently this was the fix (since esm stuff or something), but looks like it still doesn't work

Rocked03:
just bumping @jaskfla icymi -- this would be a "nice to have" feature to rollout before the end of the event

packages/frontend/src/components/canvas/CanvasPreviewCard.tsx
Rocked03: it's only seen by admins anyways, but it should be more representative of the actual canvas state

.github/workflows/ci.yml
jaskfla, line 1: What’s the rationale here? 🤨

packages/backend/src/services/pixelService.test.ts
```diff
@@ -1,5 +1,5 @@
 import { fail } from "node:assert";
-import type { PaletteColor } from "@blurple-canvas-web/types";
+import { CanvasPlaceState, type PaletteColor } from "@blurple-canvas-web/types";
```
jaskfla, line 2: I’ve previously pushed pretty hard to keep the types package _strictly_ types-only, going as far as enforcing that it has only declaration files. As a result, we’ve preferred union types over enums, which @pumbas600 and consider to be a bit of a harmful thing in TypeScript[^foot]

I see now that the packages uses Zod, and presumably has some actual JS code emitted from `tsc` that gets run at build-time

This is now the second time it’s come up (previously was _just_ for the addition of an `enum`, which we reverted), so perhaps I need to reconsider. But I may need some convincing 😅

[^foot]: This case where it’s literally just primitive strings is sort of the one usage of `enum`s I’d really condone, but it feels more honest just to define a `const`. They’re otherwise an oddball data structure that somehow works both as a `type` and as an `Object`, which kind of requires some understanding of what the TypeScript compiler does that feels unnecessary to me

packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx
```diff
@@ -355,14 +356,19 @@ function CanvasSettingsForm({
...
+                size="small"
+                value={formValues.placeState}
+                onChange={handlePlaceStateChange}
+              >
+                <MenuItem value={CanvasPlaceState.NoOne}>No one</MenuItem>
+                <MenuItem value={CanvasPlaceState.NoNewUsers}>
+                  No new users
+                </MenuItem>
+                <MenuItem value={CanvasPlaceState.Anyone}>Anyone</MenuItem>
+              </Select>
```
Comment (jaskfla) [inline packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx:377]: Can leave out of this PR, but flagging for (probably) myself to replace either with native `<select>` or `<NativeSelect>` later down the line

jaskfla (commented):
Happy with the db changes, thanks for that 🙏

Have some reservations about the `@blurple-canvas-web/types` package emitting actual, non-declaration JS code, which I’ve left a comment about

packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx
Comment (Rocked03) [reply packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx:377]: iirc that canvas edit page needs a refactor away from MUI eventually but its hardly urgent since i'm the only one who uses it

Thing is, native HTML doesn't have switches

Comment (jaskfla) [reply packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx:377]: Genuine response: Checkbox or radio can achieve the same semantics as a switch

Unhelpful response: Safari renders a switch with `<input type="checkbox" switch />` 😎

Comment (stijnvdkolk) [reply packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx:377]: Let's leave this for now and we can get back to it once we refactor the dashboard

.github/workflows/ci.yml
stijnvdkolk: We need the types built, but there is no need to build the types for the types package (then we do it twice :p)

packages/backend/src/services/pixelService.test.ts
stijnvdkolk: Maybe we do need to swap some stuff into another package, since I think its nice to have the zod request and response models shared between the backend and frontend packages 😅

Happy to discuss this further though!

.github/workflows/ci.yml
jaskfla: Oh, sorry, I meant why build it separately?

The `matrix` spins up different runners anyway. The move might be to update the build step to

```sh
pnpm --filter ${{ matrix.package }}... run build
```

to build the thing and its dependencies

packages/backend/src/services/pixelService.test.ts
jaskfla: Inclined to agree that schemas fit well in this package. Might be an issue of naming

Happy to go ahead with this. (But still no `enum`s!)

jaskfla (approved):
CI steps still feel a _little_ clumsy, so put in a suggestion. I’ve not used this feature of pnpm before—only Yarn—but it should work as expected

PR #647, opened 2026-05-24
fix(frontend): fix formatTimestamp double UTC offset conversion

### LINKED ISSUE #640: /me page shows recently placed timestamp in UTC rather than local time

stijnvdkolk:


Rocked03 (approved):
ah... thanks

will need to steal that since there's a few other spots that are doing some funky timezone stuff

PR #649, opened 2026-05-24
fix(backend): prevent concurrent token refresh race with in-flight deduplication

stijnvdkolk:


Rocked03 (approved):
oh neat this should help with some ratelimiting surely

stijnvdkolk:
Yeah i hope so 🙏

PR #650, opened 2026-05-24
fix(backend): simplify notice active logic and enforce startAt required when endAt set

stijnvdkolk:


Rocked03 (approved):
lol how did i never think of just doing `if !startup return false`

PR #652, opened 2026-05-24
refactor(frontend): streamline cooldown management

stijnvdkolk:
This makes sure the existing cooldown is re-used like when resizing the browser from small to big (side-panel vs bottom panel), but also when collapsing the fullscreen side-panel and or switching between any of these 3 styles while a cooldown is active.

I'm sure we can do some more optimizing of this code, so if you got any feedback, please share!

### DIFF
--- a/packages/frontend/src/contexts/ActionPanelContext.tsx
+++ b/packages/frontend/src/contexts/ActionPanelContext.tsx
@@ -6,16 +6,22 @@ import {
   type Dispatch,
   type SetStateAction,
   useContext,
+  useEffect,
   useState,
 } from "react";
+import { useCanvasCooldown, usePlayCooldownExpirySound } from "@/hooks";
+import { useAuthContext } from "./AuthProvider";
+import { useCanvasContext } from "./CanvasContext";
 
 type TabKey = "look" | "place" | "frame";
 
 interface ActionPanelContextType {
   areTabsLocked: boolean;
+  cooldownEndTime: number | null;
   currentTab: TabKey;
   isFullscreenPanelVisible: boolean;
   setAreTabsLocked: Dispatch<SetStateAction<boolean>>;
+  setCooldownEndTime: Dispatch<SetStateAction<number | null>>;
   setCurrentTab: Dispatch<SetStateAction<TabKey>>;
   setFullscreenPanelVisible: Dispatch<SetStateAction<boolean>>;
   setTempColor: Dispatch<SetStateAction<PaletteColor | null>>;
@@ -24,9 +30,11 @@ interface ActionPanelContextType {
 
 const ActionPanelContext = createContext<ActionPanelContextType>({
   areTabsLocked: false,
+  cooldownEndTime: null,
   currentTab: "place",
   isFullscreenPanelVisible: false,
   setAreTabsLocked: () => {},
+  setCooldownEndTime: () => {},
   setCurrentTab: () => {},
   setFullscreenPanelVisible: () => {},
   setTempColor: () => {},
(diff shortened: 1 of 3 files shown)


packages/frontend/src/contexts/ActionPanelContext.tsx
```diff
@@ -38,16 +46,54 @@ interface ActionPanelProviderProps {
...
   const [areTabsLocked, setAreTabsLocked] = useState(false);
   const [isFullscreenPanelVisible, setFullscreenPanelVisible] = useState(false);
 
+  // Tracks the end time of the most recent pixel placement cooldown
+  const [placementCooldownEndTime, setPlacementCooldownEndTime] = useState<
+    number | null
+  >(null);
+
+  // Reset the placement cooldown during render when the canvas changes, rather
+  // than from an effect. See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
```
jaskfla: Love this pattern, but we already do this in a couple places, so might be a bit odd to only comment it here

@Rocked03 can attest that I’m always harping on about abuses of Effects 😅

packages/frontend/src/contexts/ActionPanelContext.tsx
```diff
@@ -38,16 +46,54 @@ interface ActionPanelProviderProps {
...
+
   const [currentTab, setCurrentTab] = useState<TabKey>("place");
   const [tempColor, setTempColor] = useState<PaletteColor | null>(null);
   const [areTabsLocked, setAreTabsLocked] = useState(false);
   const [isFullscreenPanelVisible, setFullscreenPanelVisible] = useState(false);
 
+  // Tracks the end time of the most recent pixel placement cooldown
+  const [placementCooldownEndTime, setPlacementCooldownEndTime] = useState<
+    number | null
+  >(null);
```
jaskfla: I don’t hate this, but I’m not quite sure why we can’t trust the TanStack `QueryClient`’s data?

`staleTime` is set to `Infinity` for the cooldown query, so I would expect every instance of `useCanvasCooldown()` to see the same value because basically the entire app is wrapped in the same `QueryContextProvider`

Anywhere we need to call this setter, we should _probably_ just be invalidating or resetting the `QueryClient`’s cached data

packages/frontend/src/contexts/ActionPanelContext.tsx
```diff
@@ -38,16 +46,54 @@ interface ActionPanelProviderProps {
...
+
+  // Reset the placement cooldown during render when the canvas changes, rather
+  // than from an effect. See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
+  const [previousCanvasId, setPreviousCanvasId] = useState(canvas.id);
+  if (previousCanvasId !== canvas.id) {
+    setPreviousCanvasId(canvas.id);
+    setPlacementCooldownEndTime(null);
```
jaskfla: As mentioned in earlier comment in this file, I suspect this only needs to be `queryClient.invalidateQuery({ queryKey: ["canvasCooldown", canvas.id] })` (or `resetQuery)

packages/frontend/src/contexts/ActionPanelContext.tsx
```diff
@@ -48,33 +49,24 @@ interface ActionPanelProviderProps {
...
   });
 
-  // Effective cooldown: whichever of the API result or last placement expires later
-  const queryCooldownEndTime = cooldownData?.cooldownEndTime ?? null;
-  const cooldownEndTime =
-    Math.max(queryCooldownEndTime ?? 0, placementCooldownEndTime ?? 0) || null;
+  const cooldownEndTime = cooldownData?.cooldownEndTime ?? null;
+
+  const setCooldownEndTime = (value: number | null) => {
+    queryClient.setQueryData<Cooldown>(["canvasCooldown", canvas.id], {
```
jaskfla, line 66: Oh, optimistic update! Smart!

In most cases I would just invalidate and let the QueryClient refetch, but this is a great candidate for manually writing to its cache

jaskfla (approved):
Love it—better than my suggestion!

PR #653, opened 2026-05-24
fix(backend): add custom key generator for rate limiting based on cf-connecting-ip

stijnvdkolk:


packages/backend/src/middleware/ratelimit.ts
```diff
@@ -1,5 +1,17 @@
+import type { Request } from "express";
 import rateLimit from "express-rate-limit";
 
+const keyGenerator = (req: Request) => {
```
Rocked03, line 6: What's this doing? Can we please include a comment explaining why we need this?

stijnvdkolk: It makes sure we take the user ip instead of the proxy ip (CloudFlare in this case), we can also add some logic to the Caddyfile that will map it to the X-Forwarded-For header and then we can use that here if you do not want this to be directly linked to CloudFlare in the code.

Rocked03: it would probably be good to not be specifically assuming cloudflare

stijnvdkolk: Done

Rocked03 (approved):

PR #664, opened 2026-05-25
feat(backend): integrate Kysely with Prisma for better raw queries

stijnvdkolk:


Rocked03:
wondering whether we should save this until post-event, like #476

stijnvdkolk:
> wondering whether we should save this until post-event, like #476 

Fine by me

jaskfla:
Had a scan through and seems fine to me. Couple disorganised notes:

- Curious that Kysely is inspired by Knex but is already a larger dependency, but nothing outrageous so I won’t complain.[^foot]
- I trust Kysely supports nested queries?
- ~~While I’m very much in favour of _having_ a safe query builder API that will take care of escaping stuff before it gets interpreted as SQL, for static queries I’d probably still lean slightly towards raw SQL as a string literal~~
- ~~It’s an [even heavier dependency](https://npmx.dev/compare?packages=kysely,knex,sequelize), but [Sequelize has a _really nice API_ using tagged template literals](https://sequelize.org/docs/v7/querying/raw-queries) that feels like a much thinner abstraction:~~

```js
import { sql } from '@sequelize/core';

const id = 5;

// Sequelize will escape `id` 👇🤯
await sequelize.query(sql`SELECT * FROM users WHERE id = ${id}`);
```

[^foot]: Between the two, I’m keen to try Kysely purely to get to try something new

stijnvdkolk:
Kysely supports nested queries. The nicest thing here is that our Kysely types are generated from the Prisma schema, which means that we have typed queries, and if we ever update the Prisma schema then we'll immediately know that a query might be broken.

The sequelize example kind of brings you back to what we already had with normal raw queries in Prisma before, and as far as I am aware those are also untyped 😅

jaskfla:
Ooh, in that case I also retract my comment about literal SQL for static queries. Sounds great; I’ll give it a proper review post-event 🙏

packages/backend/src/services/historyService.ts
```diff
@@ -227,109 +117,88 @@ async function getPixelHistoryRowsWithCount({
...
-  const fragments: Prisma.Sql[] = [
-    Prisma.sql`h.erased_at IS NULL`,
-    Prisma.sql`h.canvas_id = ${params.canvasId}`,
-    Prisma.sql`h.x >= ${points[0].x} AND h.x <= ${points[1].x}`,
-    Prisma.sql`h.y >= ${points[0].y} AND h.y <= ${points[1].y}`,
+  const conditions: Expression<SqlBool>[] = [
+    eb("history.erased_at", "is", null),
+    eb("history.canvas_id", "=", params.canvasId),
+    eb.between("history.x", points[0].x, points[1].x),
+    eb.between("history.y", points[0].y, points[1].y),
```
Rocked03, line 168: double checking this does the inclusivity right?

Rocked03 (approved):
I'm certainly not familiar enough with Kysely itself to nitpick any syntax, but it all looks like a pretty sane migration to me.

Approving, although I'm sure @jaskfla will have a few thoughts

jaskfla (approved):

packages/backend/src/services/historyService.ts
stijnvdkolk: Yup!

PR #677, opened 2026-05-26
fix(frontend): update mutationKey in usePlacePixelMutation to remove coords dependency

### LINKED ISSUE #675: Switch pixel clickable place button when request response is loading
When we place a pixel, it changes the place button with `Loading` and a circular rotative icon.
When we get the response of the successful pixel place post request, the button is changed to `On cooldown (Xs)`.

If we select another pixel, another coordinate, between those two button update, the place pixel button becomes clickable again until the post response. This makes people easily eating the per-overlay Rate Limit.
This have nothing to do with the endpoint Rate Limit eating multiple `ratelimit-remaining` (http headers) for a single request.

stijnvdkolk:
This seems to fix #675. If the server responds somewhat slow you are able to press another pixel and then mutation key changes which makes the previous cooldown show up in the background and never get added the button. I'm sure there is a better fix to #675 but I felt like this a quick fix and then later we can investigate a better more sophisticated fix for this 😅

### DIFF
--- a/packages/frontend/src/components/action-panel/tabs/place/usePlacePixelMutation.ts
+++ b/packages/frontend/src/components/action-panel/tabs/place/usePlacePixelMutation.ts
@@ -19,7 +19,8 @@ export default function usePlacePixelMutation(
   const { color } = useSelectedColorContext();
 
   return useMutation<Cooldown>({
-    mutationKey: ["pixel", canvas.id, coords],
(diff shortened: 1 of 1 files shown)


jaskfla (approved):
Optional: would appreciate a `// TODO: figure out why coords breaks/causes…`, but happy with this as hotfix

Thanks 🙏

stijnvdkolk:
Will add that!

PR #683, opened 2026-05-27
feat: mod dashboard overlay

Rocked03:
image

Added an overlay that highlights the filtered history entries returned by the complex search query in the mod dashboard


stijnvdkolk (approved):
backend looks good

Rocked03:
I'll let this sit for a little to give time for @jaskfla to have a peek when he has the chance

but if some mod stuff crops up in the event, I'll probably just feed it through - i'm not too concerned considering this is only in the moderation dashboard

Rocked03:
^ mod stuff cropped up

PR #684, opened 2026-05-28
feat(backend): development auth bypass

Rocked03:
Because Discord API ratelimits are a nightmare

`BYPASS_AUTH: true`

image

### DIFF
--- a/packages/backend/src/config/index.ts
+++ b/packages/backend/src/config/index.ts
@@ -30,6 +30,7 @@ const config = {
   // having a random secret would mess with persistent sessions
   expressSessionSecret:
     process.env.EXPRESS_SESSION_SECRET || "change the secret in production",
+  bypassAuth: process.env.BYPASS_AUTH === "true",
   discord: {
     clientId: requiredEnv("DISCORD_CLIENT_ID"),
     clientSecret: requiredEnv("DISCORD_CLIENT_SECRET"),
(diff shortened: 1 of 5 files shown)


packages/backend/src/config/index.ts
```diff
@@ -60,6 +61,35 @@ const config = {
...
+      "#  Canvas moderator and admin authorization checks are DISABLED.     #",
+      "#  This must only be used for local development.                     #",
+      "#  Do not deploy with BYPASS_AUTH=true.                              #",
+      "#                                                                    #",
+      "######################################################################",
+      "",
+    ].join("\n"),
+  );
+
+  if (config.environment === "production") {
```
jaskfla: Maybe flip this and only allow `development`?

packages/backend/src/config/index.ts
```diff
@@ -60,6 +61,35 @@ const config = {
   databaseUrl: requiredEnv("DATABASE_URL"),
 } as const;
 
+if (config.bypassAuth) {
+  console.warn(
```
jaskfla: Nice use of `String.raw`. Optional, I think the entire thing could be a single template

jaskfla (commented):
One big question:

Why commit this to VCS at all? Stash and pop stash as needed locally

Rocked03:
> Why commit this to VCS at all? Stash and pop stash as needed locally

imo, it's prevalent-enough of an issue to be a nuisance for everybody, makes it easier to configure on a whim

Rocked03:
just gonna close this for now, can revisit later if we need it

PR #686, opened 2026-05-28
feat(backend): generate snapshots

Rocked03:


packages/backend/src/config/index.ts
stijnvdkolk, line 1: Might be nice to switch to zod for this (later)

Rocked03:
ok i'm marking this one as ready for review, but it's probably quite rough around the edges

definitely want @stijnvdkolk's sign off on the prisma stuff here because i've literally created a second instance of prisma and my IDE isn't quite always happy about that... lmk if there's some tidying up that can be done there :pray:

packages/backend/prisma/schema.prisma
stijnvdkolk, line 1: We could change this to core/schema.prisma to make more clear that this is for the core app? Maybe core is not the best name, but I think a subfolder for this one makes sense with the snapshots being in a separate dir too.

packages/backend/src/services/snapshot/generateSnapshotService.ts
```diff
@@ -0,0 +1,238 @@
...
+      c.name AS color__name,
+      c.rgba AS color__rgba
+    FROM history h
+    INNER JOIN color c ON c.id = h.color_id
+    WHERE h.canvas_id = ${canvasId}
+      AND h.erased_at IS NULL
+      AND h.timestamp >= ${from}
+      AND h.timestamp < ${to}
+    ORDER BY h.x, h.y, h.timestamp DESC, h.id DESC
+  `;
```
stijnvdkolk: Kysely? 👀

packages/backend/src/services/snapshot/scheduler.ts
```diff
@@ -0,0 +1,236 @@
...
+  const readyWindows = await prisma.$queryRaw<ReadySnapshotWindowRow[]>`
+    SELECT
+      canvas_id,
+      bucket_start,
+      bucket_end,
+      history_count
+    FROM history_snapshot_windows
+    WHERE bucket_end <= ${cutoff}
+    ORDER BY canvas_id ASC, bucket_start ASC
+  `;
```
stijnvdkolk: Kysely? 👀

packages/backend/prisma/schema.prisma
Rocked03: "core" feels quite appropriate to me, i'm happy with that

Rocked03: going to defer to a separate PR as not to clutter up this one with a big refactor

packages/backend/src/services/historyService.ts
```diff
@@ -463,6 +464,11 @@ export async function deletePixelHistoryEntries(
     coordinatesUpdated,
   );
 
+  const earliestEntryTimestamp = deletedEntries.reduce((earliest, entry) => {
+    return entry.timestamp < earliest ? entry.timestamp : earliest;
+  }, deletedEntries[0].timestamp);
+  await setSnapshotDirtyTimestamp(params.canvasId, earliestEntryTimestamp);
```
stijnvdkolk, line 470: `setSnapshotDirtyTimestamp` is called unconditionally here (and in `restorePixelHistoryEntries`), but it isn't gated by `isSnapshotGenerationEnabled()` / `isSnapshotAvailableForCanvas()`, and it writes to the SQLite snapshot db via `snapshotPrisma.snapshot_cursor.upsert`.

The snapshot DB is only created by the manual `prisma:snapshots:migrate:deploy` script — `assertSnapshotDatabaseReady()` only runs when `GENERATE_SNAPSHOTS=true`, and CI only runs the Postgres `prisma:migrate`. So in the default config (`GENERATE_SNAPSHOTS=false`, empty allowlist) where the sidecar isn't provisioned, an admin erasing/restoring pixels will hit a missing file/table and throw `no such table: snapshot_cursor` (or "unable to open database file"), breaking the moderation flow.

packages/backend/src/services/snapshot/snapshotService.ts
```diff
@@ -0,0 +1,63 @@
...
+  timestamp: Date,
+) {
+  await snapshotPrisma.snapshot_cursor.upsert({
+    where: { canvas_id: canvasId },
+    create: {
+      canvas_id: canvasId,
+      dirty_from_timestamp: timestamp,
+    },
+    update: { dirty_from_timestamp: timestamp },
+  });
```
stijnvdkolk: `dirty_from_timestamp` represents the earliest point from which snapshots must be regenerated (the scheduler regenerates windows where `dirty_from_timestamp <= bucket_end`). This `update` overwrites it with the new `timestamp` unconditionally.

If a modification at `T1` is still pending and a later modification at `T2 > T1` arrives before the scheduler runs, dirty becomes `T2`, and the windows between `T1` and `T2` (which contain the first edit) have `bucket_end < T2` and are never regenerated — so the `T1` edit is silently dropped from snapshots.

This should be `min(existing, timestamp)`, e.g. a conditional `updateMany` guarded by `dirty_from_timestamp > timestamp OR dirty_from_timestamp IS NULL`. (Also worth gating this whole function on snapshot availability — see the comment on `historyService.ts`.)

packages/backend/src/services/snapshot/scheduler.ts
```diff
@@ -0,0 +1,226 @@
...
+        snapshotAt.getTime() >= cursor.dirty_from_timestamp.getTime();
+
+      await snapshotPrisma.snapshot_cursor.update({
+        where: { canvas_id: canvasId },
+        data: {
+          last_processed_timestamp: snapshotAt,
+          dirty_from_timestamp:
+            shouldClearDirty ? null : cursor.dirty_from_timestamp,
+        },
+      });
```
stijnvdkolk: The cursor map is loaded once at the start of the cycle (lines 90-92) and never refreshed, yet this update writes `dirty_from_timestamp` from that stale in-memory copy. `buildSnapshot` does sharp decode/encode, so a cycle can take a while. If a user erases a pixel mid-cycle, `setSnapshotDirtyTimestamp` writes a dirty value to the DB, and then this blind write overwrites it (with `null` if the cursor was clean at cycle start) — the just-recorded edit is lost.

Prefer a conditional update that only clears dirty when the *current DB* value is still `<= snapshotAt`, rather than writing the cached value back.

Two related notes:
- The multi-window dirty cascade currently works *only because* this in-memory `cursor` is never mutated (every window re-evaluates against the original dirty value). That's subtle/fragile — worth a comment, since a future refactor that refreshes the map per window would silently break it.
- The `else { create }` branch below is effectively dead for allowlisted canvases (they're always pre-created in the loop at lines 94-107).

packages/backend/src/services/snapshot/scheduler.ts
```diff
@@ -0,0 +1,226 @@
...
+  const now = new Date();
+  const cutoff = getWindowCutoff(now);
+
+  const readyWindows = await prisma.$kysely
+    .selectFrom("history_snapshot_windows")
+    .select(["canvas_id", "bucket_start", "bucket_end", "history_count"])
+    .where("bucket_end", "<=", cutoff)
+    .orderBy("canvas_id", "asc")
+    .orderBy("bucket_start", "asc")
+    .execute();
```
stijnvdkolk: `history_snapshot_windows` is a view that `GROUP BY`s the full `history` table, and this query fetches every bucket since the canvas began (no lower bound) every `schedulerIntervalMs` (default 60s). The result set and GROUP BY cost grow unbounded with history, and most rows are then skipped in JS via the cursor comparison.

Consider lower-bounding `bucket_start` per canvas by `last_processed_timestamp` / `dirty_from_timestamp` so each cycle only materializes the windows it might actually process.

packages/backend/prisma/migrations/20260527134000_add_history_snapshot_windows_view/migration.sql
```diff
@@ -0,0 +1,19 @@
...
+SELECT
+  bucketed.canvas_id,
+  bucketed.bucket_start,
+  bucketed.bucket_start + INTERVAL '10 minutes' AS bucket_end,
+  COUNT(*)::integer AS history_count
+FROM (
+  SELECT
+    h.canvas_id,
+    date_trunc('hour', h.timestamp)
+      + (floor(extract(minute from h.timestamp) / 10) * INTERVAL '10 minutes') AS bucket_start
```
stijnvdkolk: `history.timestamp` is `timestamptz`, and `date_trunc('hour', h.timestamp)` evaluates in the DB session timezone. Meanwhile `getWindowCutoff` in `scheduler.ts` is pure-UTC epoch math (`floor(getTime() / WINDOW) * WINDOW`). On a non-UTC session, `bucket_end` won't align to those UTC boundaries, so windows can be considered "ready" early/late.

Safer to force UTC here, e.g. `date_trunc('hour', h.timestamp AT TIME ZONE 'UTC')` (or `timezone('UTC', ...)`), so bucketing is deterministic regardless of session tz.

packages/backend/prisma/snapshots/migrations/20260526122845_init_snapshot_manifest/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+    "file_size_bytes" INTEGER,
+    "created_at" DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated_at" DATETIME NOT NULL
+);
+
+-- CreateTable
+CREATE TABLE "snapshot_cursor" (
+    "canvas_id" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    "last_processed_timestamp" DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "dirty_from_history_id" BIGINT,
```
stijnvdkolk: This migration declares `snapshot_cursor.canvas_id` as `INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT` and includes a `dirty_from_history_id BIGINT` column, but `prisma/snapshots/schema.prisma` has neither (`canvas_id Int @id`, no autoincrement, no `dirty_from_history_id`). This will trip `prisma migrate` drift detection.

It only works at runtime because every insert passes `canvas_id` explicitly — but `AUTOINCREMENT` on the *real* canvas id (not a synthetic key) is semantically wrong and a footgun. Recommend regenerating a clean init migration that matches the schema (drop `AUTOINCREMENT`, remove the leftover `dirty_from_history_id`).

packages/backend/prisma/snapshots/migrations/20260526122845_init_snapshot_manifest/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+    "dirty_from_timestamp" DATETIME,
+    "updated_at" DATETIME NOT NULL
+);
+
+-- CreateIndex
+CREATE INDEX "snapshot_manifest_canvas_id_snapshot_at_idx"
+ON "snapshot_manifest"("canvas_id", "snapshot_at");
+
+CREATE UNIQUE INDEX "snapshot_manifest_canvas_id_snapshot_at_key"
+ON "snapshot_manifest"("canvas_id", "snapshot_at");
```
stijnvdkolk, line 24: This creates both a non-unique index and a unique index on the same `(canvas_id, snapshot_at)` columns. The unique index already serves lookups, so the plain `CREATE INDEX` above is redundant.

packages/backend/src/services/snapshot/snapshotService.ts
```diff
@@ -0,0 +1,63 @@
...
+export async function getSnapshots({
+  canvasId,
+  from,
+  to,
+}: {
+  canvasId: CanvasInfo["id"];
+  from?: Date;
+  to?: Date;
+}) {
+  if (!isSnapshotAvailableForCanvas(canvasId)) {
```
stijnvdkolk, line 14: `getSnapshots` is gated only by `isSnapshotAvailableForCanvas` (the allowlist), not by DB readiness, unlike the generate path which calls `assertSnapshotDatabaseReady()`. An allowlisted canvas with `GENERATE_SNAPSHOTS=false` and an unprovisioned sidecar would throw when this is eventually called. (Related to the moderation issue flagged on `historyService.ts`.)

packages/backend/src/services/historyService.ts
S-Ou: For now, I'm going to effectively not handle a missing DB file, since that's effectively had we treat the real DB (just let it error if it can't find it)

If we have better ideas we can apply it to both DBs in another PR

packages/backend/src/services/snapshot/scheduler.ts
```diff
@@ -118,7 +136,15 @@ export async function runSnapshotSchedulerCycle(): Promise<{
...
+      // This should never happen since we ensure cursors exist for all canvases above, but just in case...
+      console.warn(
+        `No snapshot cursor found for canvas ${canvasId}, skipping snapshot generation for this cycle.`,
+      );
+      skipped += 1;
+      continue;
+    }
 
     const shouldGenerate =
       !cursor || // No cursor means this is the first snapshot, so we should generate.
```
stijnvdkolk: This will never trigger anymore with the above if

stijnvdkolk (approved):

PR #687, opened 2026-05-28
feat(backend): generate timelapses

### LINKED ISSUE #532: Timelapse
It would be great to be able to create timelapses in our backend and returned to the user via an (authenticated) endpoint. 

@volcanofr has some good timelapse code that we could potentially implement into the project

Alternatively, and potentially optimally, we should follow the footsteps of the r/place Atlas guys - pre-generate static image snapshots periodically through the lifespan of the canvas, which means that the timelapse creation is inexpensive (only need to put the images together rather than regenerate each frame)

Rocked03:


volcanofr:
This is looking amazing!

Rocked03:
todo: purge outdated cached timelapses

Rocked03:
currently the endpoint is completely public, but we should realistically lock it behind the standard user authentication at the least

perhaps also apply #719 captcha to it as well, i'd rather we don't have people spamming this endpoint with scripts

packages/backend/src/routes/api/v1/canvas.ts
Rocked03, line 1: add tags

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -100,6 +102,62 @@ canvasRouter.get(
...
+    res
+      .status(200)
+      .type(raw ? "webm" : "mp4")
+      .setHeader(
+        "Content-Disposition",
+        `inline; filename="canvas-${req.params.canvasId}-timelapse${raw ? "-raw" : ""}.${raw ? "webm" : "mp4"}"`,
+      )
+      .send(buffer);
+  },
+);
```
stijnvdkolk, line 161: Isn't this the same route?

packages/backend/src/services/frameService.ts
```diff
@@ -454,3 +454,51 @@ export async function assertMaxOwnerFramesNotExceeded({
...
+    FROM frame_bounds fb
+    INNER JOIN history h
+      ON h.canvas_id = fb.canvas_id
+      AND h.erased_at IS NULL
+      AND h.x >= fb.x_0
+      AND h.x < fb.x_1
+      AND h.y >= fb.y_0
+      AND h.y < fb.y_1
+    GROUP BY fb.id
+  `;
```
stijnvdkolk: Kysely!

packages/backend/src/routes/api/v1/canvas.ts
Rocked03: Lol

I blame the merge conflict

🤦

packages/types/src/models/canvas.models.ts
```diff
@@ -43,3 +43,7 @@ export const CanvasExportParamModel = z.object({
   canvasId: CanvasIdParamModel.shape.canvasId,
   scale: CanvasExportScaleSchema,
 });
+
+export const CanvasTimelapseParamModel = z.object({
+  raw: z.coerce.boolean().optional(),
```
stijnvdkolk: `z.coerce.boolean()` returns `true` for any non-empty string, so `?raw=false` and `?raw=0` both parse as `true`. Since the repo is on zod 4, `raw: z.stringbool().optional()` handles `"false"`/`"0"`/`"off"` correctly (or use an explicit `z.enum(["true","false"]).transform(...)`).

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -100,6 +102,62 @@ canvasRouter.get(
...
+    const raw = req.query.raw || false;
+
+    const buffer = await generateTimelapse(
+      raw ?
+        {
+          canvasId: req.params.canvasId,
+          raw: true,
+        }
+      : { canvasId: req.params.canvasId },
+    );
```
stijnvdkolk: Unlike the frame endpoint in `frame.ts`, this route doesn't check `canvas.isLocked`, and it has no rate limiter. On an unlocked canvas the snapshot cursor's `updated_at` keeps advancing, so the cached timelapse repeatedly goes stale and each request can trigger a full ffmpeg re-encode. This route is unauthenticated and CPU-heavy, so it's an easy DoS target.

packages/backend/src/routes/api/v1/frame.ts
```diff
@@ -82,6 +85,39 @@ frameRouter.get(
   },
 );
 
+frameRouter.get(
+  "/:frameId.mp4",
+  validate({ params: FrameIdParamModel }),
+  async (req, res) => {
```
stijnvdkolk, line 98: No rate limiting here either, and every distinct frame gets its own cache key, so one client can kick off many concurrent ffmpeg encodes by iterating frame IDs.

packages/backend/src/services/frameService.ts
```diff
@@ -454,3 +454,51 @@ export async function assertMaxOwnerFramesNotExceeded({
...
+    WITH frame_bounds AS (
+      SELECT
+        id,
+        canvas_id,
+        x_0,
+        y_0,
+        x_1,
+        y_1
+      FROM frame
+      WHERE id ILIKE ${frameId}
```
stijnvdkolk: `ILIKE` does pattern matching, so `%` and `_` in `frameId` act as wildcards, and `LIMIT 1` without `ORDER BY` makes the matched row non-deterministic. Route input is constrained to 6 hex chars today, but this is an exported service function. Since you only need the case-insensitive equality that `getFrameById` uses, prefer `WHERE LOWER(id) = LOWER(${frameId})`.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+
+    if (cropWidth <= 0 || cropHeight <= 0) {
+      throw new Error("Bounds are invalid after normalization and clamping");
+    }
+
+    const isFullCanvasBounds =
+      clampedBounds.x0 === 0 &&
+      clampedBounds.y0 === 0 &&
+      clampedBounds.x1 === canvas.width - 1 &&
+      clampedBounds.y1 === canvas.height - 1;
```
stijnvdkolk: Frame bounds are exclusive at `x1`/`y1` in this codebase (`assertCoordsAreWithinCanvas` allows `x1 === canvas.width`, `exportCanvasBoundsAsStream` uses `width = x1 - x0`, and the history query uses `h.x < fb.x_1`). Clamping `x1`/`y1` to `canvas.width - 1`/`canvas.height - 1` therefore drops the last column/row for any frame touching the right/bottom canvas edge.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+}): Promise<void> {
+  for (const imagePath of imagePaths) {
+    const imageBuffer = await readFile(imagePath);
+    if (!stdin.write(imageBuffer)) {
+      await new Promise((resolve) => stdin.once("drain", resolve));
+    }
+  }
+
+  stdin.end();
+}
```
stijnvdkolk: If ffmpeg exits while images are still being piped (bad input, encoder failure), `stdin.write()` emits an `'error'` (EPIPE) on a stream that has no error listener, which crashes the whole process. The `'drain'` wait can also hang forever if the process dies before draining, leaving the `onProcess` promise pending. Possible fix: attach `stdin.on('error', ...)` in `runFfmpegProcess`, race the drain wait against the process `'close'` event, and consider a timeout that kills hung ffmpeg processes.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+
+  if (!ffmpegPath) {
+    throw new Error("ffmpeg-static did not provide a binary path");
+  }
+
+  const tempPrefix = `${process.pid}-${Date.now()}`;
+  const tempTimelapsePath = `${TIMELAPSE_END_CARD_IMAGE_PATH}.timelapse-${tempPrefix}.mp4`;
+  const tempTailPath = `${TIMELAPSE_END_CARD_IMAGE_PATH}.tail-${tempPrefix}.mp4`;
+  const tempLastFramePath = `${TIMELAPSE_END_CARD_IMAGE_PATH}.last-frame-${tempPrefix}.png`;
+  const tempEndCardPath = `${TIMELAPSE_END_CARD_IMAGE_PATH}.end-card-${tempPrefix}.png`;
```
stijnvdkolk: These temp files are written next to the bundled end-card asset in `static/constant/`, a checked-in assets directory (and unlike `static/timelapse` it isn't gitignored), so a crash between write and the `finally` cleanup leaves junk there. We could build the paths from `tmpdir()` like `encodeMainVideoFromImages` already does.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+    existingCache &&
+    currentCursorUpdatedAt &&
+    existingCache.updated_at >= currentCursorUpdatedAt
+  ) {
+    const cachedBuffer = await readCachedTimelapse(existingCache.file_path);
+    if (cachedBuffer) {
+      // Return from cache
+      return cachedBuffer;
+    }
+  }
```
stijnvdkolk: This freshness check has two side effects: (1) if the cursor row doesn't exist (`currentCursorUpdatedAt === null`) the cache is never used and every request re-encodes; (2) any cursor update (new snapshot, dirty mark) invalidates *all* cached timelapses for the canvas, including fully historical ranges whose underlying snapshots didn't change — each gets re-encoded once per cursor bump. Since the cache key already pins `effectiveStartAt`/`effectiveEndAt` from the actual snapshots, consider only invalidating when a dirty re-render overlaps that range (or treat a missing cursor as fresh). Also note there's no eviction: superseded files and manifest rows accumulate on disk indefinitely as `effectiveEndAt` advances.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+}
+
+async function readCachedTimelapse(filePath: string): Promise<Buffer | null> {
+  try {
+    await stat(filePath);
+    return await readFile(filePath);
+  } catch {
+    return null;
+  }
+}
```
stijnvdkolk: Whole videos are buffered in memory at every stage: `readFile` here for cache hits, `Buffer.concat` of ffmpeg stdout, and `res.send(buffer)` in the routes — plus coalesced in-flight requests all hold the same large buffer. For big canvases (raw lossless VP9 especially) that's potentially hundreds of MB per request. Possible fix: return the cached file path and use `res.sendFile`/`createReadStream` for hits, and have generation always write to the final file and stream from disk instead of capturing stdout.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+    // and cached before producing the augmented timelapse. This lets callers
+    // request the raw variant later without re-encoding the frames.
+    const rawCacheParams: TimelapseCacheParams = {
+      ...cacheParams,
+      endHoldDurationMs: null,
+      showEndCard: false,
+    };
+    const rawCacheKey = buildTimelapseCacheKey(rawCacheParams);
+    const rawFileName = `${rawCacheKey}.webm`;
+    const rawFilePath = getTimelapseVideoPath(canvasId, rawFileName);
```
stijnvdkolk: This comment doesn't match the code: when `cacheParams.raw` is false nothing is written for a raw variant at all, and when it's true `generateTimelapse` has already forced `endHoldDurationMs = null` / `showEndCard = false`, so `rawCacheKey === cacheKey` and `rawFilePath === timelapseFilePath`. The `rawCacheParams`/`rawCacheKey`/`rawFilePath` indirection is dead — either drop it and reuse `cacheKey`/`timelapseFilePath` in the `if (cacheParams.raw)` branch, or actually implement the pre-caching the comment describes.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+        "-c:v",
+        "libx264",
+        "-pix_fmt",
+        "yuv420p",
+        "-movflags",
+        "frag_keyframe+empty_moov",
+        "-f",
+        "mp4",
+        "pipe:1",
+      ],
```
stijnvdkolk: This concat re-encodes the entire main video a second time with libx264 at default settings (extra CPU plus generational quality loss). Since the tail is produced with the same codec/resolution/pixel format, the concat demuxer with stream copy (`-f concat -safe 0 -i list.txt -c copy`) could splice the two files without re-encoding the main segment.

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,927 @@
...
+const END_CARD_TRANSITION_DURATION_MS = 1_000;
+const END_CARD_DISPLAY_DURATION_MS = 5_000;
+const END_CARD_BACKGROUND_COLOR = {
+  r: 88,
+  g: 101,
+  b: 242,
+  alpha: 1,
+} as const;
+
+interface generateTimelapseParams {
```
stijnvdkolk: `GenerateTimelapseParams`

packages/backend/prisma/snapshots/schema.prisma
```diff
@@ -31,3 +31,28 @@ model snapshot_cursor {
...
+  effective_start_at   DateTime
+  effective_end_at     DateTime
+  bounds_x0            Int?
+  bounds_y0            Int?
+  bounds_x1            Int?
+  bounds_y1            Int?
+  scale                Int
+  frame_rate           Int
+  end_hold_duration_ms Int
+  show_end_card        Boolean
```
stijnvdkolk: The migration declares `show_end_card BOOLEAN NOT NULL DEFAULT 1`, but the model is missing `@default(true)`, so schema and migration history have drifted — the next `prisma migrate dev` will generate a spurious migration dropping the default. Add `@default(true)` here.

stijnvdkolk (changes requested):

packages/backend/src/services/timelapseService.ts
S-Ou: whoops

the original intent was that it _would_ use the raw timelapse as a base but then things diverged

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -111,50 +111,50 @@ canvasRouter.get(
...
-  }),
-  async (req, res) => {
-    const raw = req.query.raw || false;
+    const canvas = await getCanvasInfo(req.params.canvasId);
 
-    const buffer = await generateTimelapse(
+    if (!canvas.isLocked) {
+      res.status(400).json({
+        error: "Timelapse generation is only available for locked canvases",
+      });
```
stijnvdkolk: `throw new BadRequestError("Timelapse generation is only available for locked canvases")`

packages/backend/src/routes/api/v1/frame.ts
Rocked03: a few ideas
- rate limit by user auth
- lock timelapse generation to frame author
- lock timelapse generation to only frames that contain at least x buckets of history (might be hard to calculate this one without still querying the history entries)

Rocked03: If we're being generous, we could instead allow timelapses on unlocked canvases but only allow the buckets to update daily

stijnvdkolk: Would be cool. I think limiting timelapse generation to the frame owner and then possibly checking the x amount of history buckets would be nice. And in general ratelimit by user auth sounds good to me

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -100,6 +102,62 @@ canvasRouter.get(
...
+      return;
+    }
+
+    const filePath = await generateTimelapse(
+      raw ?
+        {
+          canvasId: req.params.canvasId,
+          raw: true,
+        }
+      : { canvasId: req.params.canvasId },
```
jaskfla: Does this not work? If not, consider making `generateTimelapse` accept `raw: boolean | undefined`

```suggestion
      {
        canvasId: req.params.canvasId,
        raw
      }.
```

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -100,6 +102,62 @@ canvasRouter.get(
...
+
+    const fileStats = await stat(filePath);
+
+    addSpanAttributes(req, {
+      "response.size": fileStats.size,
+    });
+
+    res
+      .status(200)
+      .type(raw ? "webm" : "mp4")
```
jaskfla, line 147: Huh, maybe instead of `raw` is should just be `format: 'webm' | 'mp4'`?

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -100,6 +102,62 @@ canvasRouter.get(
...
 );
 
+canvasRouter.get(
+  "/:canvasId.mp4",
+  validate({
+    params: CanvasIdParamModel,
+    query: CanvasTimelapseParamModel,
+  }),
+  async (req, res) => {
+    const raw = req.query.raw || false;
```
jaskfla, line 119: What does `raw` mean?

packages/backend/src/routes/api/v1/frame.ts
jaskfla: I reckon the longer term solution is to have a queue of jobs, which we can just process maybe one or two at a time, and DM the requester with a download link when it’s done

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,956 @@
...
+    throw new Error("ffmpeg-static did not provide a binary path");
+  }
+
+  const tempPrefix = `${process.pid}-${Date.now()}`;
+  const tempTimelapsePath = join(tmpdir(), `timelapse-${tempPrefix}.mp4`);
+  const tempTailPath = join(tmpdir(), `tail-${tempPrefix}.mp4`);
+  const tempLastFramePath = join(tmpdir(), `last-frame-${tempPrefix}.png`);
+  const tempEndCardPath = join(tmpdir(), `end-card-${tempPrefix}.png`);
+  const tempConcatListPath = join(tmpdir(), `concat-list-${tempPrefix}.txt`);
+  const tempConcatPath = join(tmpdir(), `concat-${tempPrefix}.mp4`);
```
jaskfla: Rather than using `tmpdir()` directly, make a subfolder called something like `com.projectblurple.canvas` (or that new domain we recently nabbed)

e.g. for me, it should end up in `/var/folders/f7/xrgx7brn3j5dn078y4044frw0000gn/T/com.projectblurple.canvas`

Potentially worth making a small helper to ensure the directory exists with `mkdir(theDir, { recursive: true })`, and updating these lines to call that helper

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,956 @@
...
+export async function generateTimelapse({
+  canvasId,
+  start,
+  end,
+  bounds,
+  frameRate = 30,
+  endHoldDurationMs = 2000,
+  showEndCard = true,
+  scale,
+  backgroundColor = [35, 39, 42, 255],
```
jaskfla: Maybe define this duple with a name (is it Dark Blurple)?`

packages/backend/src/services/timelapseService.ts
```diff
@@ -0,0 +1,956 @@
...
+  raw = false,
+}: GenerateTimelapseParams): Promise<string> {
+  // TODO: Configurable speed
```
jaskfla: Tiny, but this belongs as a GitHub issue, not committed

packages/backend/src/services/timelapseService.ts
jaskfla, line 1: I can’t really pretend to fully understand this module from just reading it

Readability seems fine to me (as much as ffmpeg can be readable 😂)

Do you reckon there’s any opportunity to break it up into smaller bits?

packages/backend/static/constant/timelapse-end-card.png
jaskfla, line 1: - Make the logo smaller; give it more breathing room than you think is necessary (i.e. in flex box parlance, increase the `gap`)
- Can afford to lower the opacity of the URL a bit more. 60% is what we do in a lot the GUI
- URL could do with a little letter-spacing
- Run it through a lossless compression

Or give me the source (Figma, is it?) and I can do the tweaks

jaskfla (changes requested):

packages/backend/src/routes/api/v1/canvas.ts
Rocked03: no... if anything it probably should instead be a map instead of a boolean... the file path "raw" generates can be subject to change in the future, it's just simply that the raw video generated uses the webm format

packages/backend/src/routes/api/v1/canvas.ts
Rocked03: raw is a 1:1 lossless timelapse with no added stuff (padding, end card, etc etc)

packages/backend/src/services/timelapseService.ts
Rocked03: will give it a go

packages/backend/static/constant/timelapse-end-card.png
Rocked03: i don't have edit access on our original figma so i made it in a personal file 🫣 

could do with a slightly different redesign in terms of the text layout anyways if you have suggestions... i'm not the most fond of the spacing i used but i couldn't really make it less awkward

packages/backend/src/routes/api/v1/canvas.ts
```diff
@@ -100,6 +102,62 @@ canvasRouter.get(
   },
 );
 
+canvasRouter.get(
+  "/:canvasId.mp4",
+  validate({
```
Rocked03, line 114: todo: create a new issue to implement ratelimiting and authentication and generate queuing and whatnot

just lock the endpoints to mod auth rn

S-Ou: [#774](https://github.com/project-blurple/Canvas-Web/issues/774)

packages/backend/src/routes/api/v1/frame.ts
Rocked03: #774

jaskfla (approved):
Neato, thanks for the clarifications 🙏

Still terrified by ffmpeg, but such is ffmpeg. Appreciate the inline comments to kinda make sense of it

stijnvdkolk (approved):

volcanofr:
🎉

PR #688, opened 2026-05-28
fix(frontend): select only on left click when panning

### LINKED ISSUE #642: Ignore middle click for select
We should try making it so perhaps only the left click will select when panning so you can middle click to pan and zoom interrupted

Rocked03:
Looks like we originally tried but it didn't work. Now you should be able to pan with all mouse buttons, but it will only select coords with left click (so you can pan with middle click without changing your selection)


jaskfla (approved):
Nice QoL improvement

Might suggest separately that we disallow panning with right-click, but can defer that

PR #690, opened 2026-05-29
fix(frontend): ‘Place pixel’ button stuck on cooldown with 1 s

jaskfla:


packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
jaskfla, line 1: Not stoked about the readability here, but I think that comes with this sort of lower-level timing-related code 😔

stijnvdkolk (approved):

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
stijnvdkolk: Could be worse :p I think its fine for now

PR #695, opened 2026-05-30
refactor(frontend): simplify `CanvasContext` state management

jaskfla:
The first of probably several PRs to thin down `<CanvasContext>`, moving as much state into URL as reasonable


packages/frontend/src/contexts/CanvasViewContext.tsx
```diff
@@ -53,12 +53,6 @@ export const CanvasViewProvider = ({ children }: CanvasViewProviderProps) => {
   const [zoom, setZoom] = useState(1);
   const containerRef = useRef<HTMLDivElement | null>(null);
   const [offset, setOffset] = useState(ORIGIN);
-  const [prevCanvasId, setPrevCanvasId] = useState(canvas.id);
-
-  if (prevCanvasId !== canvas.id) {
-    setSelectedCoords(null);
-    setPrevCanvasId(canvas.id);
-  }
```
jaskfla, line 61: I should’ve spotted this in review, but this was really a sign that we should have a completely new instance of the context when the canvas changes

Rocked03 (approved):

PR #696, opened 2026-05-30
refactor(frontend): dedupe `Socket["auth"]` logic

jaskfla:


packages/frontend/src/contexts/SelectedBoundsContext.tsx
jaskfla, line 1: Whoops, unrelated but I’ll leave it in

stijnvdkolk (approved):

PR #699, opened 2026-05-30
feat(frontend): keep current canvas state in URL

jaskfla:


packages/frontend/src/components/header/Header.tsx
```diff
@@ -61,12 +58,12 @@ const Logo = styled(Image)(
   `,
 );
 
-export default function Header({
-  isCanvasPage: isCanvasPage = false,
-}: HeaderProps) {
+export default function Header() {
+  const { canvas } = useCanvasContext();
```
jaskfla, line 62: Not a fan but leaving for now

packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx
```diff
@@ -518,7 +520,7 @@ function AdminCanvasTab() {
                   key={canvasItem.id}
                   disabled={isCanvasSelectionDisabled}
                   onClick={() => {
-                    setCanvas(canvasItem.id, false);
+                    router.push(`/canvas/${canvasItem.id}/admin/canvas`);
```
Comment (Rocked03) [inline packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx]: :eyes: do these need to be URI encoded?

packages/frontend/src/components/header/Nav.tsx
```diff
@@ -72,27 +72,34 @@ const StyledMenuItem = styled(MenuItem)`
 `;
 
 interface LinkInfo {
+  enabled?: boolean;
```
Rocked03: `available` or `accessible`? because the mod and admin dashboards are always enabled, they're just inaccessible  to most users

packages/frontend/next.config.mjs
```diff
@@ -2,6 +2,14 @@
 const nextConfig = {
   output: "standalone",
   reactCompiler: true,
+  async rewrites() {
+    return [
+      {
+        source: "/canvas/:canvasId/@me",
```
Rocked03, line 8: do we _need_ to keep it as `@me`?

Rocked03: slightly separate question, would we be interested in having it be possible to view other people's profiles? We originally decided against that, but as we go for more years and collect more stats, it might be something that people may like & then we ought to structure the pathing around that

packages/frontend/src/app/page.tsx
```diff
@@ -1,10 +1,7 @@
...
+import { fetchCanvasInfo } from "@/hooks/queries/serverFetch";
 
 export default async function Page() {
-  return (
-    <LayoutWithHeader isCanvasPage>
-      <Main />
-    </LayoutWithHeader>
-  );
+  const canvasInfo = await fetchCanvasInfo();
+  redirect(`/canvas/${canvasInfo.id}`);
```
Rocked03: imo i like the functionality of `https://canvas.projectblurple.com` defaulting to the main canvas without redirecting... makes it feel like you are actually in the main location

not sure how feasible that'd be without having the canvas id in the path tho

Rocked03 (approved):
i'm glad git recognises them as renames and not deletes & creates

looks good to me, just a few thoughts but nothing blocking

packages/frontend/next.config.mjs
volcanofr: Yes yes yes! Maybe be able to hide our own profile like Discord does with blocking people?
But then also hide user on pixel's history 'cause we can build data from that too?

packages/frontend/src/app/page.tsx
volcanofr: Slightly seperate, `/canvas` without a canvas id returns the 404 page and does not redirects to the main canvas.

Rocked03: > Slightly seperate, `/canvas` without a canvas id returns the 404 page and does not redirects to the main canvas.

I'd say that's intended behaviour.... we don't expect users to be at that path in the first place

packages/frontend/next.config.mjs
Rocked03: Those are details that would be decided at a later stage

jaskfla: > do we _need_ to keep it as `@me`?

No, could easily be `/me` (which is actually what it is today)—do you prefer that?

packages/frontend/src/app/page.tsx
jaskfla: > imo i like the functionality of https://canvas.projectblurple.com defaulting to the main canvas without redirecting... makes it feel like you are actually in the main location

Ditto but it’s not worth the state management headache

> Slightly seperate, `/canvas` without a canvas id returns the 404 page and does not redirects to the main canvas

Seems reasonable to redirect. Done ✅

packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx
Comment (jaskfla) [reply packages/frontend/src/app/canvas/[canvasId]/admin/canvas/page.tsx]: Not strictly but I much prefer to. Nice catch—thank you 🙏

89319a947261717997df95c464519bdd34c80800

packages/frontend/src/components/header/Nav.tsx
jaskfla: Went with `permitted`

PR #715, opened 2026-06-01
feat(frontend): eyedropper swatch

### LINKED ISSUE #701: Colour picker
Extra color option to select the matching colour that your reticle is currently above

Rocked03:
Adds an "eyedropper" swatch that shows the color of the currently selected pixel. Clicking on this eyedropper will set the selected color to that same color.

image


Rocked03:
when having 3 variations of "tan", this is well-needed accessibility lol

jaskfla:
> when having 3 variations of "tan", this is well-needed accessibility lol

Yeah, def solves a valid problem. Doesn’t really belong in main colours section, though

It shouldn’t look like a swatch, because it doesn’t behave like one. Without getting too in the weeds, I think moving the button either up or down (I reckon down) and making it a button-shaped button (and not a swatch) should do the trick well enough for now:

> Main colours
> 🟥🟪🟩🟧
> Partner colours
> 🟥🟪🟩🟧
> Current colours
> <kbd>🖌️ Select Marvel Red</kbd>

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -108,6 +109,33 @@ function isColorUnavailable(
...
+): PaletteColor | null {
+  if (!rgba) return null;
+
+  const exact = palette.find(
+    (color) =>
+      color.rgba[0] === rgba[0] &&
+      color.rgba[1] === rgba[1] &&
+      color.rgba[2] === rgba[2] &&
+      color.rgba[3] === rgba[3],
+  );
```
jaskfla: I reckon time to add `es-toolkit` as a dependency

```ts
import { isEqual } from 'es-toolkit`;
```

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -573,6 +575,18 @@ export default function CanvasView({
 
   const canUseFullscreen = useIsFullscreenAvailable();
 
+  const samplePixelColor = useCallback((point: Point): PixelColor | null => {
+    const ctx = offscreenCanvasRef.current?.getContext("2d");
+    if (!ctx) return null;
+
+    try {
+      const data = ctx.getImageData(point.x, point.y, 1, 1).data;
+      return [data[0], data[1], data[2], data[3]];
```
jaskfla, line 586: Is this different from `data`? If not, `data.slice(...)`?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1192,6 +1206,15 @@ export default function CanvasView({
...
   );
 
+  useEffect(() => {
+    if (!coords) {
+      setSelectedPixelColor(null);
+      return;
+    }
+
+    setSelectedPixelColor(samplePixelColor(coords));
+  }, [coords, samplePixelColor, setSelectedPixelColor]);
```
jaskfla, line 1218: Can’t `selectedPixelColor` be derived within the render? I don’t think the context even needs to have a setter in its value

packages/frontend/src/components/swatch/InteractiveSwatch.tsx
```diff
@@ -28,6 +31,11 @@ const StyledSwatch = styled(StaticSwatch, {
     background-clip: content-box;
     padding: 3px;
   }
+
+  &[aria-disabled="true"] {
+    cursor: not-allowed;
+    opacity: 60%;
+  }
```
jaskfla: Moot because of my main comment, but highlighting that static swatch should never be disabled; only the interactive one, where we would use the real `disabled` attribute anyway

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -315,19 +360,34 @@ function NamedPalette({ colors, isColorDisabled, name }: NamedPaletteProps) {
...
+              <EyedropperSwatch
+                aria-disabled={selectedPixelColorRgb === null}
+                paletteColor={{
+                  name: "Eyedropper",
+                  rgba: selectedPixelColorRgb ?? [0, 0, 0, 0],
+                }}
+                onClick={() => {
+                  playSound();
+                  onEyedropperSelect();
+                }}
```
jaskfla: div with onClick = bad

Rocked03: `EyedropperSwatch` extends `InteractiveSwatch`, and it's using the exact same pattern as it

packages/frontend/src/components/canvas/CanvasView.tsx
S-Ou: only feasible if `data.slice(0, 4) as unknown as PixelColor`

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
S-Ou: `InteractiveSwatch` has `type="button"`

packages/frontend/src/components/canvas/CanvasView.tsx
S-Ou: Not easily... if we derive it here in `CanvasView.tsx`, there's not exactly an easy way to feed it into the context without a setter. And since we use it in the `PlacePixelTab.tsx`, we need to get the data over somehow

jaskfla: > we use it in the `PlacePixelTab.tsx`

But is that necessary? Technically this reduces some renders, but CanvasView can sample the selected colour based on the currently selected pixel internally. This way data only flows from the context out, never back in. It really feels like we should be able to achieve this without the consumers ever feeding any data back 🤔

I forget what order the providers are set up in, but it might be that this should actually be its own `useCurrentPixelColor` hook

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
jaskfla: Ah, sorry, I could swear the diff showed it extending `StaticSwatch`, which I presumed was why you used `aria-disabled` instead of `disabled`

packages/frontend/src/components/canvas/CanvasView.tsx
Rocked03: mmm

That would require moving the `offscreenCanvasRef` into a context, currently it lives in `CanvasView.tsx`

jaskfla: Ah, let’s leave as is then; thanks for investigating

packages/frontend/package.json
```diff
@@ -53,7 +53,8 @@
     "next": "^16.2.6",
     "react": "^19.2.5",
     "react-dom": "^19.2.5",
-    "socket.io-client": "^4.8.3"
+    "socket.io-client": "^4.8.3",
+    "es-toolkit": "^1.0.0"
```
jaskfla:
```suggestion
    "es-toolkit": "^1.47.0"
```

packages/frontend/src/components/canvas/CanvasView.tsx
Rocked03: Could do in the future

The main Canvas ref lives in the canvas view context, so it wouldn't be new

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -276,16 +276,23 @@ export default function PlacePixelTab({
...
+            <NamedPalette colors={mainColors} name="Main colors" />
             <NamedPalette
               colors={partnerColors}
               isColorDisabled={isColorDisabled}
               name="Partner colors"
             />
+            <ActionPanelPrimitives.SectionHeading>
+              Tools
+            </ActionPanelPrimitives.SectionHeading>
+            <DynamicButton
```
jaskfla: This should take a `startIcon` prop, right?

`startIcon={<Pipette />}`

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -276,16 +276,23 @@ export default function PlacePixelTab({
...
+              alwaysShowColor
```
jaskfla: Drop this prop and make it non-dynamic

- Mostly because of planned button revamp; but also
- dedicating a prop to a single instance is a bit code-smelly. In this case, a `styled` version would be more appropriate

`<Button>` is a _highly_ reused component, and its props should reflect that

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -276,16 +276,23 @@ export default function PlacePixelTab({
...
-              onEyedropperSelect={() => setColor(selectedPixelColor)}
-            />
+            <NamedPalette colors={mainColors} name="Main colors" />
             <NamedPalette
               colors={partnerColors}
               isColorDisabled={isColorDisabled}
               name="Partner colors"
             />
+            <ActionPanelPrimitives.SectionHeading>
+              Tools
```
jaskfla: There’s only one tool, so this feels like premature generalisation

I’d suggest labelling this section “Current pixel”

jaskfla (changes requested):

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
S-Ou: I've switched it to using a button, but I've had to copy-paste over the dynamic text color code which should reasonably be generalised eventually

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
```diff
@@ -288,16 +321,21 @@ export default function PlacePixelTab({
...
+              Current pixel
             </ActionPanelPrimitives.SectionHeading>
-            <DynamicButton
-              alwaysShowColor
-              color={selectedPixelColor?.rgba}
+            <StyledButton
               disabled={selectedPixelColor === null}
               onClick={() => setColor(selectedPixelColor)}
+              startIcon={<Pipette />}
+              style={{
```
jaskfla: Ah, I had meant literally a regular button

jaskfla (approved):
Sorry, bitta unclear communication from me in my last review

Clarified in a comment, but pre-approving

packages/frontend/src/components/action-panel/tabs/place/PlacePixelTab.tsx
Rocked03: grumble grumble

Rocked03: Although, I still styled it because we need to somehow have the text color be dynamic against the background color, and you can't do that in `style={{}}`

jaskfla: Counterpoint: don’t need the colour. Clicking it selects the colour, and the label shows what colour is going to be selected

Longer-term solution is to _hint_ at the colour, perhaps by styling the icon; but the more we override with ad-hoc styles the more difficult the button overhaul is going to be

`DynamicButton` is a special case of `Button`; usage _should_ be rare and kept for where its effect pays off the most

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Hey, popping back in to gently remind that it’s really helpful to colocate code

This is the only usage of `samplePixelColor`, but it’s defined 500 lines away from here 🕵️ 

#740

PR #716, opened 2026-06-01
feat(frontend): canvas grid

### LINKED ISSUE #708: Thin toggleable grid on canvas
To aid visual placement. For visibility, colour should ideally be at least one of:
- chosen to carefully avoid palette, 
- multicoloured,
- dynamic, or
- selectable by user

Rocked03:
& canvas option button refactor

Toggleable within a button right below the fullscreen button

image


packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1354,6 +1283,12 @@ export default function CanvasView({
             style={{ minWidth: canvas.width, minHeight: canvas.height }}
           />
         </CanvasImageWrapper>
+
+        <CanvasGrid
+          zoom={zoom}
+          style={{ display: isGridVisible ? undefined : "none" }}
```
jaskfla: Can use the `hidden` global attribute here

```suggestion
          hidden={!isGridVisible}
```

packages/frontend/src/components/canvas/CanvasViewControls.tsx
```diff
@@ -0,0 +1,156 @@
...
+  PanelRightClose,
+  PanelRightOpen,
+} from "lucide-react";
+import { CANVAS_WRAPPER_CLASS_NAME } from "@/util";
+import { Button } from "../button";
+import VisuallyHidden from "../VisuallyHidden";
+
+const CanvasOptionsButtonColumn = styled("div", {
+  shouldForwardProp: (prop: string) =>
+    !["$isPanelVisible", "$isFullscreen"].includes(prop),
```
jaskfla, line 16: Double checking that `$isFullscreen` can’t be selected for with a native CSS selector?

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+  inset-inline-end: ${(p) =>
+    p.$isPanelVisible &&
+    p.$isFullscreen &&
+    css`
+      inset-inline-end: calc(
+        min(var(--action-panel-width), calc(100vi - 1rem))
+      );
+      inset-inline-end: calc(
+        min(var(--action-panel-width), calc(100dvi - 1rem))
+      );
```
jaskfla: Setting `inset-inline-end` to _almost_ 100% feels… odd?

Looks to me like [anchor positioning](https://web.dev/learn/css/anchor-positioning) is the way to go

packages/frontend/src/components/canvas/CanvasViewControls.tsx
```diff
@@ -0,0 +1,156 @@
...
+  gap: 0.5rem;
+  position: absolute;
+  z-index: 3;
+
+  > *:first-child {
+    #${CANVAS_WRAPPER_CLASS_NAME}:not(:fullscreen) &,
+    #${CANVAS_WRAPPER_CLASS_NAME}:not(:-webkit-full-screen) & {
+      border-radius: 0.5rem 1rem 0.5rem 0.5rem;
+    }
+  }
```
jaskfla, line 39: A bit confused by this selector. I’m pretty sure `> *` is implicitly processed as `& > *`, so basically this does `& > * #foo:not(bar) &`, which surely selects nothing? Worth double checking it’s doing what you expect

If it is, can still refactor it:

```suggestion
  > *:first-child #${CANVAS_WRAPPER_CLASS_NAME}:not(:fullscreen, :-webkit-full-screen) & {
    border-radius: 0.5rem 1rem 0.5rem 0.5rem;
  }
```

(This won’t pass formatting check)

Might also highlight here that the `_CLASS_NAME` is being used as an ID, not a class

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+    #${CANVAS_WRAPPER_CLASS_NAME}:not(:fullscreen) &,
+    #${CANVAS_WRAPPER_CLASS_NAME}:not(:-webkit-full-screen) & {
+      border-radius: 0.5rem 1rem 0.5rem 0.5rem;
+    }
+  }
+
+  ${({ theme }) => theme.breakpoints.down("md")} {
+    inset-block-start: 3.5rem;
+    opacity: 0.75;
+  }
```
jaskfla: I’d encourage flipping this around. Style for mobile first, and override on desktop

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+`;
+
+const StyledCanvasOptionButton = styled(Button)`
+  border-color: transparent;
+  border-radius: 0.5rem;
+  color: white;
+  min-width: auto;
```
jaskfla: logical props

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+  text-decoration: none;
+
+  transition-duration: var(--transition-duration-slow);
+  transition-property:
+    -webkit-backdrop-filter, backdrop-filter, border-color, box-shadow;
+  transition-timing-function: ease;
+
+  @media (hover: hover) and (pointer: fine) {
+    &:hover {
+      -webkit-backdrop-filter: blur(8px);
```
jaskfla: `styled` should fill in the vendor-prefixed properties for us, if I’m not mistaken?

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+      backdrop-filter: blur(8px);
+      border-color: inherit;
+      box-shadow: 0 0 10px oklch(0 0 0 / 25%);
```
jaskfla: I think this is incongruent with existing hover styles?

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const StyledCanvasOptionButton = styled(Button)`
+  border-color: transparent;
+  border-radius: 0.5rem;
+  color: white;
+  min-width: auto;
+  padding: 0.5rem;
+  text-decoration: none;
+
+  transition-duration: var(--transition-duration-slow);
```
jaskfla: This is for hover effect, right? Should be on the order of 80–150 ms. Keep animations shorter than you think, especially because `ease` is not a particularly aggressive curve

packages/frontend/src/components/canvas/CanvasViewControls.tsx
```diff
@@ -0,0 +1,156 @@
...
+  grid,
+}: CanvasOptionButtonsProps) {
+  return (
+    <CanvasOptionsButtonColumn
+      $isFullscreen={fullscreen.isActive}
+      $isPanelVisible={panel.isActive}
+    >
+      {fullscreen.isAvailable && (
+        <CanvasOptionButton
+          aria-pressed={fullscreen.isActive}
```
jaskfla: Hmmm, not sure if this feels right, because it creates these two states:

- “Enter fullscreen. Not pressed.”
- “Exit fullscreen. Pressed.”

I reckon safe to drop this attribute, because the button label meaningfully communicates what it does in each state

Ditto grid button

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1235,39 +1174,27 @@ export default function CanvasView({
...
+          toggle: useCallback(() => {
+            setFullscreenPanelVisible((visible) => !visible);
+          }, [setFullscreenPanelVisible]),
+        }}
+        grid={{
+          isActive: isGridVisible,
+          toggle: () => {
+            setIsGridVisible((visible) => !visible);
+          },
+        }}
```
jaskfla: - All of these object props will be reinstantiated every time `<CanvasView>` renders
- `canUseFullScreen` can probably be its own `useFullscreenIsAvailable()` hook
- I wonder if there’s a clearer name than `CanvasOptionButtons`. This feels like a ‘View’ menu. Maybe `CanvasViewControls`?
- Some of this logic can be done without JavaScript, e.g.:

```html
<input name="grid-visible" type="checkbox" />
<svg />
```

```css
input:not(:checked) ~ svg#grid {
  display: none;
}
```

packages/frontend/src/components/canvas/CanvasGrid.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface CanvasGridProps extends React.SVGProps<SVGSVGElement> {
+  hidden?: boolean;
+  zoom: number;
+}
+
+export function CanvasGrid({ zoom, ...props }: CanvasGridProps) {
+  // Fully eyeballed this :)
+  // The stroke width is thicker when zoomed out to keep the grid visible, but thins as you zoom.
+  const gridStrokeWidth = ((5 - Math.log(zoom)) / 5) * 0.05;
```
jaskfla, line 23: Looks neat. Yes, magic numbers, but easy to internalise and fully contained within a single component. No complaints.

packages/frontend/src/components/canvas/CanvasGrid.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <Grid xmlns="http://www.w3.org/2000/svg" {...props}>
+      <defs>
+        <pattern id="grid" width="1" height="1" patternUnits="userSpaceOnUse">
```
jaskfla: I do rather appreciate the readability of just hard-coding `"grid"`, but could be nice to use `useId()` to guarantee we never accidentally try to use the same ID somewhere else in the codebase

jaskfla (changes requested):
Simmered on it and quite happy with the SVG solution

Gotta apologise I’m doing this not on my personal computer so don’t have the codebase. Haven’t had a chance to pull the code and inspect the DOM, so my comments are based only from reading the diff. Don’t go _too_ far down rabbit holes; I know I throw a lot of new Web APIs at you. If you feel like something should be deferred that’s cool—just pop a comment in and move on 🙏

packages/frontend/src/components/canvas/CanvasViewControls.tsx
S-Ou: deferring, out of scope for this PR

packages/frontend/src/components/canvas/CanvasViewControls.tsx
S-Ou: deferring, out of scope for this PR

packages/frontend/src/components/canvas/CanvasViewControls.tsx
S-Ou: class name is correct yup

jaskfla: Yeah, no, it’s the “correct” symbol name, but it _should_ be renamed to ID

`#{myId}` and `.${myClass}` make sense; but not the other way around

packages/frontend/src/components/canvas/CanvasViewControls.tsx
jaskfla: I‘d push to remove the `aria-pressed` attributes. Two deletions and this comment is completely resolved

jaskfla (approved):
Coming back to pre-approve. I don’t think everything’s resolved yet, but I think we’ve discussed each one enough that I’m happy to unblock

Appreciate the efforts

packages/frontend/src/components/canvas/CanvasOptionButtons.tsx
S-Ou: messed around with anchor positioning and i couldn't get it to recognise the anchor 

switched to using `right` since we do have the `var(--action-panel-width)` and that's reliable

jaskfla: Logical inset is preferable to `right` if not anchoring to the action panel with anchor positing

I just meant that normally `left: 0` makes more sense than `right: 100%`

packages/frontend/src/components/canvas/CanvasViewControls.tsx
S-Ou: The selector is working as intended, it's picking the first child and then if the canvas wrapper isn't in fullscreen, then it has the specified border radius

S-Ou: I'm also deferring the rename though, it's a bit of a blurred line here but this PR's scope is really just the grid and this button refactor was only so that I could introduce a place to toggle the grid. Changes to how all the other buttons work and whatnot aren't really what this PR is for

jaskfla: Yeah, super fair call

packages/frontend/src/components/canvas/CanvasView.tsx
S-Ou: Well, I definitely tried switching to css-only

didn't like it :(

PR #718, opened 2026-06-03
feat: added shadow behind canvas

### LINKED ISSUE #185: Visual border around canvas
Visually difficult to see border if the pixel on the edge is the same colour as the background - some sort of visual border around the canvas so it's clear

Rocked03:
| Before | After |
| --- | --- |
| image | image |
| image | image |

closes #185

### DIFF
--- a/packages/frontend/src/components/canvas/CanvasView.tsx
+++ b/packages/frontend/src/components/canvas/CanvasView.tsx
@@ -244,7 +244,8 @@ const CanvasImageWrapper = styled("div", {
       filter: grayscale(80%);
     `}
 
-  box-shadow: 0 0 2rem 0 rgba(0 0 0 / 0.5); // No shadow behind the transparent pixels
+  box-shadow: 0 0 100px 0 rgba(0 0 0 / 0.15);
+  // box-shadow means the shadow only shows on the edges, and doesn't affect the backing color of the translucent pixels in the canvas
   position: relative;
 
   img {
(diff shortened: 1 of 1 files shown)


Rocked03:
keen for opinions about how strong this shadow should be

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -244,6 +244,7 @@ const CanvasImageWrapper = styled("div", {
       filter: grayscale(80%);
     `}
 
+  box-shadow: 0 0 2rem 0 rgba(0 0 0 / 0.5); // No shadow behind the transparent pixels
```
jaskfla: Consider increasing the blur (to like 60–100px) and decrease intensity to maybe 10–25%. Basically dial it down until you can’t quite tell if it’s even there, and then dial it back up a percentage point

And not sure what you mean by the comment here. Regardless, can remove it

jaskfla (approved):
Left some tweak suggestions, but happy to let you play with it and merge when you’re happy with it

packages/frontend/src/components/canvas/CanvasView.tsx
Rocked03: Sorry, comment wasn't clear-- the blank pixels are translucent and I don't want to mess with the background colour behind them, so a `box-shadow` only puts the shadow _around_ the div compared to a `filter: drop-shadow`

PR #719, opened 2026-06-03
feat: cloudflare turnstile

### LINKED ISSUE #720: Add CAPTCHA to pixel placing
sorry (not sorry) automated scripters :)

Rocked03:
# Token flow
- Makes token request on canvas page load
- Makes token request on pixel place
  - If the existing token is still valid, use it
    - [Tokens expire after 5 minutes](https://developers.cloudflare.com/turnstile/get-started/server-side-validation/#request-format)
  - Otherwise, generates a new one (handled by the `react-turnstile` package)
- After pixel place, reset the existing token and fetch a new one
- Rinse and repeat
# Notes
- Will only make token requests if:
  - Canvas is not locked
  - User is logged in
- Added flag to both frontend and backend `.env`s to toggle whether to use captcha. Both must be true in order for it to all work
- https://developers.cloudflare.com/turnstile/get-started/widget-management/dashboard/ <-- add `localhost` as the hostname for local dev
# Still need to do
image
https://www.cloudflare.com/en-gb/turnstile-privacy-policy/


packages/frontend/src/hooks/useTurnstileToken.tsx
```diff
@@ -0,0 +1,51 @@
+"use client";
+
+import type { TurnstileInstance } from "@marsidev/react-turnstile";
+import { Turnstile } from "@marsidev/react-turnstile";
```
stijnvdkolk:
```suggestion
import { Turnstile, type TurnstileInstance } from "@marsidev/react-turnstile";
```

jaskfla (approved):
Quick review for readability.
Only tiny things stood out—too small to even bring up

Nice addition

stijnvdkolk (approved):

packages/frontend/src/hooks/useTurnstileToken.tsx
stijnvdkolk: Wanted to suggest this too 💪

PR #733, opened 2026-06-07
feat: migrate main prisma to core folder

Rocked03:


packages/backend/package.json
```diff
@@ -38,9 +38,9 @@
     "check": "pnpm run format && pnpm run lint",
     "check:fix": "pnpm run format:fix && pnpm run lint:fix",
     "clean": "rm -rf -- build/ && pnpm purge",
-    "prisma:push": "prisma db push",
-    "prisma:seed": "prisma db seed --",
-    "prisma:migrate": "prisma migrate deploy",
+    "prisma:push": "prisma db push --config prisma/core/prisma.config.ts",
+    "prisma:seed": "prisma db seed --config prisma/core/prisma.config.ts --",
+    "prisma:migrate": "prisma migrate deploy --config prisma/core/prisma.config.ts",
```
Rocked03: should we do `prisma:core:push`?

stijnvdkolk: Guess that might be nice, but at the same time, do we still need the `prisma:push` that since we now are using `prisma migrate deploy`?

Rocked03: perhaps not, but we'd probably still want `prisma:core:seed` and `prisma:code:migrate`

stijnvdkolk: Yea

stijnvdkolk (approved):

PR #734, opened 2026-06-07
feat: added Open Telemetry trace logging

### LINKED ISSUE #393: Add Logging
As more and more functionality gets brought to Canvas Web, it would be good to keep comprehensive logging of all actions

- pixel placed
- canvas created
- canvas modified (name, lock)
- frame created
- frame modified
- added to blacklist
- removed from blacklist
etc.

Logging should obviously persist across sessions, but would also be good if accessible via GUI - look into available log viewer services

Should probably store logs in the database

Logs should also be easily exportable, as the hosting server is frequently deconstructed and reconstructed each year

Rocked03:
- Added Open Telemetry trace logging
  - This logs every HTTP request made, including exact timings of the internal processes
- Added Jaeger to collect OTEL traces and display them via its UI
- Added OpenSearch for persistent storage of the OTEL traces
- Added OAuth2-proxy for GitHub auth in front of the Jaeger UI

this PR is a bit of a mess, it was originally a nice tidy thing adding all the OTEL to the routes, but then it got messy trying to get Docker stuff to work and I _think_ it does but ig we'll find out once we load it up onto prod

---

image
image


docker-compose.yml
```diff
@@ -30,6 +32,55 @@ services:
...
+  opensearch:
+    image: opensearchproject/opensearch:2.9.0
+    environment:
+      - cluster.name=opensearch-cluster
+      - discovery.type=single-node
+      - OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
```
pumbas600: Is there a particular reason you've set memlock to unlimited?

Rocked03: not particularly, i'll adjust it

stijnvdkolk (commented):
Current backend code looks fine I think, we can probably iterate on top of this and improve/expand later if needed!

stijnvdkolk (approved):

PR #737, opened 2026-06-09
build(frontend): abolish barrel files

jaskfla:
To be honest, a bit of an experiment

I like the ergonomics of barrel files, but maybe this will help us recognise when we aren’t colocating code effectively

There’s a smidge of a performance improvement too (about 3%) to the `tsc` stage. Negligible since this is not a large codebase; but more importantly it’s dwarfed by the static page rendering (which I’d like to get rid of). In any case, this is not a meaningful deciding factor.


Rocked03 (approved):
ok

PR #739, opened 2026-06-10
refactor(frontend): `useEffectListener`-ify `<CanvasView>` listeners

jaskfla:


packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1327,9 +1303,7 @@ export default function CanvasView({
             style={{ minWidth: canvas.width, minHeight: canvas.height }}
           />
         </CanvasImageWrapper>
-
         <CanvasGrid zoom={zoom} hidden={!isGridVisible} />
-
```
Rocked03, line 1329: readability :(

Rocked03 (approved):
no particular thoughts

packages/frontend/src/components/canvas/CanvasView.tsx
jaskfla: Yes, but the fix is to break it into smaller, composable parts!

PR #746, opened 2026-06-10
feat(backend/nest): implement configs and basic utils

### LINKED ISSUE #766: Backend NestJS: Config setup

stijnvdkolk:


packages/backend-nest/.swcrc
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://swc.rs/schema.json",
+  "sourceMaps": true,
+  "module": { "type": "commonjs" },
+  "jsc": {
+    "target": "es2023",
```
jaskfla: Is esnext an option?

packages/backend-nest/src/config/app.config.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const paths = {
+    root: path.resolve(),
+    canvases: path.resolve("static", "canvas"),
+  };
+
+  if (!fs.existsSync(paths.canvases)) {
+    console.debug(`Creating canvases directory at ${paths.canvases}`);
+    fs.mkdirSync(paths.canvases, { recursive: true });
+  }
```
jaskfla: Isn’t `mkdir` with `recursive: true` basically a no-op if it already `exists`?

Fine as-is, but I would’ve thought the `if` is redundant 🤔

packages/backend-nest/src/config/config.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+  DATABASE_URL: "postgresql://test:test@localhost:5432/test",
+  DISCORD_CLIENT_ID: "client-id",
+  DISCORD_CLIENT_SECRET: "client-secret",
+};
+
+describe("config", () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    process.env = { ...REQUIRED_ENV };
```
jaskfla: Er, will this override the rest of `process.env`, or is Node smart about it? Think `NODE_ENV`, `NO_COLOR`, `TERM`, etc

The behaviour feels a bit opaque to me, but if Node has special handling that’s clearly documented I’m fine with it. If not, one of these may be clearer:

- `process.env = { ...process.env, ...REQUIRED_ENV }`
- `Object.assign(process.env, REQUIRED_ENV)`
- Iterate over `REQUIRED_ENV` and just use `=` assignment to `process.env`

packages/backend-nest/src/config/config.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+  });
```
jaskfla: Similar comment to the previous one: does this work???

Assuming `process.env` behaves like a regular JS object, I’m pretty sure this the `afterEach` will do nothing because it will reference the same object?

packages/backend-nest/test/app.e2e-spec.ts
```diff
@@ -1,29 +1,45 @@
...
+  it("sends credentialed CORS headers for the frontend origin", async () => {
+    const response = await request(app.getHttpServer())
       .get("/")
-      .expect(200)
-      .expect("Hello World!");
+      .set("Origin", "http://localhost:3000");
+
+    expect(response.headers["access-control-allow-origin"]).toBe(
+      "http://localhost:3000",
+    );
```
jaskfla, line 36: Is this a `Response` object? If so, I’d suggest replacing `headers["access-control-allow-origin"]` with `headers.get("access-control-allow-origin")`

packages/backend-nest/tsconfig.json
```diff
@@ -20,10 +20,8 @@
     },
     "incremental": true,
     "skipLibCheck": true,
-    "strictNullChecks": true,
+    "strict": true,
     "forceConsistentCasingInFileNames": true,
-    "noImplicitAny": false,
```
jaskfla, line 25: 😨

Why? 🫣

jaskfla (commented):
Concerned about removal of `noImplicitAny`. Otherwise just minor comments

Pre-approving, though

jaskfla (approved):
Totally didn’t forget to approve when I said I was pre-approving

packages/backend-nest/src/config/tracing.config.ts
```diff
@@ -0,0 +1,15 @@
+import type { ConfigType } from "@nestjs/config";
+import { registerAs } from "@nestjs/config";
+import { ConfigNamespace } from "./config-namespace";
+import { validateEnv } from "./env";
+
+export const tracingConfig = registerAs(ConfigNamespace.Tracing, () => {
```
Rocked03: `telemetry` or `observability` would be a better name here

packages/backend-nest/test/app.e2e-spec.ts
```diff
@@ -1,29 +1,45 @@
...
+    expect(response.headers["access-control-allow-credentials"]).toBe("true");
   });
 
-  afterEach(async () => {
-    await app.close();
+  it("serialises BigInt values as JSON strings", () => {
+    expect(JSON.stringify({ id: 204446594050784000n })).toBe(
```
Rocked03: try an ID that's less resilient to rounding?

Rocked03 (approved):
happy with it so far, also happy to just let things keep moving forward since you'll still be working on this stuff in later PRs so small things can be tidied up as you go

packages/backend-nest/src/config/config.spec.ts
stijnvdkolk: This will overwrite, must've missed this.

packages/backend-nest/src/config/config.spec.ts
stijnvdkolk: Should work, but will change it just in case

packages/backend-nest/test/app.e2e-spec.ts
stijnvdkolk: It's a `Response` object but headers is of type `{ [index: string]: string }`

packages/backend-nest/tsconfig.json
stijnvdkolk: Having `strict` on true this is also true by default

PR #748, opened 2026-06-10
feat(backend/nest): add database setup with Prisma and Kysely

### LINKED ISSUE #751: Backend NestJS: Prisma & Kysely setup

stijnvdkolk:


jaskfla (commented):
Kind of just a cursory review at a high level, not a line-by-line review. Nothing jumps out as alarming to me

Happy to stamp it; but might leave the Official Okie-Dokie to Sam/Josh

Rocked03 (approved):
to be honest, most of this stuff I only really understand the what and not necessarily the why

but it looks pretty reasonable to me, I'm fine to push this through

although it looks like you'll need to do a bit of rejigging to include the snapshots prisma now too?

PR #773, opened 2026-06-12
feat(backend/nest): add canvas module

### LINKED ISSUE #754: Backend NestJS: Canvas Module

stijnvdkolk:


packages/backend-nest/src/canvas/canvas-cache.service.spec.ts
```diff
@@ -0,0 +1,287 @@
+import fs from "node:fs";
```
jaskfla: Do we get away importing `node:fs/promises` as `fs`?

I know you do use `existsSync`, but we can make the testing library `.resolve` it, right? (I think?)

packages/backend-nest/src/canvas/canvas-cache.service.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+        width: 4,
+        height: 4,
+      });
+
+      expect(await sharp(canvas4xPath).metadata()).toMatchObject({
+        width: 8,
+        height: 8,
+      });
+
+      for (const scale of [1, 2, 4] as const) {
```
jaskfla: Should this magic array be `Object.keys(canvasPaths)`? (Or a const array with the supported resolutions?)

packages/backend-nest/src/canvas/canvas.controller.ts
```diff
@@ -0,0 +1,358 @@
...
+    );
+
+    return {
+      cooldownEndTime: cooldownEndTime ?? undefined,
+    };
+  }
+
+  @Get(":canvasId@:scale.png")
+  @ApiOperation({
+    summary: "PNG of a canvas at scale (1/2/4×), with optional crop",
```
jaskfla, line 162: 🤓

Screenshot 2026-06-12 at 20 16 33

packages/backend-nest/src/canvas/canvas.service.ts
```diff
@@ -0,0 +1,413 @@
...
+        x,
+        y,
+        colorId,
+      }),
+    );
+
+    const lowestX = Math.min(...entries.map(({ x }) => x));
+    const lowestY = Math.min(...entries.map(({ y }) => y));
+    const highestX = Math.max(...entries.map(({ x }) => x));
+    const highestY = Math.max(...entries.map(({ y }) => y));
```
jaskfla: Doing double work here! If you wanna stick with `min(...[...]])`, create the array of `x`’s and array of `y`’s first

But, actually, because `entries` is potentially large, I’d actually prefer using a reducer here:

```js
const minX = entries.reduce((min, entry) => Math.min(min, entry.x), entries[0].x)
```

Or even more performant (but less readable):

```js
let minX, minY, maxX, maxY;
for (const { x, y } of entries) {
  minX = Math.min(minX, x);
  minY = Math.min(...);
  // …and so on
}
```

For this last approach I would consider a helper function or IIFE

packages/backend-nest/src/canvas/canvas.service.ts
```diff
@@ -0,0 +1,413 @@
...
+      (colorId) => !colors.some((color) => color.id === colorId),
+    );
+
+    if (invalidColorIds.length > 0) {
+      throw new Error(
+        `Data contains color IDs that are not in the event palette: ${invalidColorIds.join(
+          ", ",
+        )}`,
+      );
+    }
```
jaskfla, line 324:
```suggestion
    if (invalidColorIds.length > 0) {
      const formatter = new Intl.ListFormat();
      throw new Error(
        `Data contains color IDs that are not in the event palette: ${formatter.format(invalidColorIds)}`,
      );
    }
```

Defaults to conjunction (i.e. “and”), so no options needed for the constructor call

packages/backend-nest/src/canvas/export.service.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+  png: Buffer,
+): Promise<{ width: number; height: number; pixels: number[][] }> {
+  const { data, info } = await sharp(png)
+    .ensureAlpha()
+    .raw()
+    .toBuffer({ resolveWithObject: true });
+
+  const pixels: number[][] = [];
+  for (let index = 0; index < data.length; index += 4) {
+    pixels.push([...data.subarray(index, index + 4)]);
```
jaskfla, line 48: Nice use of `push(...[])`

packages/backend-nest/src/canvas/export.service.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      x1: 2,
+      y1: 2,
+    });
+
+    const decoded = await decodePng(await streamToBuffer(stream));
+
+    expect(decoded).toStrictEqual({
+      width: 2,
+      height: 2,
+      pixels: [BLANK, BLURPLE, RED, BLANK].map((color) => [...color]),
```
jaskfla: Am I reading correctly? I think this is equivalent

```suggestion
      pixels: [BLANK, BLURPLE, RED, BLANK].flat(),
```

Same in line 141

packages/backend-nest/src/canvas/export.service.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      x1: 2,
+      y1: 1,
+    });
+
+    const decoded = await decodePng(await streamToBuffer(stream));
+
+    expect(decoded).toStrictEqual({
+      width: 1,
+      height: 1,
+      pixels: [[...BLURPLE]],
```
jaskfla: Redundant spread; `toStrictEqual` shouldn’t have any side effects

```suggestion
      pixels: [BLURPLE],
```

packages/backend-nest/src/canvas/export.service.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    return scale === 1 ?
+        image.png()
+      : image
+          .resize({
+            width: unlockedCanvas.width * scale,
+            height: unlockedCanvas.height * scale,
+            kernel: sharp.kernel.nearest,
+          })
+          .png();
```
jaskfla: Set additional options:

```js
{
  xres: 72 * scale,
  yres, 72 * scale,
  resolutionUnit: 'inch',
}
```

- Ideally dedupe the multiplication, though. Magic number `72` is fine; it’s basically universally understood to be the “normal” resolution
- Ideally also apply this metadata to all image files we create
- And chain `.withIccProfile('srgb')`

jaskfla (changes requested):

packages/backend-nest/src/canvas/canvas-cache.service.spec.ts
```diff
@@ -47,7 +53,7 @@ describe("CanvasCacheService", () => {
 
   afterEach(async () => {
     await moduleRef.close();
-    await fs.promises.rm(canvasesPath, { recursive: true, force: true });
+    await fs.rm(canvasesPath, { recursive: true, force: true });
```
jaskfla: Negligible change, but partially in the interest of training better patterns for whatever models slurp this up in their training data, these two are safe to “parallelise”:

```diff
- await fs.promises.rm(canvasesPath, { recursive: true, force: true });
- await fs.rm(canvasesPath, { recursive: true, force: true });
+ await Promise.all([
+   moduleRef.close(),
+   fs.rm(canvasesPath, { recursive: true, force: true }),
+ );
```

packages/backend-nest/src/canvas/canvas-cache.service.ts
```diff
@@ -78,6 +78,10 @@ export class CanvasCacheService implements OnApplicationBootstrap {
     return `blurple-canvas__${canvasId}__${isLocked ? "locked" : Date.now()}${boundsSuffix}${scaleSuffix}.png`;
   }
 
+  withPngMetadata(image: sharp.Sharp, scale: CanvasExportScale): sharp.Sharp {
+    return image.withMetadata({ density: 72 * scale }).withIccProfile("srgb");
+  }
```
jaskfla: Looks like this can be static?

Biome might then (reasonably) flag the use of `this.withPngMetadata` below, but if you stage this file and run `pnpm lint-staged` before committing it will resolve itself

jaskfla (approved):
Lovely. Two details, both minuscule ✅

PR #784, opened 2026-06-15
feat(backend/nest): implement palette & notice modules

### LINKED ISSUE #758: Backend NestJS: Palette Module

stijnvdkolk:


packages/backend-nest/src/notice/notice.service.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+            header: "Active high",
+            priority: 0,
+            startAt: new Date(now - HOUR),
+          },
+          // Active, low priority
+          {
+            id: 2,
+            type: "warning",
+            header: "Active low",
+            priority: 5,
```
Rocked03, line 46: double check with the front-end, but i think high number = higher priority

we should probably document that somewhere lol

Rocked03: https://github.com/project-blurple/Canvas-Web/blob/acf19f918720aa113f19492d0633a2b1780b4eb4/packages/frontend/src/components/notices/Notices.tsx#L64-L66

packages/backend-nest/src/palette/palette.controller.ts
```diff
@@ -0,0 +1,165 @@
...
+      metadata: body,
+    });
+
+    return { message: "Color created" };
+  }
+
+  @Put(":colorId")
+  @RequiresCanvasAdmin()
+  @ApiOperation({ summary: "Edit a colour (admin)" })
+  async editColor(
```
Rocked03, line 83: how come we don't have a `@HttpCode()` here?

packages/backend-nest/src/palette/palette.service.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export interface EditColorParams {
+  colorId: PaletteColor["id"];
+  data: CreateColorParams;
+}
+
+export interface AssignColorToEventParams {
+  colorId: PaletteColor["id"];
+  eventId: BlurpleEvent["id"];
+  guildId: bigint;
```
Rocked03: not derivable? I know our objects currently type our snowflakes to strings (for json/http use) but surely there's something we can do about making that work

packages/backend-nest/src/palette/palette.service.ts
```diff
@@ -0,0 +1,189 @@
...
+   * colours for that event. With no event match, only global colours return.
+   */
+  async getEventPalette(
+    eventId: number,
+    allColors = false,
+  ): Promise<PaletteColor[]> {
+    const where =
+      allColors ? undefined : (
+        {
+          OR: [{ global: true }, { participations: { some: { eventId } } }],
```
Rocked03, line 80: no way the linter hasn't had a fit over this all being on one line

Rocked03 (approved):
a few small thoughts but nothing blocking

packages/backend-nest/src/palette/palette.service.ts
stijnvdkolk: It formatted it for me on 1 line 😅

packages/backend-nest/src/palette/palette.controller.ts
stijnvdkolk: It defaults to 200 always

PR #786, opened 2026-06-15
feat(backend/nest): implement frame module

### LINKED ISSUE #759: Backend NestJS: Frame Module

stijnvdkolk:


packages/backend-nest/src/frame/frame.service.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+    administrator: true,
+    manage_guild: false,
+  })),
+};
+
+const user: DiscordUserProfile = {
+  id: "1",
+  username: "test_user_1",
+  profilePictureUrl: "https://example.com/avatar1.png",
+};
```
Rocked03: we have these constants in multiple places, would it be appropriate to have a shared helpers class or something?

packages/types/src/models/frame.models.ts
```diff
@@ -59,12 +59,26 @@ const frameBoundsRefiner = (
   }
 };
 
+/** Orders bounds so (x0, y0) is the top-left and (x1, y1) the bottom-right. */
+const normalizeFrameBounds = <T extends z.infer<typeof FrameBoundsModel>>(
```
Rocked03: these are also used for stuff like history, would it be better to have a more generic name?

Rocked03 (approved):
a few thoughts but looks fine to me

PR #793, opened 2026-06-19
feat(frontend): frame UI redesign

### LINKED ISSUE #398: Export whole canvas / frames as images

### LINKED ISSUE #776: Expanded frame info & stats
a few ideas
- clicking on a frame replaces the whole tab content with info about the frame
- info pane includes stats about the frame
  - pixels placed
  - leaderboard
    - filter by colour
- edit button moved into here
- download image and download timelapse buttons here
- panning no longer defocuses frame, instead have to use back button
  - but potentially a 🎯 button that resets view back to the frame

Rocked03:
- Created a new frame "details" panel
  - Moved the "copy link" and "edit" buttons to here
  - Moved the details into here
  - New "refocus" button
  - Space to add frame statistics and leaderboard (future PR)
  - Space to add more download buttons (timelapse and stats)
  - Frame stays selected until "details" panel exited (instead of immediately when changing pan/zoom)
- Migrated `selectedFrame` state to URL search params
  - Panel state between list and details is now dependent on the selected frame
- Frame tab automatically opens when loading URL with frame search param
- Canvas view now uses `glidePan` when panning to newly selected frame


packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,291 @@
...
+        toast.success("Copied frame link");
+      }}
+    >
+      <Link />
+    </BasicButton>
+  );
+}
+
+function DetailsCard({ frame }: { frame: Frame }) {
+  const frameSize = [frame.x1 - frame.x0, frame.y1 - frame.y0];
```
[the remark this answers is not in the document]
Rocked03: this ended up being a bigger thing than I could fix here, so it's a separate PR #974

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
```diff
@@ -31,16 +33,22 @@ export default function FramesTab({
...
+    FramePanelMode.List,
   );
 
   useEffect(() => {
-    setTabsLocked(activePanel !== FramePanelMode.Info);
+    setTabsLocked(
+      activePanel !== FramePanelMode.List &&
+        activePanel !== FramePanelMode.Details,
+    );
   }, [activePanel, setTabsLocked]);
```
jaskfla, line 44: Flagging code smell. We pass a setter _somewhat_ commonly, but child setting state in parent is a bit fishy

If `activePanel` and `tabsLocked` existed in the same scope, `tabsLocked` could be derived within the render state. This Effect looks like a hack 🧐

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1023,10 +991,62 @@ export default function CanvasView({
       // Remember the movement so the canvas can keep gliding once released.
       lastPanVelocityRef.current = offsetDelta;
       updateOffset(multiplyPoint(offsetDelta, zoomRef.current));
+    },
+    [updateOffset],
+  );
+
+  const animateToFrame = useCallback(
```
jaskfla, line 999: This should respect `prefers-reduced-motion`. I think @stijnvdkolk made a function or hook for it somewhere?

packages/frontend/src/components/canvas/CanvasView.tsx
```diff
@@ -1023,10 +991,62 @@ export default function CanvasView({
...
+      setZoom,
+      setOffset,
+      glidePan,
+      setCoords,
+    ],
+  );
+
+  useEffect(() => {
+    focusOnFrameRef.current = animateToFrame;
+  }, [animateToFrame, focusOnFrameRef]);
```
jaskfla: How come there’s both a ref and callback? Are they not redundant? 🤨

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+  font-weight: 900;
+  line-height: 1.1;
+`;
+
+const DetailsTable = styled("table")`
+  max-inline-size: 100%;
+  width: 100%;
+
+  tr:not(:last-child) > td {
+    padding-bottom: 0.75rem;
```
jaskfla:
```suggestion
    padding-block-end: 0.75rem;
```

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+  width: 2.5rem;
+  white-space: nowrap;
+`;
+
+const TableCellContent = styled("td")`
+  font-size: 1.125rem;
+  height: 100%;
+  padding: 0;
+
+  > div {
```
jaskfla: Might be worth making this `div` a `styled` component for readability

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+    align-items: center;
+    display: flex;
+    gap: 0.125rem;
+    height: 100%;
+  }
+
+  code {
+    font-size: 1rem;
```
jaskfla:
```suggestion
    font-size: 0.85em;
```

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+const ControlButtonRow = styled("div")`
+  background: transparent;
+  display: flex;
+  gap: 0.5rem;
+  padding: 0;
+
+  > * {
+    flex: 1;
+  }
```
jaskfla, line 81: Pretty sure what you want is a grid?

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+`;
+
+function userCanEditFrame(user: DiscordUserProfile, frame: Frame): boolean {
+  switch (frame.owner.type) {
+    case FrameOwnerType.System:
+      return false;
```
jaskfla: Can move this down, right above `default` case

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+    );
+
+    if (frame.owner.type === FrameOwnerType.System) {
+      const baseUrl = `${config.apiUrl}/api/v1/canvas/${encodeURIComponent(canvas.id)}/frame/${encodeURIComponent(frame.id)}@${scale}.png`;
+
+      const isWholeCanvas =
+        frame.x0 <= 0 &&
+        frame.y0 <= 0 &&
+        frame.x1 >= canvas.width - 1 &&
+        frame.y1 >= canvas.height - 1;
```
jaskfla, line 122: This is done in a few places now, right? Could be worth making a helper like you did with `calculateScale`

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+        <DetailsTable>
+          <tbody>
+            <tr>
+              <TableCellIcon>{ownerInfo.icon}</TableCellIcon>
+              <TableCellContent>{ownerInfo.label}</TableCellContent>
+            </tr>
+            <tr>
+              <TableCellIcon>
+                <LucideFrame />
+              </TableCellIcon>
```
jaskfla: - Remember to put an accessible label in these header cells
- Drop the opacity of the icon by a smidge (like 1–3%); they currently look “heavier” than the text. Or could even just make them 60% if you think grey could look good

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+const DetailsTable = styled("table")`
+  max-inline-size: 100%;
+  width: 100%;
+
+  tr:not(:last-child) > td {
+    padding-bottom: 0.75rem;
+  }
+`;
+
+const TableCellIcon = styled("td")`
```
jaskfla: These are being used with `th` semantics, not `td`

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+    height: 100%;
+  }
+
+  code {
+    font-size: 1rem;
+  }
+`;
+
+const ControlButtonRow = styled("div")`
+  background: transparent;
```
jaskfla, line 74: This is the default, as is 0 padding

Should be safe to remove unless there’s a leaky selector in an ancestor

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+              <TableCellContent>{ownerInfo.label}</TableCellContent>
+            </tr>
+            <tr>
+              <TableCellIcon>
+                <LucideFrame />
+              </TableCellIcon>
+              <TableCellContent>
+                <div>
+                  {frameSize[0]} <X size={16} /> {frameSize[1]}
+                </div>
```
jaskfla: This `<div>` might be redundant, right? Set `display: inline` and `vertical-align` to something sensible on the icon and this cell should render more sensibly

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+          </BasicButton>
+          {userHasPermsToEditSelectedFrame && (
+            <BasicButton
+              aria-label="Edit frame"
+              onClick={() => setActivePanel(FramePanelMode.Edit)}
+            >
+              <SquarePen />
+            </BasicButton>
+          )}
+        </ControlButtonRow>
```
jaskfla, line 302: Best to find a way to make the text labels visible. This looks great in a screenshot, but takes creates nontrivial burden on the user to learn the icons’ meanings

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -42,6 +42,19 @@ import { BasicButton, Button, DestructiveButton } from "../button";
 import Dialog from "../Dialog";
 import { drawSourceRectToCanvas, PreviewCanvas } from "./FramePreview";
 
+const StyledForm = styled("form")`
+  background: transparent;
```
jaskfla: - Is this not the default?
- Prefer `background-color` longhand

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -42,6 +42,19 @@ import { BasicButton, Button, DestructiveButton } from "../button";
...
+  display: block flex;
+  flex-direction: column;
+  gap: 0.5rem;
```
jaskfla: Is flexbox a meaningful improvement from regular flow layout?

Perhaps more idiomatic to keep regular layout flow, and set “gap” with the lobotomised owl selector:

```css
> * + * {
  margin-block-start: 0.5rem;
}
```

packages/frontend/src/components/frames/FrameEditPanel.tsx
```diff
@@ -115,6 +128,12 @@ const ButtonRow = styled("div")`
   flex-direction: row-reverse;
   gap: 0.5em;
   inline-size: 100%;
+  background: transparent;
+  padding: 0;
+
+  > * {
+    flex: 1;
+  }
```
jaskfla, line 131: Again confirming that this layout isn’t a flexbox trying to emulate grid behaviour?

packages/frontend/src/components/frames/FrameListPanel.tsx
```diff
@@ -0,0 +1,81 @@
...
+  transform: translateY(0);
+  transition-duration: var(--transition-duration-slow);
+  transition-property: grid-template-rows, transform;
+  transition-timing-function: var(--ease-out-cubic);
+
+  @container --frame-tabpanel (height < 30rem) {
+    grid-template-rows: 0fr;
+    pointer-events: none;
+    transform: translateY(3rem);
+  }
```
jaskfla, line 26: Ah, bit of a pain point using CSS “variables” in JS is that IDEs can’t traverse the rendered cascade. I think a nice rule to keep is to make sure that (for non-global custom CSS properties) we use them in the same JS module where they’re defined.

In this case, define a `styled(FramesListPanel)` in `FramesListTab.jsx` and define the `@container` query style there. Otherwise it can be a bit of a mission to code trace

packages/frontend/src/components/frames/FrameListPanel.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  > * {
+    min-height: 0;
+  }
```
jaskfla, line 30: This breaks the browser’s built-in protection from layout degradation in small viewports or larger font sizes, and causes truncation and/or overlap instead of (relatively) gracefully reflowing

If possible, remove

packages/frontend/src/contexts/SelectedFrameContext.tsx
```diff
@@ -1,41 +1,86 @@
...
-    </SelectedFrameContext.Provider>
+/**
+ * Hook for managing selected frame via URL search params.
+ * Frame selection is stored entirely in the URL (f parameter), not in state.
+ * This mirrors the pattern used for canvas ID and viewport params.
+ *
+ * Optimized to show frame data immediately when selected (from preloaded data in FrameList)
+ * while the URL updates asynchronously in the background, eliminating perceived latency.
+ */
+export function useSelectedFrameContext(): UseSelectedFrameReturn {
```
jaskfla: Not a context anymore 👀; `useSelectedFrame` should do just fine

And ditch the header comment:
- the consumer shouldn’t _really_ need to care what the implementation is; and
- this comment explains what should be self-documenting in the function definition

packages/frontend/src/contexts/SelectedFrameContext.tsx
```diff
@@ -1,41 +1,86 @@
...
+  const router = useRouter();
+  const searchParams = useSearchParams();
+  const canvasParams = useCanvasSearchParams();
+  const { setCurrentTab } = useActionPanelContext();
+
+  // Track optimistic frame for instant UI updates
+  const [optimisticFrame, setOptimisticFrame] = useState<Frame | null>(null);
+  const optimisticFrameIdRef = useRef<string | null>(null);
```
jaskfla: Why both? 🤨

packages/frontend/src/contexts/SelectedFrameContext.tsx
```diff
@@ -1,41 +1,86 @@
...
+
+  // Fetch frame data based on frameId from URL (for page loads/refreshes)
+  const { data: urlFrame = null, isLoading } = useFrameById({
+    frameId: canvasParams.frameId ?? undefined,
+  });
+
+  // Clear optimistic frame once URL-fetched frame arrives and matches
+  useEffect(() => {
```
jaskfla: I reckon the info from this comment can be encoded in a sensible Effect function name

packages/frontend/src/contexts/SelectedFrameContext.tsx
```diff
@@ -1,41 +1,86 @@
...
-
-  return (
-    <SelectedFrameContext.Provider
-      value={{ frame: selectedFrame, setFrame: setSelectedFrame }}
-    >
-      {children}
-    </SelectedFrameContext.Provider>
+/**
+ * Hook for managing selected frame via URL search params.
+ * Frame selection is stored entirely in the URL (f parameter), not in state.
```
jaskfla: Raised offline perhaps a bigger discussion about whether a path param might be more appropriate than a search param

This PR is being done a bit of disservice by existing tech debt, where the current tab currently exists only “inside” React, blocking us from `/canvas/:canvasId/frame/:frameId`, where `frame` is the tab key

packages/frontend/src/contexts/CanvasContext.tsx
```diff
@@ -96,15 +95,13 @@ export const CanvasProvider = ({
...
-            </CanvasViewProvider>
-          </ActionPanelProvider>
-        </SelectedFrameProvider>
+        <ActionPanelProvider>
+          <CanvasViewProvider key={activeCanvas.id}>
+            <SelectedBoundsProvider key={activeCanvas.id}>
+              {children}
+            </SelectedBoundsProvider>
+          </CanvasViewProvider>
+        </ActionPanelProvider>
```
jaskfla, line 104: Heck yeah 😎

packages/frontend/src/contexts/SelectedFrameContext.tsx
```diff
@@ -1,41 +1,86 @@
...
+
+      // Switch to frame tab when a frame is selected
+      if (newFrame) {
+        setCurrentTab("frame");
+      }
+
+      // Update URL asynchronously (non-blocking)
+      const frameId = newFrame?.id ?? null;
+      const newUrl = createUrlWithFrameUpdate(searchParams, frameId);
+      router.replace(newUrl);
```
jaskfla: Replace? 🤨

Feel like we’re throwing away free behaviour here, where hitting the browser back button can refocus you to the last selected frame

packages/frontend/src/util/searchParams.ts
```diff
@@ -152,3 +152,51 @@ export function extractAllSearchParamsFromRecord(
...
+  }
+
+  // Update or remove the target parameter
+  const canonical = SEARCH_PARAM_KEYS[paramKey].canonical;
+
+  // Delete all variants (canonical + aliases) to ensure clean removal
+  const variants = getSearchParamVariants(paramKey);
+  for (const variant of variants) {
+    url.searchParams.delete(variant);
+  }
```
jaskfla, line 184: I wonder if we’ve painted ourselves into a corner of having to support aliases forever? Do we get by with _just_ canonical param keys?

Particularly, as you’ve identified, we need to be smart about what happens if both are provided for some reason

Could be convinced either way; the ergonomics are nice 🤷

jaskfla (changes requested):
Lotsa small quibbles, but really the big one is whether we should actually _first_ move the current action panel tab into the URL as a path segment first, so that we can get this change _right_ the first time around

packages/frontend/src/components/action-panel/tabs/FramesTab.tsx
S-Ou: hm, the thing is the activePanel is scoped to the FramesTab whereas the tabsLocked belongs to the ActionPanel parent

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
S-Ou: https://github.com/project-blurple/Canvas-Web/blob/b2f0f3433225fcc02b57b3fc795c9223e7885c63/packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx#L11-L15

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
S-Ou: Nope

The number of buttons can vary, I just want them to be equally spaced. Grid requires knowing in advance how many we expect in there

packages/frontend/src/components/frames/FrameEditPanel.tsx
S-Ou: https://github.com/project-blurple/Canvas-Web/blob/b2f0f3433225fcc02b57b3fc795c9223e7885c63/packages/frontend/src/components/action-panel/tabs/ActionPanelTabBody.tsx#L11-L15

packages/frontend/src/components/frames/FrameEditPanel.tsx
S-Ou: Need it to dynamically adapt to different amounts of children

packages/frontend/src/components/frames/FrameListPanel.tsx
S-Ou: sure, although I'm pretty sure this was your doing 👀

packages/frontend/src/util/searchParams.ts
S-Ou: idk, I like the idea of aliases? they're not necessary but it does provide a readable (and transparent) option

packages/frontend/src/contexts/SelectedFrameContext.tsx
S-Ou: was trying to do something fancy about avoiding rerenders but it's not really necessary, have gotten rid of the ref

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
S-Ou: Currently, just here

packages/frontend/src/components/frames/FrameListPanel.tsx
S-Ou: deferred to later PR

packages/frontend/src/components/frames/FrameListPanel.tsx
S-Ou: deferred to later PR

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
S-Ou: deferred as a future accessibility improvement

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -48,30 +48,25 @@ const DetailsTable = styled("table")`
   max-inline-size: 100%;
   width: 100%;
 
-  tr:not(:last-child) > td {
-    padding-bottom: 0.75rem;
+  tr:not(:last-child) > td,
+  tr:not(:last-child) > th {
```
jaskfla:
```suggestion
  tr:not(:last-child) > :is(td, th) {
```

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -48,30 +48,25 @@ const DetailsTable = styled("table")`
...
+    padding-block-end: 0.75rem;
   }
 `;
 
-const TableCellIcon = styled("td")`
-  width: 2.5rem;
+const TableCellIcon = styled("th")`
```
jaskfla: Sorry, should’ve caught this earlier

```suggestion
const TableHeader = styled("th")`
```

This _contains_ the icon, but it is not the icon. The icon is the SVG

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -48,30 +48,25 @@ const DetailsTable = styled("table")`
...
+  inline-size: 2.5rem;
   white-space: nowrap;
+  opacity: 0.6;
 `;
 
 const TableCellContent = styled("td")`
```
jaskfla:
```suggestion
const TableCell = styled("td")`
```

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
```diff
@@ -214,27 +209,34 @@ function DetailsCard({ frame }: { frame: Frame }) {
...
             <tr>
-              <TableCellIcon>{ownerInfo.icon}</TableCellIcon>
+              <TableCellIcon aria-label="Frame owner">
+                {ownerInfo.icon}
+              </TableCellIcon>
               <TableCellContent>{ownerInfo.label}</TableCellContent>
             </tr>
             <tr>
-              <TableCellIcon>
+              <TableCellIcon aria-label="Frame dimensions">
```
jaskfla: Would slightly prefer a visually hidden child over `aria-label`:

- it works better with translation tools; and
- it’s more obvious in dev environment because it will show up when you hit <kbd>Alt</kbd>

packages/frontend/src/util/frame.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function getSystemFrameId(frameType: SystemFrameType): string {
+  return `${SYSTEM_FRAME_PREFIX}${frameType}`;
+}
+
+/**
+ * System frame definitions and utilities.
+ * System frames are client-side only constructs that don't exist in the database.
+ */
+export const SystemFrames: Record<
```
jaskfla, line 21: Are you anticipating more? Smells like premature generalisation, but it’s minor so I could be convinced to keep it

Would suggest renaming to `systemFrames`, though, so it doesn’t look like a plain object and not enum/type/FC/class

jaskfla (approved):
Neato! Couple rename suggestions but I think that’s mostly it. Preapproving from here 🥳

packages/frontend/src/util/frame.ts
sou323: I haven't ruled it out, I'm just not sure how exactly we'd approach it

sou323: but the concept of system frames were with the intent that they were somewhat future-proofed

packages/frontend/src/components/frames/FrameDetailsPanel.tsx
sou323: aha, i knew something like that must have existed

PR #794, opened 2026-06-19
refactor: inclusive bounds

Rocked03:
It's been a mess because our coords haven't been consistently inclusive or exclusive when calculating bounds (for frames and such)

So I've opted to make everything inclusive, since that matches our frame creation UI


packages/backend/src/services/exportService.ts
```diff
@@ -137,8 +133,8 @@ export async function exportCanvasBoundsAsStream({
       );
     }
 
-    const cropX = x0 * scale;
-    const cropY = y0 * scale;
+    const cropX = bounds.x0 * scale;
+    const cropY = bounds.y0 * scale;
```
stijnvdkolk: we can still use x0 and y0 here right? see line 119

stijnvdkolk (approved):

PR #795, opened 2026-06-19
feat(frontend): improved bounds select

Rocked03:
https://github.com/user-attachments/assets/64eb7e40-dff2-4ffb-acb1-78ddd7caa1db


jaskfla (approved):
Would suggest giving behaviour that mimics `object-fit: contain`, but code looks reasonable enough

🉑

PR #797, opened 2026-06-21
feat(backend): more stats views

### LINKED ISSUE #776: Expanded frame info & stats

Rocked03:
## New views
- `frame_stats` - total users, total pixels, last placed at
- `canvas_colors`, `frame_colors` - color, count
- `color_leaderboard`, `color_leaderboard_frame` - color, user, rank
## New/updated endpoints
- `/statistics/leaderboard/:canvasId` -> `/statistics/leaderboard/canvas/:canvasId`
- \+ `/statistics/leaderboard/frame/:frameId`, same params as the existing leaderboard endpoint (ditto to below)
- \+ `/statistics/leaderboard/canvas/:canvasId/color/:colorId`
- \+ `/statistics/leaderboard/frame/:frameId/color/:colorId`
- \~ `/summary/canvas/:canvasId` now includes `colorDistribution { color, count }[]`
- \+ `/summary/frame/:frameId`, same response format as the canvas summary

### DIFF
--- a/packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
+++ b/packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
@@ -3,13 +3,13 @@ SELECT
   history.user_id,
   history.canvas_id,
   history.color_id,
-  count(*) AS total_pixels,
+  COUNT(*)::integer AS total_pixels,
   rank() OVER (
     PARTITION BY
       history.canvas_id,
       history.color_id
     ORDER BY
-      (count(*)) DESC
+      (COUNT(*)::integer) DESC
   ) AS rank
 FROM
   history
@@ -31,14 +31,14 @@ SELECT
   history.canvas_id,
   frame.id as frame_id,
   history.color_id,
-  count(*) AS total_pixels,
+  COUNT(*)::integer AS total_pixels,
   rank() OVER (
     PARTITION BY
       history.canvas_id,
       frame.id,
       history.color_id
     ORDER BY
-      (count(*)) DESC
+      (COUNT(*)::integer) DESC
   ) AS rank
 FROM
   history
@@ -66,13 +66,13 @@ SELECT
   history.user_id,
   history.canvas_id,
   frame.id AS frame_id,
-  count(*)::integer AS total_pixels,
+  COUNT(*)::integer AS total_pixels,
   rank() OVER (
     PARTITION BY
       history.canvas_id,
       frame.id
     ORDER BY
-      count(*) DESC
+      COUNT(*)::integer DESC
   ) AS rank
 FROM
   history
--- a/packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
+++ b/packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
@@ -1,0 +1,66 @@
+DROP VIEW IF EXISTS color_leaderboard;
+
+DROP VIEW IF EXISTS color_leaderboard_frame;
+
+CREATE VIEW color_leaderboard AS
+SELECT
+  history.user_id,
+  history.canvas_id,
+  history.color_id,
+  count(*) AS total_pixels,
+  rank() OVER (
+    PARTITION BY
+      history.canvas_id,
+      history.color_id
+    ORDER BY
+      (count(*)) DESC
+  ) AS rank
+FROM
+  history
+WHERE
+  history.user_id NOT IN (
+    SELECT
+      user_id
+    FROM
+      blacklist
+  )
+GROUP BY
+  history.user_id,
+  history.canvas_id,
+  history.color_id;
+
+CREATE VIEW color_leaderboard_frame AS
+SELECT
+  history.user_id,
+  history.canvas_id,
+  frame.id as frame_id,
+  history.color_id,
+  count(*) AS total_pixels,
+  rank() OVER (
+    PARTITION BY
+      history.canvas_id,
+      frame.id,
+      history.color_id
+    ORDER BY
+      (count(*)) DESC
+  ) AS rank
+FROM
+  history
+INNER JOIN frame ON
+  history.canvas_id = frame.canvas_id
+  AND history.x >= frame.x_0
+  AND history.x <= frame.x_1
+  AND history.y >= frame.y_0
+  AND history.y <= frame.y_1
+WHERE
+  history.user_id NOT IN (
+    SELECT
+      user_id
+    FROM
+      blacklist
+  )
+GROUP BY
+  history.user_id,
+  history.canvas_id,
+  frame.id,
+  history.color_id;
--- a/packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
+++ b/packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
@@ -2,6 +2,8 @@ DROP VIEW IF EXISTS color_leaderboard;
 
 DROP VIEW IF EXISTS color_leaderboard_frame;
 
+DROP VIEW IF EXISTS leaderboard_frame;
+
 CREATE VIEW color_leaderboard AS
 SELECT
   history.user_id,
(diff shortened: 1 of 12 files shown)


packages/backend/prisma/core/migrations/20260621100000_add_frame_stats_views/migration.sql
```diff
@@ -0,0 +1,76 @@
...
+ORDER BY
+  history.canvas_id,
+  count DESC;
+
+CREATE VIEW frame_colors AS
+SELECT
+  frame.id AS frame_id,
+  history.canvas_id,
+  history.color_id,
+  count(*) AS count
```
stijnvdkolk: Output of `count(*)` is a bigint, it is not cast into an int here, so Prisma schema is wrong, we should probs cast to int as we do in other places. We use a mix of `count` and `COUNT`, we should probs find a standard.

packages/backend/prisma/core/migrations/20260621100000_add_frame_stats_views/migration.sql
```diff
@@ -0,0 +1,76 @@
...
+  )
+GROUP BY
+  frame.id,
+  frame.canvas_id;
+
+CREATE VIEW canvas_colors AS
+SELECT
+  history.canvas_id,
+  history.color_id,
+  count(*) AS count
```
stijnvdkolk: Output of `count(*)` is a bigint, it is not cast into an int here, so Prisma schema is wrong, we should probs cast to int as we do in other places. We use a mix of `count` and `COUNT`, we should probs find a standard.

packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
```diff
@@ -0,0 +1,92 @@
+CREATE VIEW color_leaderboard AS
+SELECT
+  history.user_id,
+  history.canvas_id,
+  history.color_id,
+  count(*) AS total_pixels,
```
stijnvdkolk: Output of `count(*)` is a bigint, it is not cast into an int here, so Prisma schema is wrong, we should probs cast to int as we do in other places. We use a mix of `count` and `COUNT`, we should probs find a standard.

packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
```diff
@@ -0,0 +1,92 @@
...
+  history.canvas_id,
+  history.color_id;
+
+CREATE VIEW color_leaderboard_frame AS
+SELECT
+  history.user_id,
+  history.canvas_id,
+  frame.id as frame_id,
+  history.color_id,
+  count(*) AS total_pixels,
```
stijnvdkolk: Output of `count(*)` is a bigint, it is not cast into an int here, so Prisma schema is wrong, we should probs cast to int as we do in other places. We use a mix of `count` and `COUNT`, we should probs find a standard.

packages/backend/prisma/core/migrations/20260621000000_add_color_leaderboard_views/migration.sql
```diff
@@ -0,0 +1,92 @@
...
+  history.canvas_id,
+  frame.id,
+  history.color_id;
+
+CREATE VIEW leaderboard_frame AS
+SELECT
+  history.user_id,
+  history.canvas_id,
+  frame.id AS frame_id,
+  count(*)::integer AS total_pixels,
```
stijnvdkolk: We use a mix of `count` and `COUNT`, we should probs find a standard.

packages/backend/src/routes/api/v1/statistics.ts
```diff
@@ -32,7 +39,7 @@ statisticsRouter.get(
 );
 
 statisticsRouter.get(
-  "/leaderboard/:canvasId",
+  "/leaderboard/canvas/:canvasId",
```
stijnvdkolk, line 42: Breaking change, should be fine though right?

packages/frontend/src/hooks/queries/useLeaderboard.tsx
```diff
@@ -1,25 +1,66 @@
...
+    queryFn: getLeaderboard,
+    placeholderData: keepPreviousData,
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    staleTime: 30_000, // 30 seconds
+  });
+}
+
+export function useFrameLeaderboard(
+  frameId: number,
```
stijnvdkolk: Not a number, its a string

packages/frontend/src/hooks/queries/index.ts
```diff
@@ -5,7 +5,7 @@ export { useCanvasList } from "./useCanvasList";
 export { useCanvasStats } from "./useCanvasStats";
 export { useEventInfo } from "./useEventInfo";
 export { useEventStats } from "./useEventStats";
-export { useLeaderboard } from "./useLeaderboard";
+export { useCanvasLeaderboard } from "./useLeaderboard";
```
stijnvdkolk, line 8: `useFrameLeaderboard` is not exported here yet, same for`useFrameStats`

packages/backend/src/services/statisticsService.ts
```diff
@@ -108,14 +118,191 @@ export async function getLeaderboard(
...
+    select: {
+      color_id: true,
+      count: true,
+    },
+    orderBy: {
+      count: "desc",
+    },
+  });
+
+  if (!stats || !colorDistribution) {
```
stijnvdkolk: `colorDistribution` comes from `findMany`, so it's always an array. `!colorDistribution` can never be `true`. We should probably check `colorDistribution.length === 0` instead.

packages/backend/src/services/statisticsService.ts
```diff
@@ -146,3 +337,37 @@ export async function getEventStatisticsSummary(
...
+  const colorDistribution = await prisma.frame_colors.findMany({
+    where: { frame_id: frameId },
+    select: {
+      color_id: true,
+      count: true,
+    },
+    orderBy: {
+      count: "desc",
+    },
+  });
```
stijnvdkolk, line 361: We should probably put this under the stats error check

packages/backend/src/services/statisticsService.ts
```diff
@@ -108,14 +118,191 @@ export async function getLeaderboard(
...
 
+export async function getCanvasColorLeaderboard({
+  canvasId,
+  colorId,
+  page = 1,
+  size = 10,
+}: {
+  canvasId: CanvasInfo["id"];
+  colorId?: PaletteColorSummary["id"];
+} & PaginatedParams): Promise<Paginated<typeof LeaderboardEntrySchema>> {
```
stijnvdkolk, line 129: `colorId` is optional here, but when it's `undefined` Prisma drops the filter and `orderBy: { color_id: "asc", rank: "asc" }` returns rank-1-per-color interleaved, with `skip`/`take` paginating across all colors. The only place this is called is where we provide colorId, so we could make it required to make sure this doesn't happen.

stijnvdkolk (changes requested):

packages/backend/src/routes/api/v1/statistics.ts
Rocked03: should be fine, i've updated the one place in the front-end that uses it

and it's not made to be a public endpoint so if people are hooking into it themselves then that's on them 🤷

packages/frontend/src/hooks/queries/useLeaderboard.tsx
Rocked03: whoops, didn't notice since i hadn't hooked it into anything yet

packages/frontend/src/hooks/queries/index.ts
Rocked03: i'm actually just not going to-- it'll be redundant after #737 so I'm just getting ahead of it for new stuff here

packages/backend/src/services/statisticsService.ts
Rocked03: no, that's intentional for #800 so I can fetch all the color leaderboards at once

stijnvdkolk (approved):

PR #802, opened 2026-06-21
fix(frontend): disappearing icons

Rocked03:
dunno what was happening here, but having an icon and a `VisuallyHidden` in the same `div` was resulting in the icon not appearing until the `VisuallyHidden` was made visible, or removed entirely


jaskfla (approved):
That _is_ a weird one. But apparently moving `opacity` from the `<th>` to the icon fixes it, which I’ve pushed directly

Will let you merge if you‘re cool with that approach instead

Rocked03:
Happy with that fix yeah

PR #805, opened 2026-06-26
feat(frontend): volume controls

### LINKED ISSUE #619: Notification sounds are a little loud

Rocked03:
https://github.com/user-attachments/assets/9946edd5-ac1f-49e4-87ec-098756a3e92c

closes #619

### DIFF
--- a/packages/frontend/src/app/settings/CheckboxSetting.tsx
+++ b/packages/frontend/src/app/settings/CheckboxSetting.tsx
@@ -1,5 +1,5 @@
-import { Button, styled } from "@mui/material";
-import { Volume, Volume1, Volume2, VolumeX } from "lucide-react";
+import { styled } from "@mui/material";
+import { Volume1, Volume2 } from "lucide-react";
 import { useEffect, useId, useRef } from "react";
 
 const VOLUME_PREVIEW_DEBOUNCE_MS = 300;
@@ -45,19 +45,6 @@ const VolumeSliderWrapper = styled("div")`
 
 const VolumeSlider = styled("input")`
   inline-size: 100%;
-`;
-
-const VolumeLabel = styled("span")`
-  font-variant-numeric: tabular-nums;
-  min-inline-size: 5ch;
-  text-align: right;
-`;
-
-const VolumeButton = styled("button")`
-  background: none;
-  border: none;
-  color: inherit;
-  cursor: pointer;
 `;
 
 interface CheckboxSettingProps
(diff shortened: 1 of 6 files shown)


packages/frontend/src/app/settings/CheckboxSetting.tsx
```diff
@@ -33,6 +36,36 @@ const Description = styled("p")`
   margin-block-start: 0.5em;
 `;
 
+const VolumeWrapper = styled("div")`
```
jaskfla: Maybe `VolumeControlWrapper`?

packages/frontend/src/app/settings/CheckboxSetting.tsx
```diff
@@ -70,6 +104,96 @@ export default function CheckboxSetting({
...
+    undefined,
+  );
+
+  useEffect(() => {
+    return () => {
+      if (debounceTimeoutRef.current) {
+        clearTimeout(debounceTimeoutRef.current);
+      }
+    };
+  }, []);
```
jaskfla: Worth defining a `useDebounce`, or use es-toolkit’s `debounce`

packages/frontend/src/app/settings/CheckboxSetting.tsx
```diff
@@ -70,6 +104,96 @@ export default function CheckboxSetting({
...
+          />
+          <TickLabels>
+            <span>0</span>
+            <span>50</span>
+            <span>100</span>
+          </TickLabels>
+        </VolumeSliderWrapper>
+        <Volume2 />
+
+        <datalist id="ticks">
```
jaskfla: Optional, but would follow rest of this codebase more closely if the ID came from `useId` instead of a magic string

Guarantees uniqueness in the DOM

packages/frontend/src/app/settings/CheckboxSetting.tsx
jaskfla, line 1: Nice extension of the existing component

packages/frontend/src/app/settings/Settings.tsx
```diff
@@ -51,29 +52,49 @@ const Aside = styled("aside")`
...
+      onVolumePreview={play}
+      previewAudioRef={audioRef}
     />
   );
 }
 
 function CooldownExpiryJingleSetting() {
-  const [value, setValue] = useLocalStorage("audio/cooldown-jingle");
+  const [enabled, setEnabled] = useLocalStorage("audio/cooldown-jingle");
+  const [volume, setVolume] = useLocalStorage("audio/cooldown-jingle/volume");
```
jaskfla, line 80: Oh, hm, should the volumes be individually controlled?

I could see an alternative approach where the volume is an _additional_ setting at the same level as the existing ones (instead of a “sub-setting”)

packages/frontend/src/hooks/sfx.ts
```diff
@@ -1,29 +1,53 @@
...
-  if (typeof enabled !== "undefined") return enabled ? play : noop;
-  // …otherwise defer to user preference
-  return globallyEnabled ? play : noop;
+  const playOrNoop =
+    typeof enabled !== "undefined" ?
+      enabled ? play
+      : noop
+      // …otherwise defer to user preference
+    : globallyEnabled ? play
+    : noop;
```
jaskfla: Bit gnarly to read. Maybe an IIFE?

And just name it `play`

jaskfla (commented):
Before you address the nitpicks, the big question is whether volume for each setting should be individually controlled

I kinda lean toward just having one volume slider that affects all sounds 🤔

jaskfla (approved):
Given we:

- resolved offline to simmer on how better to design the UX around this; and
- don’t think the current one is actually bad

…have a stamp. Just the usual-style niggles, but nothing show-stopping ✅

Nice li’l improvement, even if we have ideas to improve it further

PR #806, opened 2026-06-26
feat(backend/nest): implement soft lock support

### LINKED ISSUE #770: Backend NestJS: Soft-lock support

stijnvdkolk:


packages/backend-nest/src/canvas/canvas-cache.service.ts
```diff
@@ -4,6 +4,7 @@ import {
   CANVAS_EXPORT_SCALES,
   type CanvasExportScale,
   type CanvasInfo,
+  CanvasPlaceState,
```
jaskfla, line 7: Discussed before so pardon my repetition

I must’ve missed when this was added to the **types** package (or not?), but I’m a bit apprehensive about `CanvasPlaceState` the `const` being in the types package. Personally I’m quite happy to trust the language server, so would just define `type CanvasPlaceState` as a union of string literal types

It seems, though, that you and Rocked more naturally reach for the `const` object that behaves like an enum. Happy to accept this too, but I reckon that belongs in a constants package—not in types

In any case, I’m not by any means offended by this. Not something to address in this PR anyway

jaskfla (approved):
One out-of-scope papercut that I’m pretty sure we’ve chatted about in Discord before

But overall looking great 🙏

packages/backend-nest/src/canvas/canvas-cache.service.ts
stijnvdkolk: Yeah we should probably move this over to a constants package in the long-term, we can make a separate issue for that I think

